PR #11, opened 2016-07-17
Retry network requests on failure, move search into background task

marauder37:
This isn't 100% ready for merge but I wanted to share work-in-progress (especially since it's late here). The main thing this does is run the search code as a background thread, so you can reload the map page as often as you like and instantly see whatever the current results are.  The other big thing it does is retry all network requests on failure, so when the servers are being flaky like they are right now, the search still goes on.  Also, some formatting clean-ups.

### DIFF
--- a/example.py
+++ b/example.py
@@ -10,30 +10,42 @@ import struct
 import json
 import time
 import argparse
+import threading
+import atexit
+import sys
+
+import werkzeug.serving
+
 import pokemon_pb2
+
 from google.protobuf.internal import encoder
-from s2sphere import *  
+from s2sphere import *
 from datetime import datetime
 from geopy.geocoders import GoogleV3
+from geopy.exc import GeocoderTimedOut, GeocoderServiceError
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
+from requests.adapters import ConnectionError
+from requests.models import InvalidURL
+
 requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
This looks pretty good, good job! Do you think we should merge it yet?

BrodrickChilds:
Dang this is nice.  I want to fork a version with this merged in!

Not 100% confident enough in my python knowledge since it's been about 6 years since I last wrote with it, but there doesn't seem to be any obvious mistakes.  Does it run fine locally?

me-12:
works for me

me-12:
only one thing. it looks like i cant stop the app anymore with ctrl + c. this is a bit annoying because then i have to close the whole window, open a new console and type the whole command again.

marauder37:
I've rebased the PR so that it merges cleanly with all the new stuff from other PRs.

Thanks @me-12 , ccf71f2 also fixes the problem with ^C not stopping the background thread.

The Australian servers are down so I can't test it thoroughly. If anyone else wants to checkout my branch and test it that'd be appreciated. Otherwise I'll try again tomorrow.

AHAAAAAAA:
I've been testing this out and the map doesn't seem to load anymore, the redundant calls are working and it's getting the data but everytime you refresh it tries to reload the data. It was working a few commits ago with some merge work.

marauder37:
@AHAAAAAAA try again now, latest commit should fix the unwanted reload

PR #25, opened 2016-07-17
fixed teams

me-12:
For me the teams have been wrong. I changed values and compared that with the Pokemon Go app.


AHAAAAAAA:
You're right, good catch!

Issue #27, opened 2016-07-17
Allow specifying Flask options on the command line


Slashbunny:
It would be nice if we could specify a different bind address, port, and disable the built-in debugging for the Flask server. Something like this:

http://flask.pocoo.org/snippets/133/

It looks like the argument parsing and the flask server startup might need to be re-arranged in the file.

DerScreever:
would be really great

marauder37:
Working on this

marauder37:
@Slashbunny @DerScreever test the PR please, if you can

DerScreever:
I will do this later.
Now i need to work ^^

DerScreever:
done

PR #54, opened 2016-07-18
Accept command line arguments for Flask listen address and port

marauder37:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

-Accept command line arguments for Flask listen address and port
-Move arg parsing to another file
-Provide search completion % to page template

Relates to #27

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,8 +1,5 @@
 # encoding: utf-8
 
-from flask import Flask, render_template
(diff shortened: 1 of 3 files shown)


marauder37:
If Filthy Frank says it works, that's good enough for me

PR #55, opened 2016-07-18
Added transformation of coordinates for China players.

MonkeyLeeT:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

Through use of "-c" flag, coordinates transformation will be enabled for locations inside China (outside China it's disabled even with -c flag). Tested with friends and all is working perfectly.

Reference: http://gis.stackexchange.com/questions/141542/what-causes-the-gps-offset-shift-in-china


JZ6:
@marauder37 i think at this point we should just merge the other changes. way too many pulls to rewrite

JZ6:
@AHAAAAAAA this seems like a good feature, testing time?

AHAAAAAAA:
Seems to be working, I can't confirm though since I'm not in China haha 😋

JZ6:
yep it doesn't modify the original code, only two flow controllers, plus a standalone file, lets commit this

JZ6:
Thanks for the work monkey!

PR #75, opened 2016-07-18
Added Gym badges / pokestop icons

JZ6:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

-
-
-


medghaim:
ooo, that's an awesome pokestop icon!

CarlTaylor1989:
I do like the pokestop icon too!

nakami:
that bright blue of the pokestop icon might be really hard to read on the map. how about adding a black outline/stroke?

haagiboy88:
Agreed. Add an outline and remove the two smaller icons below the main icon? Less clutter perhaps?

PR #79, opened 2016-07-18
queue and then replace lists so markers do not get stale

simplydallas:
This pull request includes a

- [ x] Bug fix
- [x ] New feature
- [ ] Translation

-added queue lists for pokestops, gyms, and pokemons
-fills new queue lists during loop
-at end of loop copies contents of queue loops over existing primary list contents using deepcopy

### DIFF
--- a/example.py
+++ b/example.py
@@ -12,6 +12,7 @@ import time
 import requests
 import argparse
 import threading
+import copy
 
 import werkzeug.serving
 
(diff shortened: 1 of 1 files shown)


DerScreever:
think this is usefull if we want to add something like despawn of pokemons etc.

simplydallas:
Added a second commit that allows the first run loop to allow partial updates instead of waiting for a full loop to be queued.  That way we get the best of all worlds.  Quick partial results on start and full loop updates that replace stale ones that may have already timed out without worrying about partial results after it has been running a while.

favll:
Shouldn't lines 465 - 468 in example.py be unidented by one level?

marauder37:
This PR might need some love @simplydallas

DerScreever:
#101 
this is the better solution we think.

AHAAAAAAA:
I agree with @DerScreever, closing this for now.

simplydallas:
Agreed #101 is a better solution, assuming that id is unique.

PR #81, opened 2016-07-18
Added binding to any network interface

lukahn:
This pull request includes a fix to allow the code to bind to any network interface, not just 127.0.0.1. This also means that the server can also be connected to externally.

### DIFF
--- a/example.py
+++ b/example.py
@@ -560,4 +560,4 @@ def fullmap():
 
 if __name__ == "__main__":
     register_background_thread(initial_registration=True)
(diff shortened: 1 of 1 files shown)


frdmn:
This should not be hardcoded IMHO - same with the port setting.

frdmn:
@MelonSmasher raised #106 for a configurable setting.

MelonSmasher:
I agree with @frdmn, it should listen locally unless it is configured otherwise.

DerScreever:
Alread done with #106

PR #85, opened 2016-07-18
Enforce password length limit

Nyubis:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- The undocumented maximum password length is now applied. This should fix the issues that some people had while logging in. Previously they would send a password that was too long and get a "wrong password" reply.

### DIFF
--- a/example.py
+++ b/example.py
@@ -250,7 +250,9 @@ def login_ptc(username, password):
         return None
 
     # Maximum password length is 15 (sign in page enforces this limit, API does not)
-    password = password[:15]
+    if len(password) > 15:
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
This solution looks clean and cuts off passwords at 15 chars. I feel like adding some output along the lines of "trimming password" would be good.

Nyubis:
I've added the output.

AHAAAAAAA:
+1 from me

marauder37:
+1 lgtm

PR #94, opened 2016-07-18
Add two arguments to disable display of pokestop and gym on the map

denia42:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Add two arguments to disable display of pokestop and gym on the map

### DIFF
--- a/example.py
+++ b/example.py
@@ -346,8 +346,8 @@ def main():
     parser.add_argument("-i", "--ignore", help="Pokemon to ignore (comma separated)")
     parser.add_argument("-d", "--debug", help="Debug Mode", action='store_true')
     parser.add_argument("-c", "--china", help="Coord Transformer for China", action='store_true')
-    parser.add_argument("-dp", "--disable_pokestop", help="Disable Pokestop display", action='store_true', default=False)
-    parser.add_argument("-dg", "--disable_gym", help="Disable Gym display", action='store_true', default=False)
+    parser.add_argument("-dp", "--disable-pokestop", help="Disable Pokestop display", action='store_true', default=False)
+    parser.add_argument("-dg", "--disable-gym", help="Disable Gym display", action='store_true', default=False)
     parser.set_defaults(DEBUG=True)
     args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


DerScreever:
maybe you shoudn't name it "Disable Pokestop display".
ether write "Display Pokestop" and use the default value of True.
Then its is easier to read. Then you also do not need to do this  == False check.

favll:
Also, usually command line options do not use underscores. Please use --disable-pokestop and --disable-gym instead.

AHAAAAAAA:
I second DerScreever and favll, but otherwise this looks really clean!

denia42:
Accordling all comments, the changes are made on the last commit.

DerScreever:
Nah im not happy with the "Disable Pokestop display" ^^ like my comment before :D but if the other say its okay then i doesn't say anything ^^

AHAAAAAAA:
Yeah, I agree with DerScreever. It's much more intuitive to understand display-pokestop False instead of disable-pokestop True.

denia42:
@DerScreever : I have misunderstood your comment.  
Commited with the good logic.

example.py
```diff
@@ -346,8 +346,8 @@ def main():
     parser.add_argument("-i", "--ignore", help="Pokemon to ignore (comma separated)")
     parser.add_argument("-d", "--debug", help="Debug Mode", action='store_true')
     parser.add_argument("-c", "--china", help="Coord Transformer for China", action='store_true')
-    parser.add_argument("-dp", "--disable-pokestop", help="Disable Pokestop display", action='store_true', default=False)
-    parser.add_argument("-dg", "--disable-gym", help="Disable Gym display", action='store_true', default=False)
+    parser.add_argument("-dp", "--display-pokestop", help="Display Pokestop", action='store_true', default=False)
+    parser.add_argument("-dg", "--display-gym", help="Display Gym", action='store_true', default=False)
```
AHAAAAAAA: Gym display should be on by default in my opinion. Thoughts?

DerScreever: Yes should be default True. 
with the parameter you can turn them off.

AHAAAAAAA:
Alright I'll merge when you update that. Great work!

denia42:
To be sure, I need to change --display-gym to --hide-gym (to keep coherence d = display)
and keep --display-pokestop

AHAAAAAAA:
I was thinking more like this:
`parser.add_argument("-dp", "--display-pokestop", help="Display Pokestop", action='store_true', default=False)`
`parser.add_argument("-dg", "--display-gym", help="Display Gym", action='store_true', default=True)`

DerScreever:
@AHAAAAAAA 
good idea, but we should have one way we go with that. if we switch with similar things then we confuse other persons.

denia42:
The argument --display-gym is useless in this case, 
If you don't use it, the value is True because the default value is True
If you use it, the value is True, because action put True in the variable

DerScreever:
@denia42 
no it isn't
the default true show them. to disable the display just pass false as argument.

DerScreever:
@denia42 
your are right.
so set default to false. then everyone need to disable the display of one of these things.

denia42:
@DerScreever So my last commit is the good one :)

DerScreever:
@denia42 
Yes for me is it the good one :)

PR #43, opened 2016-07-18
Pull Request #24 Resolution (Google Login)

acadiansith:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

- Merged the changes from pull request #24 (Google login) with b9703b3 since the codebase has changed quite a bit from the time the pull request was made (and because the pull didn't work out for whatever reason).

If this is related to an existing ticket, include a link to it as well.

PR #95, opened 2016-07-18
new pokemon.proto - generate new pokemon_pb2.py

DerScreever:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- new pokemon.proto
- generate new pokemon_pb2.py
- this work for me and should also fix https://github.com/AHAAAAAAA/PokemonGo-Map/pull/43


derbeneviv:
i do confirm this works with google acc

DerScreever:
so we could also merge the google login if the owner rebuild the base #43

AHAAAAAAA:
Works on PTC. I +1 this.

PR #99, opened 2016-07-18
Add retina icons

marauder37:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

-	Add retina icons
-
-


AHAAAAAAA:
+1, will look great when we get through the frontend

marauder37:
DavidB is also +1 so in it goes

PR #101, opened 2016-07-18
Changed pokemons list into a dictionary to get rid of duplicates; also clears outdated pokemons.

nborrmann:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

- I refactored the pokemons list into a dictionary, using the SpawnPointId as key which should be sufficiently unique. This resolves the problem of duplicate pokemons in the list. It is also a much cleaner solution imo.
- The data for a pokemon is no longer a list but a dictionary as well. This makes the code a bit clearer.
- I also added a method that clears the dictionary of any outdated pokemons.

If this is appreciated, i can make the same change for gyms and pokestops.


DerScreever:
wow nice. was waiting for the clears outdated pokemons. when i'm back from work i want to do this on my own but now you did. great. but you need to rebase your branch, because of conflicts.

favll:
Yes, use this instead of #79.

simplydallas:
Agreed.  This is a better solution than the queue I made.

olipayne:
This is great, but the `pokemon['Name']` in printed in debug needs to be `pokemon['Name'].encode('utf-8')` because of Nidoran:

`[+] removing stale pokemon Nidoran♂ at 52.376481, 4.819766 from list
`

PR #104, opened 2016-07-18
Break main() into several functions

marauder37:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

The following changes were made

-Refactor main() into several functions
-Should be no changes to functionality, only factoring

If this is related to an existing ticket, include a link to it as well.

PR #109, opened 2016-07-18
Auto refresh

GuusK:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added global variable and flag to set it
- Added multiple REST calls to allow for AJAX request
- Added Jquery and AJAX request to update the map

### DIFF
--- a/templates/example_fullmap.html
+++ b/templates/example_fullmap.html
@@ -5,19 +5,68 @@
             <meta http-equiv="X-UA-Compatible" content="IE=edge">
             <title>Flask Google Maps Full Map Example</title>
             <link rel="stylesheet" href="">
+
+            <script type=text/javascript src="{{ url_for('static', filename='js/jquery-3.1.0.js') }}"></script>
             {{fullmap.js}}
 
-            {% if auto_refresh %}
-            <script type="text/javascript">
-                window.setInterval(function(){
-                  location.reload(true)
-                }, {{ auto_refresh }});
-            </script>
-            {% endif %}
     </head>
     <body>
         <h1>Flask Google Maps Full Map Example</h1>
         {{ fullmap.html }}
+
+        {% if auto_refresh %}
+    <script type="text/javascript">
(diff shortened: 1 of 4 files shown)


GuusK:
Sorry for the mess with git. I tried to fix it and pull the changes of the master into my fork and then into the branch, but it seems to have failed

DerScreever:
Nah no problem. ^^

nborrmann:
Looking at the code I really like it. However untested as the servers are down for me right now.

There will be conflicts with #101 and #104 though.

favll:
Just a few minor remarks concerning the command line option:

- change `--auto_refresh` to `--auto-refresh` to maintain consistency (dashes instead of underscores in cl arguments)
- maybe provide a default value (30 seconds maybe)?

Also I would change the `/getData` route to simply `/data` and the `/getConfig` route to `/config`, etc. Since this is a REST api, the verb of the operation is already part of the HTTP method (i.e. `GET /data HTTP/1.1`). 

Furthermore, change camelCase method names to snake_case to adhere to python style guide; `getConfig()` to `get_config`, etc.

Finally, Flask's `jsonfiy` method should be used to return json data. The added benefits are listed [here](http://flask.pocoo.org/docs/0.11/api/#flask.json.jsonify); e.g. change `return json.dumps(center)`, etc.

templates/example_fullmap.html
```diff
@@ -5,12 +5,69 @@
...
+        $.get(baseURL + "/getConfig", function(response){
+                        var json_obj = $.parseJSON(response);//parse JSON
+                        options["lat"] = json_obj["lat"];
+                        options["lng"] = json_obj["lng"];
+                        options["zoom"] = json_obj["zoom"];
+                        options["identifier"] = json_obj["identifier"];
+                    });
+
+        window.setInterval(function(){
+            // A new map is created because the original one isn't saved
```
rochacbruno: Np need to create a new map! now it is possible to reference `map` object direclty

PR #110, opened 2016-07-18
Addes a statistics argument option

galaktb:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added a -s argument option that makes a stats.txt file at the end of an iteration. It list the Pokemons found in descending order from the most common to the rarest


nborrmann:
This will count duplicates like crazy.

galaktb:
What kind of duplicates?

nborrmann:
You will count the same pokemon twice when it is found in two consecutive scans.

galaktb:
That's intended behaviour, why wouldn't I want to count it?

nborrmann:
Because it only spawned once.

Say for example, Gyarados spawns next to you for 15 minutes. It will be scanned each time and your statistics will tell you there were 15 Gyarados's there. 

This is of course misleading because it was all the same spawn.

medghaim:
One single pokemon can be found in multiple scans. Ie, The same Drowzee can be found in the data a few times. That's why we have `visible` and `seen`.

ryantheleach:
It's not misleading, as the gyarados was catchable for 15 times as long as a pokemon with a 1 minute despawn timer.

favll:
Writing stats into a flat file is not a very clean solution for various reasons. I also believe that it is way to early to add stats. Adding stats now, would require us to continuously update the stats function and keep an eye on it, everytime new changes and features are added. Also as mentioned before, your implementation seems to be faulty. I suggest closing this pull requests for now.

Issue #61, opened 2016-07-18
Countdown timers in infobox indicating when pokemon will disappear


AHAAAAAAA:
Currently we show despawn time statically based on when we load it, would be cool if we could have a countdown timer instead. Additionally, it would be great if we removed the marker when they do despawn

Djimusic:
I will take this on via javascript if that seems like the best idea!

sykhro:
Frontend team will take care of this

DerScreever:
Same as #31

sykhro:
Will close #31 just to not bother pinging assignees again. It wasn't a bug anyway, just not implemented

nborrmann:
Instead of updating the countdown timer, i changed the message to display the exact time at which the pokemon will despawn in PR #88. Imo this is much easier but just as good.

haagiboy88:
At the moment I agree nborrmann. But when we are able to properly despawn the icons at the correct time we should go back to "despawns in x time".

DerScreever:
Nah wont go back to this. take the soulution from @nborrmann for the UI the user see.
But internal use the "despawns in x time" and then update the map.
But i think it is easiert to read that the pokemon despawn at "14:30:02".

ortschun:
Countdown timer was there, but some recent commit seems to have that removed?

JZ6:
complete

PR #113, opened 2016-07-18
Update despawn time every second

foodforarabbit:
This pull request includes a
- [ ] Bug fix
- [x] New feature
- [ ] Translation

- added timestamp and class to label for pokemon timer (disappears at)
- added jquery + function to calculate time difference and update label

Notes:

- not too familiar with flask so I just wrote the js right on the page. Please feel free to refactor it as necessary.
- jQuery is probably overkill for what this does.
- haven't been able to test it "for real" because I'm unable to get any data. Should work though.

https://github.com/AHAAAAAAA/PokemonGo-Map/issues/61


nborrmann:
Whats the point of 
`if (true) {...} else {...}`
?

foodforarabbit:
I was thinking it'd be useful in the future if you want to toggle between
time remaining and time of despawn. The second block sets the label to time
of despawn, which is mostly static. Currently there is no point and it can
be removed.

favll:
This could be detrimental to performance if the map is accessed via smartphone and the project is running with a large step size (e.g. 10) and is displaying a lot of Pokémon.

foodforarabbit:
Yea thats true. Ideally we should default to showing the static "disappears
at" number, and then we have a toggle somewhere that starts the setInterval
when its checked, and clears it when it gets unchecked. I'm not too
familiar with flask though so if someone could set up the js skeleton then
I can go in and start adding stuff but as you can see, I have it all in an
script tag at the top which doesn't feel right.

AHAAAAAAA:
I agree with foodforarabbit, that's the cleaner more efficient solution (you only care about the pokemon you're hunting, not every single pidgey in the map).

PR #130, opened 2016-07-18
Credentials + credential loading - congregates credentials into one file.

medghaim:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

-
-
-


medghaim:
I'll add ANDROID_ID, SERVICE, and CLIENT_SIG to credentials as well. 

but i'll leave the auth links in example.py

DerScreever:
Approved, Test passed :+1:

medghaim:
Should be more pythonic now with .get(key, defval). Also safer, using `with open`

medghaim:
Redoing

PR #137, opened 2016-07-18
Added bypass for geocoder if raw lat,lng values are inputted

JakeCooper:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- If the user inputs raw latitude or longitude coordinates (e.g "48.405703, -123.350780"), `set_location` will now recognize this and not call the geocoder API. This is useful if users reach the rate limit for the API, or just useful in minimizing API calls in general.

### DIFF
--- a/example.py
+++ b/example.py
@@ -130,14 +130,22 @@ def retrying_set_location(location_name):
 
 
 def set_location(location_name):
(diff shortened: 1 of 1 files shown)


sykhro:
Tested, works fine

PR #142, opened 2016-07-18
settings.py file

borisrozumnuk:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added settings.py file to project folder
- Moved some constants to settings.py
- added logic to host and port selection when app starts

### DIFF
--- a/example.py
+++ b/example.py
@@ -29,18 +29,9 @@ from requests.packages.urllib3.exceptions import InsecureRequestWarning
 from requests.adapters import ConnectionError
 from requests.models import InvalidURL
 from transform import *
+from settings import *
 
 requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
-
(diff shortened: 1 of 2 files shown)


DerScreever:
Approved, Test passed :+1:

medghaim:
#130 ?

DerScreever:
@medghaim 
i will test #130 in a second

borisrozumnuk:
@medghaim   #130 is not a pythonic way, but if you want to get values of variables this way, you can make default value with dict.get(key, default=value_here)

borisrozumnuk:
#130

medghaim:
Ah, just figured since the keys aren't user input there's no chance of not finding the dict value since it's in the JSON file.

JZ6:
I might need to ask you to rebase in a bit.

JZ6:
outdated rebase and do new pull

PR #155, opened 2016-07-18
added werkzeug version requirement

macgngsta:
This pull request includes a

- [x ] Bug fix
- [ ] New feature
- [ ] Translation

- added a werkzeug version requirement in requirements.txt , the issue occurs if an older version of werkzeug was previously installed that is less than 0.10.x which has a new method signature: `werkzeug.serving.is_running_from_reloader()`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ s2sphere==0.2.4
 flask==0.11.1
 gpsoauth==0.3.0
 coveralls==1.1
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
Perfect, works and quashes a few issues submitted!

DerScreever:
Approved, Test passed :+1: 
We merge this later, after some really big merge ^^

PR #166, opened 2016-07-18
resized oversized pokemon icons

rgnvldr:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

Resized and centered all Pokemon icons so map doesn't end up looking like http://imgur.com/Wt0zXD3


DerScreever:
sorry but at the moment we are not accepting pull requests for icons etc. 
nice that you did this work, but we wait for the frontend-team to get ready.

rgnvldr:
Should I submit to the frontend branch then?

DerScreever:
no at the moment we are not accepting pull request for icons etc, until the frontend-team say something else.

favll:
I strongly advise to merge this change. It does not modify example.py and therefore should not conflict with the current blockers. It also simply resizes all icons to address a critical bug that makes the map unusable.

hhu94:
I originally made them a that size because I thought any smaller would make them harder to identify, especially on higher resolution screens (i.e. smartphones).

I think at medium-high zoom the bigger sprites look better, while at low zoom the smaller sprites look better.

I personally don't run into the problem @rgnvldr has, because I generally use the map at a higher zoom rate.

Is there a way to make the sprites has a size relative to the map zoom, instead of a static size?

DerScreever:
or maybe convert this to svgs and do some scaling?

rgnvldr:
Unless someone else tackles it, I can work on converting to svg and implementing scaling based on screen DPI tonight.

hhu94:
@rgnvldr That would be great! I don't have the experience to do it, sorry for being no help.

hhu94:
After using the tiny icons for the whole day I still think the bigger ones are better overall. Does anybody else think the same? If so could we revert upstream to the bigger icons?

benemorius:
I kept the bigger ones too (I guess that's no secret thanks to github :/) but eventually we probably need something more sophisticated than static sizes. Either scaling by zoom level or grouping or something like that.

hhu94:
@benemorius I completely agree, but until someone (maybe @rgnvldr) implements that we could use the bigger ones. I had to create a separate branch that keeps the bigger icons for my personal use haha.

into:
@hbu94 Could you add removing a pokemon from the map, maybe a check mark you can click on after clicking on the pokemon to remove it? Just so when I catch it, I can remove it from the map. Would be useful but not all that important. Thanks. :+1:

hhu94:
@into you should make a separate enhancement request in Issues for that.

PR #167, opened 2016-07-18
Adding Heroku deployment instructions & Procfile

Danm72:
This pull request includes a

- [ ] Bug fix
- [x ] New feature
- [ ] Translation

- Adding a Procfile for Heroku deploys
- Updating Readme to point people towards heroku for remote deploys
-


ur0:
The way twelve factor apps are meant to work, you shouldn't have to change any file to deploy your application. 

I've been working on Heroku's platform for a long time, and this isn't the recommended way to do so.
As a rule of thumb, NEVER store credentials in your code. 
See http://12factor.net for more info regarding this. 

Those purple-colored 'Deploy to Heroku'  buttons, which allow you to deploy the app in a single click, require an app.json to work.

mastershaig:
@ur0 you are right!

PR #171, opened 2016-07-18
Automatically remove gone Pokemon from map

RubenGM:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Saved disappear time to the pokemon array
- The map only shows those pokemon who are still there

==========
I had never touched Python before yesterday, so if I hope I didn't break something accidentally.
I tested this and it seems to be working.


DerScreever:
should already fixed with #101

JZ6:
#101 is blocked currently and this seems to work

medghaim:
So if 101 isn't getting merged, we aren't switching `pokemon` to a `dict`? 

Lines like: `currLat, currLon = pokemon[-3], pokemon[-2]` can cause issues down the line, if we change up how many arguments are passed, or their order.

nborrmann:
101 is going to be merged soon.

PR #175, opened 2016-07-18
Resized and centered icons to better fit map

rgnvldr:
This pull request includes a

- [x ] Bug fix
- [ ] New feature
- [ ] Translation

- Resized and centered all Pokemon icons so map doesn't end up looking like http://imgur.com/Wt0zXD3


JZ6:
already merged

PR #207, opened 2016-07-18
improved loop logs

KonstantinKuklin:
This pull request includes a

- [ ] Bug fix
- [*] New feature
- [ ] Translation

- Now loop logs stay on the same line

### DIFF
--- a/example.py
+++ b/example.py
@@ -573,7 +573,7 @@ def main():
     origin_lat = FLOAT_LAT
     origin_lon = FLOAT_LONG
     for step in range(steplimit**2):
-        stdout.write("\r[{}] Looping: step {} of {}".format(str(datetime.now()), step, steplimit ** 2))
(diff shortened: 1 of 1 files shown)


KonstantinKuklin:
it seems, that location can be changed inside the loop and I broke logging new coordinates, I`ll fix it right now :)

AHAAAAAAA:
Just tested it, you got my +1 when you sort out the coordinates!

nborrmann:
Don't like the use of flush. It will get weird when there are error messages. (Which is inevitable when the servers crash the next time)

KonstantinKuklin:
I'm done

AHAAAAAAA:
This introduces more problems than it's worth, erases other error messages that might come up during pinging niantics servers. I'm sorry I'll close this PR for now

PR #209, opened 2016-07-18
Map Autoupdate Fixes

foodforarabbit:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

- fixed tooltip for markers on map
- map does not need to be recreated every single update... just the first time
- stale markers cleared from map

### DIFF
--- a/example.py
+++ b/example.py
@@ -785,9 +785,9 @@ def next_loc():
 def get_pokemarkers():
     pokeMarkers = [{
         'icon': icons.dots.red,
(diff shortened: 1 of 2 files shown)


h1pp0:
Is this fix related to #210 ?

foodforarabbit:
No, I was working on auto-updating and then realized someone already implemented it. However, the tooltips were broken so I went ahead and fixed that.

PR #222, opened 2016-07-19
Use only one thread

seanogden:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

- Removed recursive thread spawning.


JZ6:
does not work, rebase and pull

PR #223, opened 2016-07-19
Ajax api for map markers update

loretoparisi:
This pull request includes a

- [X] Bug fix
- [X] New feature
- [ ] Translation

- added javascript functions `updateMarkers` and `cleanupMarkers` to example map page
- added a `calculate_markers` function in the example python
- added a flask endpoint `/api/v1/markers` to fetch markers
- a typo fix in javascript function `setLabelTime`
- Requires GoogleMaps update for supporting global map object - see https://github.com/rochacbruno/Flask-GoogleMaps/issues/29

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
-from flask import Flask, render_template
+from flask import Flask,jsonify,render_template
 from flask_googlemaps import GoogleMaps
 from flask_googlemaps import Map
 from flask_googlemaps import icons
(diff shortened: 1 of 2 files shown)


nborrmann:
Has been implemented in #109, #209 among others.

rochacbruno:
@nborrmann in #109 a new map is created, now FlaskGoogleMaps supports referencing the `map` object direclty, so no new map should be created, also it is now possible to reference existing markers by `map_markers` so easy to move or change marker icon or position.

this @loretoparisi PR provides better solution.

nborrmann:
Everything here has been implemented in #246. This here might be the better solution (i dont know, i didnt really look at it), but it's based on a really old codebase. I advise to close this and ask for a rewrite or go with what we currently have.

loretoparisi:
@nborrmann thanks,  I have seen the `/data` and `/config` endpoints. As @rochacbruno pointed out, there is no need to create a new map every time since - as I have asked him to expose the `google.map` in the `FlaskGoogleMaps` module and he kindly did it :), now the map object is globally available in the page as you can see from the `console`:

```shell
window.map
Ag {gm_bindings_: Object, __gm: Hf, gm_accessors_: Object, center: _.H, zoom: 15…}
window.map.getDiv()
<div id=​"fullmap" style=​"height:​ 100%;​ width:​ 100%;​ top:​ 0px;​ left:​ 0px;​ position:​ absolute;​ z-index:​ 200;​ overflow:​ hidden;​ transform:​ translateZ(0px)​;​ background-color:​ rgb(229, 227, 223)​;​">​…​</div>​
```

Btw, happy to have inspired something here 🎱

Issue #231, opened 2016-07-19
Scanned are slowly drifts away from starting location.


nborrmann:
I started at the red marker on the top left corner, the scanned are slowly walked into south-east direction away from the starting point.

![image](https://cloud.githubusercontent.com/assets/1723176/16935494/3bfcbc0c-4d60-11e6-934f-09bd5b08dcc7.png)

I used the `-l "Streetname, City"` option for setting the location.

JZ6:
whats your steps?

nborrmann:
it was 4 in the pc, but it doesnt matter.

timbennett:
Duplicate of #202

macgngsta:
can confirm. it seems that the start of the next scan takes the last scanned coordinate, instead of the origin coordinates. ill have to check the code to make sure my hypothesis is correct

![pokedrift](https://cloud.githubusercontent.com/assets/1333829/16936146/4da668b4-4d19-11e6-98a5-c6f4fbfbcd03.jpg)

nickgarvey:
These are global variables and are shared across threads. That's probably the problem. Use a thread local variable instead:
https://docs.python.org/2/library/threading.html#threading.local

https://github.com/AHAAAAAAA/PokemonGo-Map/blob/master/example.py#L191

amal:
@nickgarvey we can just reset location to original at the end of looping steps – it's already enough. see #239

nickgarvey:
Yup, makes sense.

Issue #229, opened 2016-07-19
Coordinate checking slowly moves towards the right, away from starting point


skxu:
GitHub issues are for bug reporting and feature requests.  
For general support, join [our discord server](https://discordapp.com/invite/qbpVH6b).
To make bug reports easy to find and organized, we close issues that are deemed
out of scope for GitHub Issues and point people to Discord.

Please provide the following information.
The more information you provide, the more easily we will be able to offer
help and advice.

### Environment info
Operating System: OSX

Installed version of Python: 2.7

### Steps to reproduce
1. Start the server (any example), step size 3 works well so you can see the movement better
2. Leave the server running for 15-20 min
3. The pokemon that appear on the map are no longer near the original point

### What have you tried?
1. Restarting the server fixes the issue

### Logs or other output that would be helpful
(If logs are large, please upload as attachment).

hhu94:
Experiencing the same issue.

XoneFobic:
I was just going to make the same issue.
To my feeling it seems to move South East from it's original starting point, each cycle.

Original Start point: `52.5195922 5.4806713`
After one cycle (100 steps / `-st 10`): `52.4995922 5.5056713`

nborrmann:
#231

h1pp0:
I noticed this too. Used -st 7 and GPS coords for the start location. I also experienced the "drift" towards the South East direction

Issue #202, opened 2016-07-18
Map center shifts to SE after refresh


AHAAAAAAA:
New bug, the map center is shifting when I refresh the second time after it starts loading pokemon.

AHAAAAAAA:
Fixed it, it was problem with yang's mods to location scanner algo and map display.

towicode:
@AHAAAAAAAI think the pointer is still on the original location but it seems to me the scan is still shifting to the SE.

nickgarvey:
See my comment on https://github.com/AHAAAAAAA/PokemonGo-Map/issues/231#issuecomment-233512629 for a possible root cause

JZ6:
fixed

PR #238, opened 2016-07-19
SouthEast drift fix

macgngsta:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- fixed the debug display statements on loop and completion
- fixed the SE drift by adding a reset to origin in the main loop

#202, #229, #231


macgngsta:
Seems to be fixed by Amal on #239

hhu94:
@macgngsta Your fix might be better, @amal's doesn't work. I'm going to test yours now.

hhu94:
[log.txt](https://github.com/AHAAAAAAA/PokemonGo-Map/files/370633/log.txt)

Locations of the two loops don't seem to match.

macgngsta:
@hhu94  - the main loop hasn't finished yet? the log.txt shows only 1 loop? was there another file?

hhu94:
@macgngsta Didn't paste the whole thing, sorry. I updated the link with a complete version.

hhu94:
@macgngsta You're absolutely right, I tested your master branch instead of your se_fix branch. The se_fix branch does indeed restore the original location at the start of each loop!

PR #246, opened 2016-07-19
ar fixes

foodforarabbit:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

- kept flask google maps in
- markers only update if new thing found
- old markers disappear properly (pokemon at 0s, gyms when icon changes on refresh, pokestops never)
- ugly js error caused by killing the flask element as soon as we initialize the google map from js


Gaudon:
This is not the optimal way to address this problem, however it's better than not addressing it at all and at least it temporarily restores prevoius functionality.

Issue #249, opened 2016-07-19
Steps 15-25 not moving according to logged coordinates


hhu94:
Using -st 5, steps 15 to 25 result in the exact same coordinates being scanned, not movement. Tested in 2 different locations.

### Environment info
Operating System:
Arch Linux

Installed version of Python:
2.7.12

### Steps to reproduce
1. Use the script and check the coordinates logged from steps 15 to 25

### Logs or other output that would be helpful
[log.txt](https://github.com/AHAAAAAAA/PokemonGo-Map/files/370641/log.txt)

amal:
confirmed

macgngsta:
it seems that by step 16 the if loop that increments the movement fails to execute (line 591):

```
if -steplimit / 2 < x <= steplimit / 2 and -steplimit / 2 < y \
            <= steplimit / 2:
```

kevvv:
Been running on 6 and 8 steps recently and have been noticing that it is indeed missing certain spots on the map.  Couldn't really pinpoint what it was though.

JZ6:
fixed

PR #252, opened 2016-07-19
Step fix

macgngsta:
This pull request includes a

- [x ] Bug fix
- [ ] New feature
- [ ] Translation

- fixed display of step and completion debug messages
- added a local var steplimit2 - to store step limit**2, which was referenced in the if condition as steplimit, but did not contain the squared number.

#249


hhu94:
+1, works perfectly!

JZ6:
accepted

PR #264, opened 2016-07-19
Moved data to persistent sqlite storage

bricej13:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Pokemon, pokestops, and gyms all stored in sqlite
- Data is persistent across restarts and removed when stale
- Used Peewee ORM for easy usage


JZ6:
useful!
Could be used when servers are down as well i presume

nborrmann:
I like it, nice code.

Maybe we could even drop the `clear_stale_pokemons()` and use a sql clause to only show the currently available pokemon. That way we keep the data about the pokemon spawns and can later run stats on them.

nborrmann:
@bricej13 can you resolve the conflicts? then we merge.

nborrmann:
I just noticed, we can't make SpawnPointId unique, if we don't `clear_stale_pokemons()`. There will be other pokemon at the same spawnpoint. But we dont have a unique identifier for the pokemons.

Maybe we can use the despawn time, but we will have to round it to seconds or maybe 10 seconds, because we will never get right to the millisecond.

bricej13:
@nborrmann Sorry, I went to bed after I committed this.

1. I went back and forth on whether or not to clear the stale pokemon. It could be useful for long-term metrics, but since that isn't built out yet I didn't see any use in letting the database slowly grow. 
That being said, the thumbs-ups have spoken and I'll pull it out.

2. I think we could use a compound primary key of the the spawnPointId and the despawn_time. I don't see any reason we'd need to truncate the despawn_time as we'd just be using `<` to extract active data.

DerScreever:
@bricej13  
pls wait with resolving conflicts. we are working on a big change.

KevakaBuel:
@bricej13 If your code is ready to use would you mind giving a detailed tutorial how to use it. I would really like to fill a database with data from the script do create heatmaps but im afraid that my knowledge will not be enough to set it up all by myself.

bricej13:
@DerScreever No problem. I can hardly keep up with all the changes anyways.

@KevakaBuel Check out the readme on the GitHub repo for [peewee](https://github.com/coleifer/peewee) and the [QuickStart Guide](http://peewee.readthedocs.io/en/latest/peewee/quickstart.html)

thedewpoint:
If this PR is merged will we need to do additional setup to get the SQLite db running or will it be bundled with the code

bricej13:
@thedewpoint The database is automatically created (if it doesn't exist) when you run `example.py`. It should be a seamless setup experience.

One thing it doesn't address, however, is if there are changes to the models in `models.py`, the tables would have to be deleted and re-added. The easiest way to do this is to just delete `pokemap.db`. This would cause you to lose your historical data.

DerScreever:
@bricej13 

maybe you can handle all the conflicts ^^ only told you, so you dont need to do the work twice :D

KevakaBuel:
So in the end i just run the script like i did before and it will create and fill a databse? How exactly is the db model and how can i access it? I know im a bit early but im really exited for this change.

bricej13:
@DerScreever is there a pr of but I can watch to know when you're done with the major changes?

@KevakaBuel The db model is in `model.py`. `example.py` imports the models from that file and creates/populates the database in the `if __name__ == '__main__':` function. Here's the code to create and populate:
``` python
    db.connect()
    Pokemon.create_table(fail_silently=True)
    Pokestop.create_table(fail_silently=True)
    Gym.create_table(fail_silently=True)
```

MikeDX:
I'd really love to see this working, and tried to merge but it fell on its ass a little - then got put off by the having to clear the db on schema change

How easy would it be to change the db to sqlite which would allow new columns to be added and be readable by a much wider range of things?

bricej13:
@MikeDX It is using sqlite. Peewee is the ORM running on top of it.

MikeDX:
Oh well in that case very easy!

Are you working on a resolve/merge? Is it worth me pursuing it?

bricej13:
I won't be able to get to it for another 5-6 hours or so, so you can go for it.

MikeDX:
I'll see how i get on, if you dont see a pull request from me before you can make a start, you can assume i've failed and or fallen asleep :)

KevakaBuel:
Is this part of the developer branch because i saw an issue with the not installed peewee package?
If yes could anyone update the usage section to cover the basics and link to it? Cant even run the developer branch because im not at home right now. 

Thanks you all for your work.

PR #266, opened 2016-07-19
Issue template added

olsynt:
This pull request includes a

- [ X ] New feature - Adding template to handle issues

-
-
-


DerScreever:
we have already a good template i think

PR #279, opened 2016-07-19
Display Gym Prestige in Infobox

Scyntrus:
This pull request includes a

- [ ] Bug fix
- [* ] New feature
- [ ] Translation

-Add Gym Prestige value to Infobox


Scyntrus:
It would be nice to show the gym level, as it can be determined by the prestige. However, I don't know the prestige to level values.

Issue #269, opened 2016-07-19
Lured pokestops disappear after the first update tick


ayalfishey:
GitHub issues are for bug reporting and feature requests.  
For general support, join [our discord server](https://discordapp.com/invite/qbpVH6b).
To make bug reports easy to find and organized, we close issues that are deemed
out of scope for GitHub Issues and point people to Discord.

Please provide the following information.

Not much to say after the first -ar lured pokestops disappear 

### Environment info
Operating System:
Windows 7

Installed version of Python:
2.7.12

### Steps to reproduce
1. start it with a value for -ar, after that many seconds the pokestops will vannish
2.
3.

### What have you tried?
1.

### Logs or other output that would be helpful
(If logs are large, please upload as attachment).

ryantheleach:
Are these actual lured pokestops? or some sort of bug?

ayalfishey:
im not at a location that's lured right now but based on central park its actually lured locations

Plop:
Got the same bug. Lured Pokestops disappears after the first auto-refresh.

PR #297, opened 2016-07-19
Fix bug where lured Pokestops disappear after 10 seconds

Scyntrus:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Fixes https://github.com/AHAAAAAAA/PokemonGo-Map/issues/269
- Previously, the disappear_time was unset for lured pokestops, causing a task to remove it from the map to run after 10 seconds.
- The changes to example_fullmap.html also handle the case where a lure expires or a new lure is used, forcing the map icon to refresh.

### DIFF
--- a/example.py
+++ b/example.py
@@ -873,6 +873,7 @@ def get_pokemarkers():
             pokeMarkers.append({
                 'type': 'lured_stop',
                 'key': stop_key,
+                'disappear_time': -1,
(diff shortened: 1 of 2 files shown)


benemorius:
:+1:

Octylpy:
+1 Tested it, Pokestops with lures stay on the map

PR #302, opened 2016-07-19
Fix phrasing in output when searching for Pokémon

svenluijten:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation
- [x] None of the above

- Fixed phrasing in the output to command line.
- Changed "pokemons" to its correct plural form; "Pokémon".

### DIFF
--- a/example.py
+++ b/example.py
@@ -630,7 +630,7 @@ def main():
 
 def process_step(args, api_endpoint, access_token, profile_response,
                  pokemonsJSON, ignore, only):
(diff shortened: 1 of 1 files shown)


svenluijten:
Actually, just noticed this outputs a "broken" character:

```
Searching for Pok├®mon at location [...]
```

I'm on Windows, can anyone confirm this? A workaround would be to replace the `é` with `u'\xe9'`. (@DerScreever)

PR #308, opened 2016-07-19
Added Rattata to ignore Pokémon example

Hjsurk:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [X] Translation

- Added Rattata to ignore Pokémon example in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ _Note:
 
 ## Working examples
 ### ignore Pokémon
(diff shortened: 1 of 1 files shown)


DerScreever:
think this one is not needed?

Hjsurk:
Nobody cares about Rattata, so why not make it easy for them and add it to the example? ;)

sashgorokhov:
I dont agree, rattatas are a good way to obtain stardust.

PR #309, opened 2016-07-19
Add Dockerfile to allow running in Docker

ultrafez:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

- Add a Dockerfile to build a Docker container running Python and the map application
- Update README to include instructions on how to use the application in Docker

PR #301, opened 2016-07-19
Dockerfile added

sashgorokhov:
This pull request includes a

- [   ] Bug fix
- [x ] New feature
- [   ] Translation

The following changes were made

- Added a Dockerfile for running map server inside a docker container without need to install all packages.

PR #317, opened 2016-07-19
Docker container for running master branch code

Ashex1:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

* Introduce Docker file for building docker image
* entrypoint script that pulls the latest code from master on initial start and any restarts

There are two existing PRs for this however I'd already built this last night and shared it on the Discord channel, I was just too sleepy to submit a PR. Also corrected some issues, it runs as desired. An image has been published on docker hub:

https://hub.docker.com/r/ashex/pokemongo-map/

Related PRs:

https://github.com/AHAAAAAAA/PokemonGo-Map/pull/309
https://github.com/AHAAAAAAA/PokemonGo-Map/pull/301

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,6 +12,7 @@ RUN apk add \
     build-base \
     unzip \
     wget \
+    git \
   && rm -rf /var/cache/apk/*
 
 RUN pip install geopy protobuf requests flask-googlemaps s2sphere flask gpsoauth coveralls werkzeug
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,24 @@
+# Basic a docker image for PokemonGo-Map
+# Usage:
+#   docker build -t pokemongo-map .
+#   docker run -d --name pokemongo-map pokemongo-map -a ptc -u UserName -p Password -l "Seattle, WA" -st 5
+FROM gliderlabs/alpine:3.3
+
+RUN apk add --update openssl ca-certificates
+RUN apk add \
+    python \
+    python-dev \
+    py-pip \
+    build-base \
+    unzip \
+    wget \
+  && rm -rf /var/cache/apk/*
+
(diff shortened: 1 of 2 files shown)


Swatto:
Looking good to me since Alpine is much lighter and the pull to get the latest code for such an active project is perfect.

Ashex1:
There's a need to be able to replace the API keys in `credentials.json` to the API quotas. I'm working on a change that'll support changing them via environmental variables (pass a variable starting with POKEMON_).

Swatto:
I don't you should. It would be way better to have an CLI argument for it like proposed here #361

Ashex1:
Excellent, then I won't bother with that :)

ultrafez:
This is the wrong approach to take when writing a Dockerfile to build an image. An image built from a Dockerfile is supposed to represent the fully-built application at a particular moment in time (including installed dependencies), that just needs to be started as a container to run the application. Pulling repo changes and reinstalling packages inside a freshly-spun up container isn't idiomatic Docker - this is treating it like a VM. The right approach is to pull the code, then rebuild the image so that it contains the new code, and then start the container.

ultrafez:
Publishing the app on Docker Hub is a good idea, and with a properly-built Dockerfile users will only have to pull a single filesystem layer when updating the application on their local machines. Building and publishing to Docker Hub can happen as part of CI builds too - that way, users don't even have to clone the code repo to use the app.

Stealthii:
I agree with @ultrafez - this is not really a docker image, containing the application I want to run. With this Dockerfile I have no concept of updated images for updated version of the application, nor a guarantee that the Docker image reflects a version of the app at a particular point in time.

Ashex1:
@ultrafez Hence why the title of the PR indicates it pulls the master branch ;)

I agree that a docker image should represent a fully-built application, currently the project has such frequent commits that users would already be running an out of date image as soon as it's running, additionally the lack of a branching model means that master isn't always in a functioning state.

KingScooty:
Would be nice to be able to pass a maps api key in to replace the existing one.

PR #328, opened 2016-07-19
Added pokemon artwork

POPSuL:
Like that

image

### DIFF
--- a/example.py
+++ b/example.py
@@ -826,9 +826,12 @@ def get_pokemarkers():
         if is_ampm_clock:
         	dateoutput = datestr.strftime("%I:%M%p").lstrip('0')
         pokemon['disappear_time_formatted'] = dateoutput
+        pokemon['img_id'] = '%03d' % pokemon['id']
 
         LABEL_TMPL = u'''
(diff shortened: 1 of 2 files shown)


mkody:
Just a bit unrelated, but I wonder if it's possible to do the same with Pokéstops.

MikeDX:
interesting, I wonder where the pokestops get their image url from - assuming it isnt covered in the pgo/api we use already and just isn't plugged in

POPSuL:
@mkody pokestops? You're sure? Meybe gyms?
If you talking about gyms - maybe, but it possible if anyone implements getting the list of the gym's pokemons

mkody:
@POPSuL Yes, I was talking about Pokéstops, having the picture of the place and maybe the name (but yeah, gyms too).

medghaim:
We had this a while ago, but we ended up getting rid of it. It is my opinion that that's overkill. We have their name, their icon, and a link linking them to their pokedex entry.

hgentry:
If you don't want it on by default, how about give a flag we can pass to turn it on? I think it's cute. :D

PR #288, opened 2016-07-19
Updated -i and -o to accept pokemon ids as well as names

MikeDX:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

The following changes were made

- Updated ignore and only parsing to accept ids from args in addition to names

PR #330, opened 2016-07-19
enable search by Pokedex ID

NgJason:
[ ] Improvement

This commit enables users to search for Pokemons by using their associated Pokedex ID.

Activated by using the -on flag, similar to -o. 
Ex: Zubat

-on "41"


DerScreever:
#288

already testing this.

PR #340, opened 2016-07-19
Fixed indentation formatting in credentials.json

h1pp0:
This pull request includes a

- [ x] Bug fix
- [ ] New feature
- [ ] Translation

- Fixed Alignment
-
-


medghaim:
Yeah thanks, no idea why it freaked out like that when I wrote it. Only noticed right before I went to bed.

PR #387, opened 2016-07-19
live map for mobile

cprogrammer1994:
This pull request includes a

- [x] New feature

`live-map.py` uses the running `example.py` as the backend.

`live-map.py` is currently for Python 2.7
**Pokemons** are visible on the map, they appear and disappear automatically. (*NEED NO REFRESH*)

- One can click on the Pokemons to see the countdown in seconds. (this should be improved)
- Clicking on multiple Pokemons will only open **1** **InfoWindow**
- Countdown is visible on high DPI devices as well.
- using **setTimeout** and **ajax** (cause only traffic if the browser is active)

### DIFF
--- a/live-map.py
+++ b/live-map.py
@@ -118,10 +118,16 @@ html = '''
 class Handler(SimpleHTTPServer.SimpleHTTPRequestHandler):
 	def do_GET(self):
 		if self.path == '/':
+			self.send_response(200)
+			self.send_header("Content-type", "text/html")
+			self.end_headers()
 			self.wfile.write(html % config)
 			self.wfile.close()
 
 		if self.path == '/rawdata':
+			self.send_response(200)
+			self.send_header("Content-type", "application/json")
+			self.end_headers()
 			self.wfile.write(urllib2.urlopen(DATA_URL).read())
 			self.wfile.close()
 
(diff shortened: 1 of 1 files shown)


Bitti09:
@cprogrammer1994 
Error:
Google Maps API warning: NoApiKeys https://developers.google.com/maps/documentation/javascript/error-messages#no-api-keys

cprogrammer1994:
The **Google Maps API** key is taken from the `credentials.json`

Bitti09:
Now it works as expected for me :+1:

SophisticaSean:
@cprogrammer1994 I was not able to get this to work, it just serves a webpage with the body being the html text.

live-map.py
```diff
@@ -110,18 +109,24 @@
 			reload();
 		}
 	</script>
-	<script src="https://maps.googleapis.com/maps/api/js?key=%(gmaps_key)s&callback=initMap" async defer></script>
+	<script src="https://maps.googleapis.com/maps/api/js?callback=initMap" async defer></script>
```
Bitti09: removing the apikey  ?

cprogrammer1994: things messed up when created a new branch for another Pull Request.

PR #384, opened 2016-07-19
Bugfix for endpoint  /next_loc?lat=&&lon=    Add new endpoint /location?loc= for text location request

actondong:
This pull request includes a

- Bug fix
- New Feature

The following changes were made

- update origin_lat/lon at beginning of each new search if FLOAT_LAT/LONG updated(this way of fixing consistent with #303 )
- new endpoint added for text location request /location?loc=

If this is related to an existing ticket, include a link to it as well.

Issue #355, opened 2016-07-19
Show device location


Jared87:
It would be great if the map could show the position of my used device on the map. 
Also, a "Reset base position to my current position"-Feature would be nice.

mkody:
+1 for the "reset".
For the first suggestion, see #164

PR #303, opened 2016-07-19
 localized map for English and German & fixed next_loc (#314)

chuparCh0pper:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [x] Translation

The following changes were made

- localized map for English and German
- If selected language is not available it defaults to English

If this is related to an existing ticket, include a link to it as well.

Issue #164, opened 2016-07-18
Get location from browser


JakeCooper:
It would be fantastic if we had the option to get the location from the browser.

CarlTaylor1989:
This for future development, this feature will get implemented.

KonstantinKuklin:
+1

volksport:
This is the killer feature for quick swapping between the app and this running on a personal server. I wouldn't ignore it for too long!

Darkrift411:
I would really prefer a choice to either get current location, or to tap a spot on the map (long press? right click?) to select the option to "travel here". If a short distance, it could walk there. If longer than say a block, it could "drive" there (45-60mph). I am always worried about changing the script to a far off location so I normally do it in steps or stop it for the time I think it takes to get there. I understand this would be harder, but maybe implement in portions.

Scyntrus:
There is actually a web API to change the location while the server is running, it looks like this:

localhost:5000/next_loc?lat=38.897660&lon=-77.036551

And the next time the server updates, it will move to that location.
However the larger your -st value, the longer it will take to move to the new location.

Darkrift411:
Nice. I stopped browsing the code when so many people took this up or I might have noticed that. Thanks for the input. I was writing a bash script to start and stop this (before refresh was implemented) and it was going to read from a coords file that could be updated via a web form. As soon as I start to think of implementing something someone beats me to it so I just decided to make requests instead.

PR #390, opened 2016-07-19
Add current location marker to the map

rnarian:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Add current location marker to the map

bildschirmfoto 2016-07-19 um 18 09 36


rnarian:
These issues might be relevant: #164, #355

volksport:
@rnarian any way you want to have this optionally send the coords to the next_loc api endpoint and have a parameter in the server startup that turns it on/off?

rnarian:
@volksport I'm more of a front-end guy. I'm honestly not really sure what you mean by that. 😔

sgraut:
Chrome is stating that _getCurrentPosition()_ is deprecated on insecure (http) origins. Does not work for me - neither on my mac nor on my phone

volksport:
There is a jquery geoloc plugin that may handle cross-browser issues more friendly? Never used it though.

sgraut:
The jquery.geolocation plugin does not work either. Feels like it is just a wrapper for the html5 geolocation functions.

SophisticaSean:
@sgraut Chrome doesn't allow http location gathering, so you'd have to either open it in Firefox or Safari or setup SSL termination on the server you're running it on. (I think running it on localhost:someport will work too)

rnarian:
Ok, it seems that there is no work-around in Chrome 50+ (other than running the app on https). But this doesn't change the fact that it's working in mobile safari which it was intended for. For people using the affected versions of Chrome the feature simply doesn't appear, but doesn't break any other functionality.

mkody:
Welp, using Let's Encrypt and a reverse-proxy, it works at least :+1:

volksport:
If this gets merged in I can add a param flag to have it update the next_loc endpoint automagically (or i might make it a button we'll see).

rnarian:
@volksport Awesome! But maybe make it optional because this only works if only one person is using the app. In my case there are 3 people using the same app so we would override each others current location.

rushingfitness:
I'm a rails dev, no Python, so this isn't my forté. But what I would do is have a geo flag, if the flag is present do two things. 
1.) In example.py on the location settings, check to see if there is a geolocation set in the session (whatever variable you want). If not sleep for 5 seconds (or whatever your refresh time is set as, may also need to make that flag required if geo flag is set). Then when the session geolocation variable is present, use that.
2.) On the view when you call `updateLocation()`, have it post a req to the server with the coords, which on the server side saves it as a session variable.

Idk if you can get the session varaible from the example.py file or not. But if so, I'd say this should work fine.

nekitk:
to add next_loc support for switching search center you just need to replace popup content with
`content: 'Current location<br><a href="/next_loc?lat='+position.coords.latitude+'&lon='+position.coords.longitude+'" target="_blank">Search from here</a>'`
but I'm not sure if `next_loc` itself is working correctly now

rushingfitness:
Ahhh! Interesting didn't see that. Well then if next_loc is working, could just add that as an ajax request in the updateLocation function.

volksport:
I think #384 or #303 will fix next_loc

Need to merge in a dozen of these PR before it makes sense to try to work on creating an auto geo updater

rushingfitness:
👍 Oh no doubt. Do you homie. Just trying to get it primed now while you're busy with the other PR's.
Also I'm going to look into making this a wearable apk if possible.

rnarian:
To clarify: I did not raise this PR to support updates of the backend location calls – instead I just wanted to add an indicator of my current position when I have the map running on my iPhone while hunting for Pokemon. 

I don't wanna be the party pooper so please don't get me wrong but I think the auto update feature of the backend calls should be discussed (and added) in a separate PR. Maybe we can focus on the initial PRs intention?

rushingfitness:
@rnarian Yes it would be a separate PR. We're just talking right now. No harm in talking things out :)

rushingfitness:
@rnarian p.s. I can't really get this to work on my mac. It works every once in a while, but when it does (blue dot showing instead of red arrow), no pokemon show up, or if they do it's like 1. If it just shows the red arrow like usual then all of them show up (> 20 pokemon)

rnarian:
@rushingfitness The pokemon don't show up because the servers are down right now.

rushingfitness:
@rnarian this was happening a couple of hours ago. But let me re-pull it down and see if it works (my connection to the servers is fine fyi)

rnarian:
@rushingfitness I just noticed the current location indicator didn't show up on the initial page load. I added another commit to fix this. Can you try again? Both the start position (red indicator) and the current location (blue indicator) should show up right away:

image

rushingfitness:
@rnarian confirmed the icon shows up right away. Can't do any more until the servers come back.

rnarian:
@rushingfitness Right now its working for me if you wanna give it another try. 😁

rushingfitness:
Just got it back. Confirmed. Looks good now. Thanks!

PR #403, opened 2016-07-19
Put script inside <body> tag

jangrewe:
This pull request includes a

- [ x] Bug fix
- [ ] New feature
- [ ] Translation

- move <body> end tag to after scripts


MikeDX:
I think the window.update() be in a closure, so it doesn't try to fire before the page loads and cause a (potential) infinite loop on a fast refresh setting.

PR #419, opened 2016-07-19
Fix for #'s 243, 313, 319 and 405 (same issue)

blakebjorn:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

 - Forces LABEL_TMPL and pokemon_key to ASCII characters
 - ♂ and ♀ are represented as (M) and (F), respectively
 - Changes on lines 823, 832, 843 in example.py
 - Not sure if it breaks existing functionality (program was unusable before making the changes)

AHAAAAAAA#313
AHAAAAAAA#319
AHAAAAAAA#405

### DIFF
--- a/example.py
+++ b/example.py
@@ -821,6 +821,7 @@ def get_pokemarkers():
 
     for pokemon_key in pokemons:
         pokemon = pokemons[pokemon_key]
+        pokemon['name'] = pokemon['name'].replace(u'\N{MALE SIGN}','(M)').replace(u'\N{FEMALE SIGN}','(F)')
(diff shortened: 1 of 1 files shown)


MikeDX:
Those markers worked (for me at least) so this much be a locale issue. I think it would be nicer to keep those symbols if possible as they look "nicer"

Not sure how you could go about checking that though since I've never seen the errors regarding the symbols

JZ6:
good change, might ask you to rebase into the rewritten file later, check https://github.com/AHAAAAAAA/PokemonGo-Map/tree/py-rewrite branch for now

Issue #359, opened 2016-07-19
Lured Pokéstops do not disappear once finished 


NitriKx:
### Environment info
Operating System: Centos7 

Installed version of Python: py27

### Steps to reproduce
1. Run the server with PokéStop + Only Lured PokéStop
2. Find a Lured POkéStop
3. Wait until it ends, the icon does not disappear

Temetz:
Can confirm.

mkody:
Confirming too.

AHAAAAAAA:
shouldn't be a terribly hard thing to fix since we've already done this with pokemon.

IMNoob1:
you have it refreshed ? (haven't noticed if gone with refreshed but can someone confirm?)

DerScreever:
Im sorry, but we dropped the support for the master branch (example.py). 
This branch is deprecated.  
The new branch contains a much cleaner structure so pls report only issues related to the develop branch. 

Your development team

PR #421, opened 2016-07-19
Fix for lured stops not disappearing and add expiration timer

Scyntrus:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

- Fixes https://github.com/AHAAAAAAA/PokemonGo-Map/issues/359
- Set disappear_time for lured pokestops to lure expiration time instead of -1.
- Store pokestop data regardless of onlylured flag to prevent stale pokestop data from being delivered to the client.
- Instead, filter unlured pokestops right before sending data to the client.
- Add check for expired pokestops alongside check for stale pokemon.
- Caveat: During the time after a lured pokestop expires and the pokestop is refreshed, no icon will be displayed on the client.

- Add a countdown timer to lured pokestops in the same way that countdown timers are added to pokemon.

### DIFF
--- a/example.py
+++ b/example.py
@@ -671,14 +671,17 @@ transform_from_wgs_to_gcj(Location(Fort.Latitude, Fort.Longitude))
 
                             elif Fort.FortType \
                                 and args.display_pokestop:
+                                expire_timestamp = 0
                                 expire_time = 0
                                 if Fort.LureInfo.LureExpiresTimestampMs:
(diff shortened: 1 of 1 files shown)


Scyntrus:
For whatever reason it says lures expire in 2 minutes

Scyntrus:
I don't particularly like the way I implemented this.

PR #426, opened 2016-07-19
Fix for lured stops not disappearing

Scyntrus:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Fixes https://github.com/AHAAAAAAA/PokemonGo-Map/issues/359
- Do not set disappear_time for lured pokestops, the new default expiration time is the autorefresh interval + 500ms.
- Store pokestop data regardless of onlylured flag to prevent stale pokestop data from being delivered to the client.
- Instead, filter unlured pokestops right before sending data to the client.
- Caveat: During the time after a lured pokestop expires and the pokestop is refreshed, no icon will be displayed on the client.
- Caveat: Timer still wrong

### DIFF
--- a/example.py
+++ b/example.py
@@ -671,14 +671,17 @@ transform_from_wgs_to_gcj(Location(Fort.Latitude, Fort.Longitude))
 
                             elif Fort.FortType \
                                 and args.display_pokestop:
+                                expire_timestamp = 0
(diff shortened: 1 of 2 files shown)


Scyntrus:
Conflicts with newer commits

PR #439, opened 2016-07-19
Allow login from token

mchristopher:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

I added the ability to send a token from an outside login process for an Electron desktop app I'm working on that packages this project into an easy interface.

I also loaded jQuery from the file on-disk and cleaned up some of the HTML. More to come on that front...


Macarse:
@mchristopher are you using this to do a google sign in?

mchristopher:
@Macarse Yes, using the Google OAuth web flow. That way it handles 2FA correctly.

Issue #357, opened 2016-07-19
Fork for each steps to parallel API calls


slbug:
It will be useful to have parameter to make parallel requests.
Fork can be done near - ```for step in range(steplimit2):```

irrenhaus:
I think this will only add to the chance that your account gets banned since this would locate you in multiple locations at the same time.

slbug:
Multiple accounts?

cprogrammer1994:
Personally I have to wait 15 seconds to get my map full of pokemons... This is bearable.

slbug:
With 5 steps I have to wait about 10 minutes for full refresh

hgentry:
I can implement this tonight with a -parallel flag to turn it on

PR #456, opened 2016-07-20
Multithreaded Pokemon searching

hgentry:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

I've added support for multithreaded requests as suggested in #357, which greatly speeds up the process. I have 15-step maps running in the time a 6-step map took before multithreading. To run with multithreading, add -mt to your arguments.

### DIFF
--- a/example.py
+++ b/example.py
@@ -40,7 +40,7 @@ APP = 'com.nianticlabs.pokemongo'
 
 with open('credentials.json') as file:
 	credentials = json.load(file)
-
+mt = False
 PTC_CLIENT_SECRET = credentials.get('ptc_client_secret', None)
 ANDROID_ID = credentials.get('android_id', None)
 SERVICE = credentials.get('service', None)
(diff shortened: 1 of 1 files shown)


example.py
```diff
@@ -471,6 +471,12 @@ def get_args():
         help='Set web server listening host',
         default='127.0.0.1')
     parser.add_argument(
+        '-mt',
+        '--multithreading',
+        help='Enable multithreaded search',
+        action='store_true',
+        default=False)
```
juharris: Why not default to `True`?

hgentry: Just in case there's a major stability issue I didn't catch. I don't want to break everyone's maps until this has been used more.

medghaim:
This is awesome! But unfortunately the entire project is being rewritten, example.py won't exist soon. Check out the rewrite-branch, as per the readme.

hgentry:
I look forward to contributing to the new project! Is there an ETA?

PR #465, opened 2016-07-20
fix script failing after 30 min or so problem

foodforarabbit:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- need to set TimeTillHiddenMs even if lower or else seen[hash] will only increase until it can't anymore
- hide "expired" pokemon from front end
-

### DIFF
--- a/example.py
+++ b/example.py
@@ -655,9 +655,9 @@ def process_step(args, api_endpoint, access_token, profile_response,
             for cell in hh.cells:
                 for wild in cell.WildPokemon:
                     hash = wild.SpawnPointId;
-                    if hash not in seen.keys() or (seen[hash].TimeTillHiddenMs < wild.TimeTillHiddenMs):
-                        visible.append(wild)
-                        seen[hash] = wild.TimeTillHiddenMs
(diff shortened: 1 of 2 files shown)


medghaim:
The entire project is being rewritten, `example.py` won't exist soon. Check out the rewrite-branch, as per the readme.

bluefi:
Isn't there still a problem at this lines?:

```
                    if hash not in seen.keys() or (seen[hash].TimeTillHiddenMs <= wild.TimeTillHiddenMs):
                        visible.append(wild)    
                    seen[hash] = wild.TimeTillHiddenMs
```

Never wrote a line python, but seen[hash].TimeTillHiddenMs does not exist if you set seen[hash] to wild.TimeTillHiddenMs, right? Either set seen[hash] = wild;  or compare seen[hash] <= wild.TimeTIllHiddenMs

Josephvb10:
Even with the fixes listed here, the problem is still present for me with auto refresh on. After 30 minutes or so Pokémon stop appearing in the map.

Herwi:
Same here. Still after 30 minutes or so Pokémon stop appearing in the map.

KingScooty:
This issue is still present.

PR #341, opened 2016-07-19
Display additional information for visible Pokemons in console

NgJason:
This pull request includes a:
- [x] New feature/ Enhancement

Display Pokemon name, coordinate, ID and remaining time for any discovered Pokemons. This is useful for users who do not need to look at the map. With the output, users can copy and paste the coordinate into any external program/gps.

Usage: 
Add the flag -dc

Ex Output in console:
<p>
===============================<br>
 Name: Zubat<br>
 Coord: (37.777523,-122.400932)<br>
 ID: 41<br>
 Remaining Time: 00m23s<br>
===============================
</p>

PR #470, opened 2016-07-20
Alert users with a sound when a Pokemon is visible

NgJason:
- [x] Feature

Alert users when a Pokemon is discovered by playing the audio file: a.wav. 

This is especially useful when a user is multitasking with the -only flag.

Usage: 
pip install --upgrade -r requirements.tx
To activate audio for console, add -s flag.
To activate audio for web browser, add -sw flag.


Josephvb10:
I think this would be better if implemented in the HTML client, not in the Python script.

NgJason:
@Josephvb10 I run the program without opening any browsers. Of course I have the Pokemon's information printed out in the console. You can see that functionality in the following pull request: #341

NgJason:
Anyone know of a way to import libraries via Travis CI so it can properly run the test?

NgJason:
Travis-ci has issues with pip install pyaudio.

frdmn:
I see your use case @NgJason, but I also agree with @Josephvb10 - I use the map with my friends so we just have the browser running and the server is an EC2 instance. 

I think it should be configurable then, if either the client or the server plays the alert sound.

NgJason:
@frdmn I added the feature to play audio via web browser.  If you don't mind, please try it out and give me your feedback. 

If your EC2 is running Linux, it might require additional steps to install the pyaudio library. Travic-ci failed when it tried to compile pyaudio.

frdmn:
Works great for me (tested only `-sw`). Nice work.

Regarding your failed Travis test case, perhaps [the shared audio libs are missing](http://stackoverflow.com/a/35708669/2940412)?

guest136:
Hello, I tried that flag but it isnt working by me or am I doing anything wrong. I added -sw to my command like this:

python example.py -a google -u ***@gmail.com -p *** -l "40.739663 -73.992102" -st 10 -ar 3 -sw

PR #472, opened 2016-07-20
Feature/update pokemon to current location

rushingfitness:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Pulled in update location from @rnarian
- Pulled in fix for next_loc
- Push the updated location to next_loc

Depends on:
 - https://github.com/AHAAAAAAA/PokemonGo-Map/pull/390
 - Used some code from https://github.com/AHAAAAAAA/PokemonGo-Map/pull/384


frdmn:
Why is #390 inside this PR? This should be separate!

rushingfitness:
Because it relies on that code. And once it gets merged in, this should be looked at. He stated that his intention when he wrote the code wasn't to do this. But he's not saying we shouldn't do this. 
Besides, is this a feature that we're not going to use because he said he didn't intend on doing that when he wrote it?
Once his gets merged in, git won't have any conflicts. Gits pretty nice about that stuff. 
How would you have liked this code to be handled? Not written at all until his code got merged in? Write my own version of what he wrote so when he merges his in, this PR will have conflicts?

rushingfitness:
Yup the optional flag is a great idea. Good looks.

frdmn:
Please see my updated post. The two features are not related to each other. Even if it's the case that your code depend on his - In my opinion it's git best practise to wait for his to merge and then raise yours once you have "productional" commit hashs to work with (perhaps the person who merges it will squash them? perhaps they will get picked?). Otherwise there will be conflicts nonetheless and the commit history will look fucked up.

Also please see my updated post above, you can not use the map on multiple clients anymore now, since they would override each other.

rushingfitness:
Actually they are related to each other. It's using his geolocator code to push the resulting location to the server and update the location. 
And yes that is your opinion, but in my opinion it's a waste of time waiting for another or to be merged in before even writing the code. If something gets rebased, it's pretty easy to fix it on my end and repush the updated code to this PR. Again the magic of git. 
It's better to have the code out and have people discussing it, than it sitting locally and me just twirling my thumbs. 
As for your last comment, I'll wait until the refactoring has been completed to make any changes. (See this is why we push PRs,to see if there are things missing, bugs etc, literally this right here.)

frdmn:
Alright 👍

So it would be nice if you could make it optional or disabled by default, because otherwise if this PR gets merged instead of the original (#390) one, it would break the functionality for me (or better said my friends) since we use one server and multiple clients connected to that.

PR #492, opened 2016-07-20
POSIX script run helper

iAbadia:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Configurable shell script run helper added, strict POSIX compliant.

### DIFF
--- a/runserver.sh
+++ b/runserver.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+
+# Fill this variables with your desired options
+loc="$1"                             # Location
+auth=""                              # PTC or google
+user=""                              # User
+passwd=""                            # Password
+steps=5                              # Number of steps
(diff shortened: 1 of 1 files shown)


frdmn:
I feel like this is not necessary. 

The argument parser already does a great job listing all available options and passing arguments should not be that complex that someone need a wrapper around it.

iAbadia:
10+ args command vs 1 is the point, I find it useful myself @frdmn

sykhro:
We will have a config file at some point in the future

frdmn:
@iAbadia But in your wrapper a lot of the possible options are hardcoded or not even checked. For example the `dg` and  `i` option. And in the future if there are new arguments, the wrapper always has to get updated as well.

sykhro:
Closing since external configuration parameters will be implemented soon

PR #503, opened 2016-07-20
Restart scan after first loop

AHAAAAAAA:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

The following changes were made:-

Scan would loop once then stop scanning. This makes it loop continuously until stopped.


DerScreever:
+1 
If you say it work we can merge.

PR #513, opened 2016-07-20
Credentials.json for custom api keys.

medghaim:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Credentials.json reintroduced.
- introduced pogom.exceptions


DerScreever:
maybe add some error message that the key is not defined in the credentials.json?

medghaim:
Good idea, I'll get on that.

DerScreever:
we should also remove that hard coded key... only make problems.

Ashex1:
Don't think pogom.db was meant to be included ;)

pogom/app.py
```diff
@@ -18,11 +18,13 @@ def __init__(self, *args, **kwargs):
         self.route("/pokemons", methods=['GET'])(self.pokemons)
         self.route("/gyms", methods=['GET'])(self.gyms)
         self.route("/pokestops", methods=['GET'])(self.pokestops)
+        self.gmaps_key = None
```
Ashex1: This isn't needed.

medghaim: ooooops

medghaim:
a _very_ basic check to see if user has inputted gmaps_key, which is now empty in credentials.json, forcing users to get their own API key.

An exception may not be the greatest way to deal with it, but it's short and to the point.
image

DerScreever:
@medghaim 
maybe you should do something like this:
config['GMAPS_KEY'] = load_credentials(os.path.dirname(os.path.realpath(__file__)))['gmaps_key'] 
if config['GMAPS_KEY'] = null
       return
app.run(threaded=True, debug=args.debug, host=args.host, port=args.port)

i think we dont need to run the program, when we have no api key ^^
This is my opinion.

medghaim:
That's _pretty_ much what's happening, but the check is in load_credentials, as is the exception raised.

pogom/utils.py
```diff
@@ -80,3 +81,12 @@ def get_pokemon_name(pokemon_id):
             get_pokemon_name.names = json.loads(f.read())
 
     return get_pokemon_name.names[str(pokemon_id)]
+
+def load_credentials(filepath):
+    with open(filepath+'/credentials.json') as file:
+        creds = json.load(file)
+        if not creds['gmaps_key']:
+            raise APIKeyException('No Google Maps API key entered.')
```
Ashex1: Just nitpicking but it'd be good to specify the Javascript API Key. `No Google Maps Javascript API key entered`

DerScreever:
ahh didn't saw your new commit.
Approved.
+1

pogom/utils.py
DerScreever: then you could also write. have a look on wiki for instruction how to generate this key.

medghaim: Fixed :)

Issue #10, opened 2016-07-17
AttributeError


malicious50:
Getting this error:
![cattura](https://cloud.githubusercontent.com/assets/11940908/16901104/bb09eb3c-4c39-11e6-8a3a-3c48848c65da.PNG)

dosenhosenjoe:
thats because pokemon club is down

BrodrickChilds:
Ah I was wondering if they'd updated the API to not include heartbeat to discourage projects like this one... Good to know that's not the case :P

mattmmeloan:
Yeah I think that's the case too. Just needs better exception handling.

marauder37:
Code was just merged to fix this

malicious50:
Works. Closing this.

malicious50:
That problem seems fixed, even if getting a lot of
[-] retrying_api_req: request error (('Connection aborted.', BadStatusLine("''",
))), retrying
[-] retrying_api_req: request error (('Connection aborted.', BadStatusLine("''",
))), retrying
[-] retrying_api_req: request error (('Connection aborted.', BadStatusLine("''",
))), retrying
[-] retrying_api_req: request error (('Connection aborted.', BadStatusLine("''",
))), retrying
[-] retrying_api_req: request error (('Connection aborted.', BadStatusLine("''",
))), retrying
[-] retrying_api_req: request error (('Connection aborted.', BadStatusLine("''",
))), retrying
[-] retrying_api_req: request error (('Connection aborted.', BadStatusLine("''",
))), retrying

now, but I guess because of servers overload. Thanks!

PR #519, opened 2016-07-20
Added pokemon artwork

POPSuL:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added pokemon artwork into tooltips

#328


DerScreever:
Approved.
+1

frdmn:
We should store those images locally and not penetrate the `pokemon.com` servers IMHO.

POPSuL:
@frdmn imho in this case we should store all data locally include page about pokemon (like http://www.pokemon.com/us/pokedex/010, it is a link `#010` in tooltip). But, it is not neccessary.

POPSuL:
I can download all images if it really necessary. It is really necessary?

rnarian:
I would think so too. You have to keep in mind that the map is used by thousands of people. This might generate plenty of traffic on the pokemon.com servers.

frdmn:
@POPSuL Well, right now we don't parse those data so I don't see the need to store them locally. But since the artworks are embedded - all the clients will actually fire tons of GET requests to the `pokemon.com` servers. The other one is just a text anchor that points to their website (which is perfectly okay).

This was just my personal opinion, but considering the growth of this repository based on the GitHub stars the past few days, I could imagin the hotlinked traffic will be excessive just for those artworks. But hey, let's wait for some input of the other guys. Maybe the majority sees this in a different light.

PhyberApex:
I'm on the same side as frdmn! Hotlinking traffic might go trough the roof

sykhro:
Nintendo has planty of servers and a wide CDN, I'm sure they can handle those requests just fine.

frdmn:
That's not a reasonable explanation which would justify the hotlinking!

POPSuL:
I added 147 pokemons which already available in pogo. If niantic adds new pokemons to the game - I will add new images as soon as I can.

sykhro:
If you can take care of this, perhaps a git module to a dedicated repo would be better. This would allow for use in other projects and have the source of images centralized

POPSuL:
@sykhro it is a great idea! But, I dont know how does work "Download ZIP" on github. If I add git submodule - user which downloaded master branch as ZIP archive (soon, when develop branch will be merged into master) will have all subrepos? 
Just a lot of users download the project as archive of master branch and trying to deploy locally as is...

frdmn:
Since we now have a `package.json` (for Grunt etc.) we could simply register a NPM package and publish via NPM updates.

IMHO that is way more comfortable than handling git's submodules.

POPSuL:
@frdmn if we have dependency of NPM we will have a tarball generator and in this case NPM module, python module or git subrepo are absolutely identical for script for tarbal generate, IMHO

rnarian:
I think you are misunderstanding.

No need to zip/unzip anything. The images would sit in a separate repo, published via npm. In the PokemonGo-Map repo we could then require the npm module as a dependency and copy them from `node_modules/*` to `/static/*` via grunt afterwards.

If new images are added to the npm module all it takes is another `npm install && grunt`.

frdmn:
@POPSuL Awesome, thank you! 👍

hhu94:
I like it!

MonkeyLeeT:
Since this will be in separate repo now, closed and appreciated your effort.

PR #534, opened 2016-07-20
delete pogom.db

alexpantyukhin:
It's a little annoying to check conflicts when pogom.db already exists. Anyway it is can be created annytime when create_tables() called


DerScreever:
dont delete that file. just add it to .gitignore. ^^

alexpantyukhin:
in this way I needed to update gitignore :)

Issue #510, opened 2016-07-20
Missing features develop branch - Overview


DerScreever:
# Features to be re-implemented
Just a little overview for the missing features on develop branch. 
Feel free to report missing features here.
If you want to reimplement some feature, post here please so everyone see that this is already in progress.

## List
- [x] Re-implementing a credentials/config.json | _Done_ | by @medghaim 
- [x] Re-implementing the 'ignore' and 'only' parameters  | _In progress_ | by @hhu94 
- [x] Auto-refresh | _in progress_ | by @Lunat1q
- [x] Display pokéstops | _in progress_ | by @invisiblek 	
- [x] Display gyms | _in progress_ | by @invisiblek 		
- [x] Locale for Pokemon names | _In progress_ | by @Scildrei
- [x] Coordinates transformer for China	
- [x] Debug mode	
- [x] Display only lured Pokestops
- [x] Route /next_loc (/next lat/lon variables)
- [x] Route /raw_data (output in this format: { "gyms": {}, "pokemons": {}, "pokestops": {} }) 
- [ ] Filter in webapp (toggle pokemon, pokestops and gyms display)

hhu94:
I'll do the ignore and only parameters.

PhyberApex:
raw_data location is also a feature I would really appreciate if brought back. Having this allows for a lot of flexibility with third party programms!

Scildrei:
I added the locale for Poke names. I'll create a pull request asap.

alexpantyukhin:
I could try `Display gyms` if it's not urgent task

Rohmin:
- heroku compatiable
- next_loc /next lat/lon variables

DerScreever:
@Rohmin 
i think heroku was never merged on master.

Lunat1q:
i'll try to do auto-refresh.

Rohmin:
@DerScreever There was a pull request for it, if it could be done in dev branch would be awesome. 
Link to it : https://github.com/AHAAAAAAA/PokemonGo-Map/pull/449

Also for  request /location?loc=

https://github.com/AHAAAAAAA/PokemonGo-Map/pull/384

DerScreever:
@Rohmin 
like you told there was a pull request. but we never merged it. 
so first steps is to get 100% of the master functions then we can do something like this.

#384 need to be done again because of the new structure.

invisiblek:
[...]

PR #563, opened 2016-07-20
Add gym icons and info

invisiblek:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

-Add gym_points to model
-Pull in gym info via json
-Shove info newly-created label

https://github.com/AHAAAAAAA/PokemonGo-Map/issues/510


invisiblek:
One thing - I did have to nuke my db or I was getting 500 errors (the gym_points field didn't exist in the database). We may need to come up with a method of handling this (db versions/migrations, willy-nilly nuking the db if things go wrong, etc etc)

PR #573, opened 2016-07-20
Docker File for building static image

Ashex1:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Docker File for building image

Unlike #317 which is designed to pull the latest code on every run/restart this is better suited for building an image that is a snapshot of the repository.

I'm providing this PR for consideration, would be good to incorporate into Travis for triggered builds to docker hub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ MAINTAINER Ahmed Osman (/u/Ashex)
 
 EXPOSE 5000
 
-RUN apk add --update openssl ca-certificates
+RUN apk add --update ca-certificates
 RUN apk add \
     python \
     python-dev \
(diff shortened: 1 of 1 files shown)


MonkeyLeeT:
Please take a look at #599 and have a decision.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+EXPOSE 5000
+
+RUN apk add --update ca-certificates
+RUN apk add \
+    python \
+    python-dev \
+    py-pip \
+    build-base \
+  && rm -rf /var/cache/apk/*
```
ultrafez: I'll add this to my `Dockerfile` since I imagine it'll shrink the image size somewhat!

PR #599, opened 2016-07-20
Add Docker support

ultrafez:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added `Dockerfile` to allow running in a container

The `credentials.json` file isn't included, so specifying a Google Maps key on the CLI using the `-k` argument is necessary.

This PR is based on my original PR #309, updated to point at the `develop` branch.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.github
+.gitignore
+credentials.example.json
+node_modules
+Easy\ Setup
+Gruntfile.js
+package.json
+pogom.db
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,6 @@
 
 FROM python:2.7-alpine
 
-# ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
-# build-base contains gcc, which is needed during the installation of the pycryptodomex pip package
-RUN apk add --no-cache \
-    ca-certificates \
-    build-base
-
 # default port the app runs on
 EXPOSE 5000
 
@@ -18,7 +12,16 @@ WORKDIR /usr/src/app
 
 ENTRYPOINT ["python", "./runserver.py"]
 
+# Copy Python requirements so we only rebuild deps if they have changed
 COPY requirements.txt /usr/src/app/
-RUN pip install --no-cache-dir -r requirements.txt
 
+# ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
+# build-base contains gcc, which is needed during the installation of the pycryptodomex pip package
+RUN apk add --no-cache \
+    ca-certificates \
+    build-base \
+ && pip install --no-cache-dir -r requirements.txt \
+ && apk del build-base
+
+# Add the rest of the app code
 COPY . /usr/src/app
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,8 @@
+# Basic docker image for PokemonGo-Map
+# Usage:
+#   docker build -t pokemongo-map .
+#   docker run -d -P pokemongo-map --host 0.0.0.0 -a ptc -u YOURUSERNAME -p YOURPASSWORD -l "Seattle, WA" -st 10 --google-maps-key CHECKTHEWIKI
+
 FROM python:2.7-alpine
 
 # ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,19 @@
-FROM python:2.7-onbuild
+FROM python:2.7-alpine
+
+# ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
+# build-base contains gcc, which is needed during the installation of the pycryptodomex pip package
+RUN apk add --update ca-certificates build-base
+
+# default port the app runs on
 EXPOSE 5000
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
 ENTRYPOINT ["python", "./runserver.py"]
+
+COPY requirements.txt /usr/src/app/
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . /usr/src/app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,10 +7,9 @@ FROM python:2.7-alpine
 
 # ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
 # build-base contains gcc, which is needed during the installation of the pycryptodomex pip package
-RUN apk add --update \
+RUN apk add --no-cache \
     ca-certificates \
-    build-base \
-    && rm -rf /var/cache/apk/*
+    build-base
 
 # default port the app runs on
 EXPOSE 5000
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM python:2.7-onbuild
+EXPOSE 5000
+ENTRYPOINT ["python", "./runserver.py"]
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,11 @@
 
 FROM python:2.7-alpine
 
+# keep this near the top of the file since we'll always need this
+# ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
+RUN apk add --no-cache \
+    ca-certificates
+
 # default port the app runs on
 EXPOSE 5000
 
@@ -15,10 +20,8 @@ ENTRYPOINT ["python", "./runserver.py"]
 # Copy Python requirements so we only rebuild deps if they have changed
 COPY requirements.txt /usr/src/app/
 
-# ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
 # build-base contains gcc, which is needed during the installation of the pycryptodomex pip package
 RUN apk add --no-cache \
-    ca-certificates \
     build-base \
  && pip install --no-cache-dir -r requirements.txt \
  && apk del build-base
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,6 @@ RUN apk add --update ca-certificates build-base
 # default port the app runs on
 EXPOSE 5000
 
-RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
 ENTRYPOINT ["python", "./runserver.py"]
@@ -16,4 +15,3 @@ COPY requirements.txt /usr/src/app/
 RUN pip install --no-cache-dir -r requirements.txt
 
 COPY . /usr/src/app
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,10 @@ FROM python:2.7-alpine
 
 # ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
 # build-base contains gcc, which is needed during the installation of the pycryptodomex pip package
-RUN apk add --update ca-certificates build-base
+RUN apk add --update \
+    ca-certificates \
+    build-base \
+    && rm -rf /var/cache/apk/*
 
 # default port the app runs on
 EXPOSE 5000


MonkeyLeeT:
#573 

Can you guys have a talk and decide which one?

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+# default port the app runs on
+EXPOSE 5000
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+ENTRYPOINT ["python", "./runserver.py"]
+
+COPY requirements.txt /usr/src/app/
```
Ashex1: This is an unnecessary step since the entire project is being copied later, additionally COPY creates the directory path for you. Would be better like so:

```
COPY . /usr/src/app

WORKDIR /usr/src/app
RUN pip install --no-cache-dir -r requirements.txt 

ENTRYPOINT ["python", "./runserver.py"]
```

Ashex1:
@ultrafez In addition to my code comments, add the following usage at the top and I'm a happy camper.

```
# Basic docker image for PokemonGo-Map
 +# Usage:
 +#   docker build -t pokemongo-map .
 +#   docker run -d -P pokemongo-map --host 0.0.0.0 -a ptc -u UserName -p Password -l "Seattle, WA" -st 10 --google-maps-key CHECKTHEWIKI
```

ellisio:
The only other thing I see with this is ensuring the `gulp` assets are compiled. Obviously this can be done prior to `docker build`; but if you automate this via Docker Hub you'll need to have the following in the build process (not including installing NodeJS itself):

```
npm install -g gulp
npm install
gulp jshint sass cssmin uglify
npm remove -g gulp
rm -rf node_modules
```

I also recommend a new Gulp Task set is created _without_ the watch statement, or just remove `watch` from the `default` set.

Dockerfile
ultrafez: Copying the `requirements.txt` and installing the dependencies before copying the code is an efficiency step. If all you change between consecutive image builds is the code (not the requirements), then Docker is able to re-use all of the intermediate images apart from the last one where it copies the code - crucially, it doesn't have to re-run the `pip install` step. This means that most builds will be really quick.

If you copied all of the code and then ran `pip install` afterwards, then any code changes would mean that the code would have to be recopied, and `pip install` would have to be re-run. Check out [this best practices article from Docker](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy) for more info.

Good point about the unnecessary `mkdir` though - I'll fix it.

Ashex1: That's a really good point, I had assumed the cache would be invalidated due to `requirements.txt` being copied in the previous step.

mikedougherty: This is the correct and idiomatic way to re-use the docker cache for dependencies (i.e. to not `pip install` on every source code change). [Here's a straightforward explanation](http://tiborsimko.org/docker-for-python-applications.html#docker-build-cache).

ultrafez: @Ashex1 it used to work that way, that an ADD line would invalidate the cache - looks like they fixed that by hashing the file contents to see whether to use the cache or not. Very useful to improving build speeds!

ultrafez:
@ellisio Before it disappeared, I saw your comment about building assets with grunt - it seems like an important point, what made you delete your comment?

MonkeyLeeT:
Please comment if this is good to go :)

Dockerfile
```diff
@@ -7,7 +7,10 @@ FROM python:2.7-alpine
 
 # ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
 # build-base contains gcc, which is needed during the installation of the pycryptodomex pip package
-RUN apk add --update ca-certificates build-base
+RUN apk add --update \
```
Ashex1: We should actually be able to use `--no-cache` here so we don't need to supply `--update` and then delete the cache after.

ellisio:
@ultrafez Because the base repository includes distributed resources by default in `/static/dist`. So we probably don't need to rebuild them.

If these get removed and/or are not updated, then yeah, we'll want to build these assets during Dockerization.

sanderploegsma:
By the way, when the server encounters an exception, the process does not exit, it just freezes. It would be better if it exited, so that docker's `--restart=always` flag kicks in...

Ashex1:
@sanderploegsma That's a separate issue regarding how exceptions are handled. 

@ultrafez I think we can marge it as is (my comment about `--no-cache` isn't a showstopper). Thoughts?

ellisio:
Hold up, I'm working on a Dockerfile that you guys can use for reference in regards to `alpine:3.4`. It also includes some `apk`-fu to remove unused build dependencies after the image is built to decrease the final image size.

Need about 30 minutes.

ellisio:
Alright, here is a container that I have successfully deployed via Kubernetes in Google Container Engine (the reason for the `ping.html` is for the external load balancer check I have to use). BUt you guys can take a look at the `Dockerfile` to get an idea of how I built my container:

https://github.com/ellisio/dockerfiles/tree/master/pokemongo-map

The container image on Docker Hub is `90MB`.

This build also custom compiles the Grunt tasks, so if we don't need that for the project's actual Dockerization then you can remove all of the nodejs, npm, and grunt stuff.

chuyskywalker:
You can reduce this image a lot by doing it this way:

```
# Basic docker image for PokemonGo-Map
# Usage:
#   docker build -t pokemongo-map .
#   docker run -ti --rm -P --env-file .env pokemongo-map

FROM python:2.7-alpine

# default port the app runs on
EXPOSE 5000

WORKDIR /usr/src/app

# Copy this file first so that we only rebuild deps if they have changed
COPY requirements.txt /usr/src/app/

# Install build tools and ca-certs (so you can donwload things)
# Install pip deps
# Remove the build tools (don't need 'em, just take up space)
RUN apk --no-cache add ca-certificates build-base \
 && pip install --no-cache-dir -r requirements.txt \
 && apk del build-base

# Copy (merge) the rest of the code back in
COPY . /usr/src/app

# Allows the use of `--env` variables, or better yet, a `--env-file`
CMD /usr/local/bin/python example.py -a $AUTH_SERVICE -u $USER -p $PASS -l "$LOCATION" -st $STEP_LIMIT -H $HOST -P $PORT
```

And then having a `.env` file that looks like this:

```
AUTH_SERVICE=ptc
USER=username
PASS=password
LOCATION=whatever
STEP_LIMIT=3
HOST=0.0.0.0
PORT=80
```

MonkeyLeeT:
@ultrafez please review comments above and make changes as needed. Otherwise similar PR for docker support will be accepted since we're aiming for a functional release soon :).

Again thanks a lot for the input, everyone.

MichaelErmer:
Standalone Dockerfile:
```
# Usage:
#   docker build -t pokemongo-map .
#   docker run -ti --rm -P pokemongo-map -a ptc -u user -p pass -l "coords/location"

# Set the base image to Ubuntu
FROM ubuntu

# File Author / Maintainer
MAINTAINER Dracena

# Update the sources list
RUN apt-get update

# Install basic applications
RUN apt-get install -y tar git curl nano wget dialog net-tools
# build-essential

# Install Python and Basic Python Tools
RUN apt-get install -y python python-dev python-distribute python-pip

# Copy the application folder inside the container
RUN mkdir -p /usr/src/app

# Clone Git
RUN git clone https://github.com/AHAAAAAAA/PokemonGo-Map.git /usr/src/app

# Get pip to download and install requirements:
RUN pip install -r /usr/src/app/requirements.txt

# Expose ports
EXPOSE 5000

# Set the default directory where CMD will execute
WORKDIR /usr/src/app

# Run pogo map, just pass full command line to docker run command...
ENTRYPOINT [ "python", "/usr/src/app/example.py" ]
```

Ashex1:
@MonkeyLeeT If you reopen my PR I'll incorporate the changes @ultrafez and I discussed along with input from others. This would be done within the next couple hours.

Ashex1:
@MonkeyLeeT I've got all the feedback from everyone here in my fork. Not trying to steal @ultrafez  thunder, just giving an option should you want to get a release published before they come back and make the necessary changes.

ultrafez:
@MonkeyLeeT: I'll catch up and incorporate changes in a few hours once I've finished work - I think we're very nearly there. Including elements such a building Grunt deps can go into a separate later PR.

General notes I'll add now though:

* Dockerfiles shouldn't clone git repos - the Dockerfile should use the code that is already present locally to build an image of that code.
* There should be no unnecessary applications included in Docker images - a Docker image should be an isolated environment to run an application and only include direct dependencies of the application (i.e. no dev tools)

Ashex1:
I was mucking around with things because I'm a horrible procrastinator, we can remove these packages after the pip run.

`RUN apk del build-base python-dev py-pip`

ellisio:
FYI, the image I posted above is now down to `34MB` from `90MB` with all of the asset building still in place. Magic!

ultrafez:
I've taken all of your comments into account when making some optimisations to the Dockerfile. I tried to make a compromise between build speed and final image size - [this is a great article](https://kamilszymanski.github.io/optimizing-docker-images-for-image-size-and-build-time/) that explains how best to optimise Docker images for size and build time.

Having as few layers as possible, and doing the cleanup on each layer will reduce the final image size - however, combining more elements into the layer will mean it's more likely to need rebuilding when something changes, due to cache invalidation. @ellisio, with your Dockerfile [here](https://github.com/ellisio/dockerfiles/blob/9c7908d07869ed640a4c622b6cab1fe9ea091031/pokemongo-map/Dockerfile#L7), you get a really small total image size, however it doesn't take good advantage of Docker caching. When the code changes, Docker would invalidate line 6 where you add the code, meaning that line 7 (containing Python, Supervisor, bash, and Python deps) would have to be rebuilt. This would mean users would have to redownload nearly the entire image every time, rather than just the parts that have changed (most of the time that'll just be the code).

@Ashex1, unfortunately adding a line to remove `build-base`, `python-dev` and `py-pip` will actually increase the size of the image, since each layer of the image is immutable and so those packages are actually still present in one of the intermediate images, but they'd be hidden by the layer where you remove the packages.

I combined the lines that install the build tools with the line that installs the package requirements, based on the reasoning that most of the time the app deps won't change, and so that layer won't have to be redownloaded - most users will just have to redownload the final layer containing the app code.

The final image size is 107MB with my Dockerfile, which is made up of:
* 51MB for the base Python 2.7 image
* 35MB for Python package deps (+ `ca-certificates`)
* 15MB for app code
Most releases will only change the app code layer, meaning that although the initial download for users will be 107MB (unless of course they already have the Python 2.7 alpine image), they'll only have to download 15MB on subsequent updates.

ellisio:
So now the only thing remaining is how are you going to compile the JS and CSS? Are you going to be building this image manually, or expecting the developers to do it themselves and remember?

There needs to be documentation explaining that you need to install NodeJS and the NPM dependencies locally and then compile the JS and CSS via `grunt`. This would also mean you need to add a `.dockerignore` file to the repository that excludes the following at a minimum:

```
.github
.gitignore
node_modules
Easy\ Setup
Gruntfile.js
package.json
pogom.db
```

That `node_modules` will sky rocket your image size. :)

ultrafez:
I thought it might be better to avoid that issue with this PR so that we can get it merged, and then open another one to solve the issue of building frontend deps - this PR has gone on for a while now :)

ultrafez:
Is it worth adding `ENV` support with an entrypoint script as mentioned by @manuelmazzuola in #904?

davidkassa:
I say yes, it would allow easy use of docker cloud.

ellisio:
Yeah, we need some sort of entrypoint for sure. Especially with all of the configurable options.

Ashex1:
Since there will be a configuration file that can be used instead of arguments I'd suggest an entry point that supports replacing entries in the configuration file from env variables.

I've seen this practice in the wild utilizing a variable match. I did this for my pokemongo-map image for credentials.json, you can see the snippet in my gist here:

https://gist.github.com/Ashex1/862e9a200d58cf39a3418f7bf765e1eb

#968 will introduce the configuration file, it will be a YAML file so you'll need to adjust that bash function for yaml.

MonkeyLeeT:
@ultrafez Please inform us if this is ready to go. It's a bit headache to see more Docker PRs with this one still open.

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# Usage:
+#   docker build -t pokemongo-map .
+#   docker run -d -P pokemongo-map --host 0.0.0.0 -a ptc -u YOURUSERNAME -p YOURPASSWORD -l "Seattle, WA" -st 10 --google-maps-key CHECKTHEWIKI
+
+FROM python:2.7-alpine
+
+# keep this near the top of the file since we'll always need this
+# ca-certificates is needed because without it, pip fails to install packages due to a certificate failure
+RUN apk add --no-cache \
+    ca-certificates
```
jnovack: `ca-certificates` is not needed after the install to run.  It can be deleted after to save space.

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# default port the app runs on
+EXPOSE 5000
+
+WORKDIR /usr/src/app
+
+ENTRYPOINT ["python", "./runserver.py"]
```
jnovack: `CMD ["-h"]` should be provided so that by DEFAULT (running the container with no options), the help is printed.

Any command-line added will overwrite the `CMD` parameter appropriately.

jnovack:
I have two comments:

* `ca-certificates` is unneeded after install, and can be deleted.
* `CMD ["-h"]` should be provided so that when the container runs with no options, the help is printed by default.  When you run a container with options, it overwrites the `CMD` provided options and runs as expected.

Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Basic docker image for PokemonGo-Map
+# Usage:
+#   docker build -t pokemongo-map .
+#   docker run -d -P pokemongo-map --host 0.0.0.0 -a ptc -u YOURUSERNAME -p YOURPASSWORD -l "Seattle, WA" -st 10 --google-maps-key CHECKTHEWIKI
```
jnovack: `--host 0.0.0.0` should be provided in the `ENTRYPOINT` line, because it will ALWAYS be on every run of the container.

Fixed line:

    ENTRYPOINT ["python", "./runserver.py","--host", "0.0.0.0"]

jnovack:
@ultrafez I PR'd your PR (PRsception!) for the changes I suggested.

Dockerfile
chuyskywalker: > ca-certificates is not needed after the install (pip) to run the script. It can be deleted after to save space.

I found that without that, afterwards, you couldn't talk to the API endpoints.

Dockerfile
chuyskywalker: Will a second `--host` override this one? In some rare cases (not that the general purpose container should care too much) you may not want `0.0.0.0`.

.dockerignore
```diff
@@ -0,0 +1,8 @@
+.github
```
chuyskywalker: You can add `.git` to this as well to really cut down build time

Dockerfile
Ashex1: Do we _really_ need to worry about edge cases? Unless you're doing something crazy with docker-compose or vpn tunnelling you'll need this argument.

.dockerignore
jnovack: added to PR on ultrafez's repo

Dockerfile
jnovack: It's a serious edge-case, and I'd like to simplify it for the users.

Even with docker-compose and linking containers, you are going to want it on `0.0.0.0` unless you are running a private network of containers.  But then why have it at all if it's not exposed?

AHAAAAAAA:
@ultrafez @jnovack is this ready?

jnovack:
@AHAAAAAAA It is not.  @ultrafez has a PR to accept before I'll sign off on this.

MonkeyLeeT:
@jnovack Please do so in your earliest convenience :)

jnovack:
@MonkeyLeeT PR #1307 are the changes.

Sorry for the delay, issues at work.

PR #636, opened 2016-07-20
Auto refresh

Lunat1q:
Added auto refresh Pokemons on the map.
Pokemons being handled (added/removed) every 10 seconds.

+ [x] Auto refresh feature
+ [x] Timezone in label fixed


rezz3x:
Would be nice to bring back the -ar Argument to change the Seconds.

pogom/app.py
```diff
@@ -26,6 +26,10 @@ def fullmap(self):
                                lng=config['ORIGINAL_LONGITUDE'],
                                gmaps_key=config['GMAPS_KEY'])
 
+<<<<<<< HEAD
```
calebthompson: merge conflict markers here.

volksport:
I think you see them, but there are some merge conflicts in some of those files.

mkody:
Yes, if you could please use the value from `args.auto-refresh` if the user uses the `-ar` or `--auto-refresh` argument.

Lunat1q:
Found an conflict, merging now.

Lunat1q:
ok, fixed.
About -ar parameter.
Autorefresh speed being controlled by JavaScript, not Python.

Lunat1q:
Also looks like i fixed wrong timezone in Pokemon label.

volksport:
You can use server variables in the template to change the JS...

Lunat1q:
i'm not that string in Python, but anyway, i think it should be easy for skilled Python master :)
still, actually, i vote for UI slider, to change request frequency

JZ6:
+1 from me

PR #576, opened 2016-07-20
Adds gym, pokestop info, ignore db journal, fix timestamps and countdown timers

invisiblek:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

-Add gym info
-Add pokestop info
-Minor code cleanups, removal of unnecessary logging
-Add database journal to gitignore

If this is related to an existing ticket, include a link to it as well.
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/510
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/589 
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/591 
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/616 
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/626

PR #640, opened 2016-07-20
Add back -dg argument to display gyms

cschmatzler:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- branch out initMap() function
- add back handling of -dg or --display-gyms argument
- show gyms on map if argument is true

#510


rezz3x:
#576 already has it included

PR #647, opened 2016-07-20
Auto refresh #2

Lunat1q:
Added auto refresh Pokemons on the map.
Pokemons being handled (added/removed) every 10 seconds.


JZ6:
on hold to be merged later

templates/map.html
```diff
@@ -93,6 +93,8 @@
     var center_lng = {{lng}};
   </script>
 
-	<script type="text/javascript" src="/static/map.js"></script>
+
   <script async defer src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&callback=initMap"></script>
+
+  <script type="text/javascript" src="/static/map.js" type="text/javascript"></script>
```
AHAAAAAAA: Leave it on top to avoid errors. Also there's 2 types in that line.

Lunat1q:
Fixed map.html, thanks!:)

AHAAAAAAA:
alright I'll comment on the few lines changes needed to pass the -ar values into map

AHAAAAAAA:
nvm, solution I had in mind is pretty hacky. lets keep it at fixed intervals. Merging

SeeingBlue:
Database error.
You have to delete the pogom.db and let it recreate.

SeeingBlue:
Loaded a few pokemon then stopped, even though logs said it was still finding them. Also expiration timers still showed 00:00:00

toumaz:
In case anybody would find it useful - since this introduces a breaking schema change, if you want to perform a database migration and keep your old data you can perform a one-off run like this:

```py
from pogom.models import db
from playhouse.migrate import migrate, SqliteMigrator, DateTimeField
migrator = SqliteMigrator(db)
migrate(migrator.add_column('pokemon', 'detect_time', DateTimeField(default=0)))
```

Thanks for fixing the auto refresh!

Issue #352, opened 2016-07-19
Restart app by rest


Fiofant:
Sometimes the program hangs and does not do the following steps.
It would be nice to add restart applications by  REST API.
Like this: 

> 127.0.0.1:5000/restart?pass=qwerty1

UPD: @mkody rights, need to use your password

mkody:
But what if you're hosting this on a server (because you want to use it on your phone) and that someone finds your instance and type himself `/restart` to annoy you?
Maybe with a password-thing, like `127.0.0.1:5000/restart?key=password`.

SilverC:
@Fiofant I agree with @mkody. I think we should get a separate PR on the backlog to implement a password feature. My PR needs the feature as well #651

PR #651, opened 2016-07-20
Add next_loc route

SilverC:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Add next_loc which moves the next search to the specified lat long
-
-

#510

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -19,7 +19,7 @@ class Pogom(Flask):
         self.route("/gyms", methods=['GET'])(self.gyms)
         self.route("/pokestops", methods=['GET'])(self.pokestops)
         self.route("/raw_data", methods=['GET'])(self.raw_data)
-        self.route("/next_loc", methods=['GET'])(self.next_loc)
+        self.route("/next_loc", methods=['POST'])(self.next_loc)
 
     def fullmap(self):
         return render_template('map.html',
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 import calendar
-from flask import Flask, jsonify, render_template
+from flask import Flask, jsonify, render_template, request
 from flask.json import JSONEncoder
 from datetime import datetime
 
@@ -19,6 +19,7 @@ class Pogom(Flask):
         self.route("/gyms", methods=['GET'])(self.gyms)
         self.route("/pokestops", methods=['GET'])(self.pokestops)
         self.route("/raw_data", methods=['GET'])(self.raw_data)
+        self.route("/next_loc", methods=['GET'])(self.next_loc)
 
     def fullmap(self):
         return render_template('map.html',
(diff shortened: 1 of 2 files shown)


pogom/app.py
```diff
@@ -19,6 +19,7 @@ def __init__(self, import_name, **kwargs):
         self.route("/gyms", methods=['GET'])(self.gyms)
         self.route("/pokestops", methods=['GET'])(self.pokestops)
         self.route("/raw_data", methods=['GET'])(self.raw_data)
+        self.route("/next_loc", methods=['GET'])(self.next_loc)
```
mikeporterdev: Should this be a POST request?

SilverC: I had originally made it a POST request because that's what makes sense I think. I changed it to a GET because I thought that's what was originally implemented and wanted to get feedback on which to use

mkody:
Can I suggest the same thing I said in https://github.com/AHAAAAAAA/PokemonGo-Map/issues/352#issuecomment-233633535 ?
> Maybe with a password-thing, like 127.0.0.1:5000/restart?key=password.

This could be optional and enabled only if we pass an argument (for example `--protect "mypassword"`)

SilverC:
@mkody Yea I agree but maybe we did a single PR to implement that feature on both of these features after they're merged in?

mkody:
@SilverC I'm okay with that! :+1:

AHAAAAAAA:
+1 works! Good job

adrianocr:
Confirm a suspicion for me, please...

If I were to add a `navigator.geolocation.getCurrentPosition` call in JS and sent an ajax request to the /next_loc I've basically created a pseudo autolocation feature, right?

SilverC:
@adrianocr that's correct. It'll only take effect on the next cycle through the scan though

megahambone:
I've already done this. I'm gonna fix it so that it uses SilverC's next_loc instead of my own and update my pull request.

adrianocr:
@SilverC @megahambone anyway to force an update/refresh along with the POST request? That way when the ajax call sends over the latitude/longitude it also forces it to update for the new location. 

But I think some kind of timer would have to be implemented. Like poll for a new location every 30 seconds only. Otherwise the poor app is gonna go crazy refreshing every second.

megahambone:
@adrianocr - #725 has this working, but I didn't want to refresh the search thread - it doesn't take very long for it to update.

Earlier tonight someone thought 20 seconds was too long, you say 1 is too short, there's no winning. Maybe 5?

adrianocr:
@megahambone hahaha I'm just trying to be reasonable here. Does anyone *really* need an up-to-the-exact-second accurate thing? I don't believe so. I would personally be content with 20. Better yet would be a config variable to let people choose. I'd probably set it between 20-30 second. If some data-junkie wants to set it to 1s then that's his/her prerogative, I guess.

megahambone:
@adrianocr - I'm not familiar enough with the chosen web framework to make it into a config variable. If you want to change it statically, it's on line 251 in map.js
window.setInterval(getLocation, 1000); <-- interval in ms.

SilverC:
@adrianocr @megahambone I think we could add a rest endpoint that would restart the scan on the server. That being said the password feature for securing the server mentioned by @mkody is becoming more and more needed.

PR #631, opened 2016-07-20
Display sound as an alert when a Pokemon appears

NgJason:
## Description
1) -s Play sound to alert the user via console when a desire Pokemon is discovered.
2) -sw Play sound to alert the user via browser when a desire Pokemon is discovered.

Usage:
pip install --upgrade -r requirements.tx
To activate audio for console, add -s flag.
To activate audio for console, add -sw flag.

## Motivation and Context
This is especially useful for users who multitask and have this program run in the background.
https://github.com/AHAAAAAAA/PokemonGo-Map/pull/470

## How Has This Been Tested?
Executed program locally.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #658, opened 2016-07-20
Move --only --ignore to run earlier

NgJason:
## Description
Move --ignore and --only "down" , don't save it in to the db at all since we don't need it.

## Motivation and Context
Increase performance. This is also part of:https://github.com/AHAAAAAAA/PokemonGo-Map/pull/631
We do not need to save Pokemons that we do not need in to the db, since they will be ignored when get_active() executes. This will make the query[] shorter by X amount less iterations, where X is number of ignored Pokemons. 

## How Has This Been Tested?
- [ ] Testing
Pokemon servers are currently down, but ran some test in https://github.com/AHAAAAAAA/PokemonGo-Map/pull/631 before.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -9,7 +9,8 @@ from base64 import b64encode
 
 from .utils import get_pokemon_name
 
-
+from pogom.utils import get_args
+args = get_args()
 db = SqliteDatabase('pogom.db')
 log = logging.getLogger(__name__)
 
@@ -20,9 +21,6 @@ class BaseModel(Model):
 
 
 class Pokemon(BaseModel):
-    IGNORE = None
-    ONLY = None
-
     # We are base64 encoding the ids delivered by the api
     # because they are too big for sqlite to handle
     encounter_id = CharField(primary_key=True)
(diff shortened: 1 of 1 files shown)


toumaz:
I can see value in committing all data regardless of the current setting.

Won't this lead to somewhat confusing behavior if the app is restarted with different --only/--ignore settings? I fear you will "lose out" on a bunch of data you might want to present later. (Or if filter settings are later exposed as part of the UI, I'm not sure if this would be the way to go - not sure about plans there, so might be wholly irrelevant.)

I'd love to see if there's a notable performance increase or if it is negligible, too - if the latter, I'd value having historical data higher.

If performance is important, what do you think about instead adding indices to the database?`disappear_time` and `pokemon_id` seem like good candidates.

Perhaps modifying the query as appropriate to include `.in` and `.not_in` [operators](http://docs.peewee-orm.com/en/latest/peewee/querying.html#query-operators) would also do the trick?

NgJason:
hi @toumaz You've raised a good point that having the entire data set and filter later is beneficial to front-end UI filter, but I am not sure if there are any plans for that yet. 

In regards to confusing behavior, I don't see that happening because when new filters is applied, the app restarts and search from 1% anyway.

toumaz:
@NgJason Yes, but the database will still be populated with data from the previous search.

Let me try to explain it better. Right now, if you show  `--only` certain Pokémon, you'll still store data for all the rest, adding that to the database.

Let's say you then restart the app, deciding to show _all_ Pokémon, instead. You are correct in that search restarts, but since results are stored in and queried from the database, the previously found Pokémon will immediately be populated onto the map.

AHAAAAAAA:
I agree with @toumaz. It won't be very consistent if we stored just specific pokemon when using those flags. It's easy just to store all and then filter.

PR #660, opened 2016-07-20
Map Enhancement with Places Search Box

macgngsta:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- ~~Moved search thread to Pogom.app in order to be able to control search thread through routes~~
- ~~Added new route `setLocation` to take in new location from the browser~~
- Added Places API to the Map
- Selecting a new place through search bar will set a new current marker and restart the search thread
- Modified to use `next_loc`
- Added argument parameter to display/not the google search bar `-dgs` or `--display-gsearch`
- Modified js so that if pokestops or gyms are not being display, the ajax call is not made

mapfeature

#1039


jangrewe:
This sounds like a great idea and i haven't tested it yet, but is it possible to make this - especially the setLocation - configurable (e.g. disable functionality, for "public" maps).
We're sharing the map for our Team on my server, and i wouldn't want anyone to update the center location every 2 minutes ;-)

mkody:
Can you apply your change properly based on the current develop commits? I see you've undone quite a lot of commits here.

sykhro:
You need to use the sass and compile it, we're not accepting direct css changes

macgngsta:
it has been compiled with sass?

ellisio:
@sykhro `static/dist/*` was never purged, only added to `.gitignore`. So whenever someone makes a change, then compiles, their commits are including the updates JS and CSS.

alloin:
cant wait for this, looking nice!

SilverC:
@macgngsta I don't think we want to move the search thread back into the flask app. I think it was separated out to decrease the coupling in the first place. I

t may be better to use the next_loc method to post the new lat and long. The new lat and long will get picked up on the next cycle the search thread makes.

macgngsta:
@SilverC  - `next_loc` is new. I'll give it a look

AHAAAAAAA:
Most of these features have been implemented by other contributors, but it would be pretty useful to incorporate:
* Modified js so that if pokestops or gyms are not being display, the ajax call is not made
if you'd like to make a PR for that!

PR #671, opened 2016-07-20
develop: use pgoapi from requirements.txt instead of copying its files

Chlodochar:
This pull request includes a new feature

- Added pgoapi to requirements.txt. See https://github.com/tejado/pgoapi/pull/40 and https://pip.readthedocs.io/en/1.1/requirements.html#requirements-file-format
- Moved utility functions that were manually added (and not in official pgoapi) from pogom/pgoapi/utilities.py to pogom/utils.py
- Refer to pgoapi instead of pogom.pgoapi in imports

Give it a go. It should install pgoapi when you do `pip install -r requirements.txt`. You can change the pgoapi commit hash in requirements.txt to update pgoapi to another version.


tekulvw:
Why use a URL with a commit hash instead of `git://github.com/tejado/pgoapi.git@master#egg=pgoapi`?

Chlodochar:
@tekulvw You never know when pgoapi is making breaking changes. Better to have a fixed version and let the PokemonGo-Map maintainers decide when it is a good time to upgrade.

AHAAAAAAA:
can you resolve conflicts?

tejado:
@Chlodochar yeah good idea, i can introduce tags if wished. Never used tags so lets see... ;)

Would be also good if you can change the credits in the readme, that pokemongo-map is using pgoapi. Further credits (e.g. to minor contributions like Milas) are in my readme. Thanks :)

tejado:
https://github.com/tejado/pgoapi/releases/tag/v1.1.0

Chlodochar:
Done. It's using v1.1.0 now. I haven't tested whether all features are still compatible with v1.1.0, but it moves things forwards anyway.

DerScreever:
Tryed your pull request. i like the idea.

but i got some errors.

`Argument since_timestamp_ms with                                 value  unknown inside GetMapObjectsMessage (Exception: '\x00\x00\x00\x00\x00\x0                                0\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' has type <type 's                                tr'>, but expected one of: (<type 'int'>, <type 'long'>))
2016-07-21 10:34:59,519 [    rpc_api] [WARNING] Argument cell_id with value ▒▒▒▒                                ▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒                                ▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒                                ▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G▒▒▒▒▒▒▒▒G unknown inside GetMapObjectsMessage (Exception: '\x80\                                x80\x80\x80\xd4\xc5\xec\xdbG\x80\x80\x80\x80\xdc\xc5\xec\xdbG\x80\x80\x80\x80\xe                                4\xc5\xec\xdbG\x80\x80\x80\x80\xec\xc5\xec\xdbG\x80\x80\x80\x80\xf4\xc5\xec\xdbG                                \x80\x80\x80\x80\xfc\xc5\xec\xdbG\x80\x80\x80\x80\x84\xc6\xec\xdbG\x80\x80\x80\x                                80\x8c\xc6\xec\xdbG\x80\x80\x80\x80\x94\xc6\xec\xdbG\x80\x80\x80\x80\x9c\xc6\xec                                \xdbG\x80\x80\x80\x80\xa4\xc6\xec\xdbG\x80\x80\x80\x80\xac\xc6\xec\xdbG\x80\x80\                                x80\x80\xb4\xc6\xec\xdbG\x80\x80\x80\x80\xbc\xc6\xec\xdbG\x80\x80\x80\x80\xc4\xc                                6\xec\xdbG\x80\x80\x80\x80\xcc\xc6\xec\xdbG\x80\x80\x80\x80\xd4\xc6\xec\xdbG\x80                                \x80\x80\x80\xdc\xc6\xec\xdbG\x80\x80\x80\x80\xe4\xc6\xec\xdbG\x80\x80\x80\x80\x                                ec\xc6\xec\xdbG\x80\x80\x80\x80\xf4\xc6\xec\xdbG' has type <type 'str'>, but exp                                ected one of: (<type 'int'>, <type 'long'>))
`

tejado:
The issue is due to the new (more correct) protos. since_timestamp_ms and cell_id are now repeated fields, so you have to provide a list (and pgoapi will parse it and assign it accordingly) See the example in pokecli.py:

    cell_ids = get_cell_ids(position[0], position[1]) #  new get_cell_ids returns a list
    timestamps = [0,] * len(cell_ids) # also a list now
    api.get_map_objects(latitude = util.f2i(position[0]), longitude = util.f2i(position[1]), since_timestamp_ms = timestamps, cell_id = cell_ids)

DerScreever:
@Chlodochar 

could you pls change what @tejado told in your pull request?
and remove also the complete "pgoapi" folder

Chlodochar:
@DerScreever Git doesn't track directories, so it can only add/remove files. I can't remove left-over directories.

@tejado It seems that v1.1.0 now uses protobuf3 and requires protobuf3 binaries to be installed on the system. I'm guessing this will cause some uproar as it probably isn't installed by default for a lot of people. (I had trouble getting this to work)

I've reverted this PR back to the old version of pgoapi (the one that is compatible with PokemonGo-Map). It can be upgraded later.

DerScreever:
@Chlodochar for me it dont user protobuf3? we just neeed to fix what @tejado wrote or not?

tejado:
No, it uses protobuf3 now which is the recommended protobuf version (by google) and also used by POGOProtos. The requirements.txt is adjusted accordingly, I added also a version check, which raises the exception PleaseInstallProtobufVersion3

But no binaries are required, just the normal protobuf package in version >=3.0.0a3

DerScreever:
@Chlodochar then pls update again to the tag. we will fix the other things then and merge this :D

tejado:
Btw: only protobuf version 3 is supported by python 3, so I had to switch for better compatibility.

yashmangal9506:
what do extracting after  the zip file

yashmangal9506:
i am not getting that step "open command window here

Chlodochar:
@DerScreever If you want to move forward, merge this PR. I got PleaseInstallProtobufVersion3 exceptions with the latest changes from @tejado and still figuring out how to solve it. I have the python-protobuf3 installed, like the requirements.txt says, but it seems it also depends on system-installed protobuf binaries. I'm not sure what is up yet. I presume, if the new version of @tejado was to be used right now, that other people would run into this problem as well.

@yashmangal9506 That is unrelated to this PR. Please create a new issue.

tejado:
I dont know what python-protobuf3 is, but if this is "pip install protobuf3" it is not the package of google and doesnt contain the complete lib:  https://pypi.python.org/pypi/protobuf3/

tejado:
@Chlodochar which OS are you using?

Chlodochar:
ArchLinux. But it seems it has something to do with one of the other dependencies of PokemonGo-Map. The cli does work here.

I removed the protobuf2 dependency from PokemonGo-Map, but it seems pip is still installing protobuf2 because (probably) some other dependency uses it or maybe it looks it up from my /usr/lib/. Maybe protobuf-to-dict is the problem, still haven't figured this one out.

tejado:
Why not installing it over pip (which should be the preferred way or?) but over an AUR?

protobuf-to-dict is also not useable with py3, so there is protobuf3-to-dict or also it is inside pgoapi ;)

Chlodochar:
@tejado Yes, I'm using pip, but I already have packages installed using AUR (dependencies of other applications).

I found out that:
```
protobuf-to-dict==0.1.0
-e git://github.com/tejado/pgoapi.git@v1.1.0#egg=pgoapi
```
will install protobuf2 and
```
-e git://github.com/tejado/pgoapi.git@v1.1.0#egg=pgoapi
protobuf-to-dict==0.1.0
```
will install protobuf3.

I'll try to get 1.1.0 in there again.

tejado:
For what do you install protobuf-to-dict? It is not anymore a requirement of pgoapi :P

Chlodochar:
Ah, because PokemonGo-Map had it in requirements.txt I thought it was still needed there. It isn't. I'll remove it ;) Thanks

tejado:
I will make the changelogs for the next releases more detailed, sorry :)

Chlodochar:
Ok, things are working now here for v1.1.0.

Chlodochar:
@tejado Just to note, it might be handy to have some of the util functions that are now in cli in the lib itself. That way they do not need to be copied (https://github.com/AHAAAAAAA/PokemonGo-Map/pull/671/commits/b655bd6991c3a2e1bacc56c1bf233302a6c549bc#diff-6c84a64c790b52a5ff2656d6bb388fefL126).

tejado:
Maybe you should check which requirements you have to list in your repo as I guess the requirements of pgoapi will be also installed recursive.

tejado:
@Chlodochar Yes, I also thought about this. Will do as soon as possible but currently have no time for it. Feel free to make a PR to move this stuff to utilities.

Chlodochar:
@tejado have to get back to work as well, but I already removed some of the other requirements that were in pgoapi and not used in PokemonGo-Map.

@DerScreever Could you give this another look and consider merging?

DerScreever:
@Chlodochar if you remove following requirements:

geopy==1.11.0
s2sphere==0.2.4
gpsoauth==0.3.0

because they are already installed with the api.
then i check your code :-)

Chlodochar:
@DerScreever Yes, but they are still used in the utilities. Yes, they are automatically installed with pgoapi, but PokemonGo-Map is still referencing those through code. When pgoapi has added the utility functions to the api, then those functions can be removed rom PokemonGo-Map and then those requirements can be removed as well.

I forgot about gpsoauth. That one is removed now as well.

DerScreever:
no if you install the pokeapi module the requirements are installed also ^^ so this packages get installed from the api.

Chlodochar:
@DerScreever I know, that isn't the point. PokemonGo-Map refers to libraries in code, then it should refer to them in requirements. Practically at this point in time it is the same. However if pgomap were to remove s2sphere from its dependencies and it is nowhere referring to that lib, then that would break PokemonGo-Map, because PokemonGo-Map is implicitly depending on s2sphere. I'd like that to be explicit, so that such hidden errors do not occur when updating dependencies.

But if you really want, I can remove it.

DerScreever:
@Chlodochar 

sorry my mistake then ^^ thinked not about this :-) 
+1 for you cool commit. 
like this ^^
i will test it later, because we doing some changes etc.

Chlodochar:
@DerScreever Could you merge this?

Chlodochar:
Or @AHAAAAAAA ? I've rebased and patches this PR quite a number of times now.

Chlodochar:
@marauder37 Could you have a look? I just did another rebase and patched things up. Changes to pogom/pgoapi shouldn't be done in PokemonGo-Map, but should be done upstream (https://github.com/tejado/pgoapi). Or else the next version of pgoapi will be painful.

andrewda:
:+1:

Chlodochar:
Another rebase, should be good again.

Chlodochar:
Rebased again. Since @Tolo had trouble with pogom/pgoapi still existing, it is now automatically being removed at startup.

DerScreever:
this looks good for me
+1

Chlodochar:
If anyone else has tested this, please merge.

DerScreever:
@Chlodochar i test when im home. so i think in about 5 hours. after that i will merge if nobody did before me.

DerScreever:
for me all work great. but i think this catch the pokemon slower then defaut....

tejado:
Maybe you can run it with an profiler?

DerScreever:
@tejado you know any profiler that i can use on my server? :-)

tejado:
I would say, this is possible with most python profilers.Try cprofile: https://julien.danjou.info/blog/2015/guide-to-python-profiling-cprofile-concrete-case-carbonara

DerScreever:
could someone else benchmark this? not working for me.

Chlodochar:
I checked how long a single iteration takes. For both methods exactly 37 seconds on my pc. With this PR it found 47 pokemon in the first iteration, with the develop version 40 pokemon. Could be coincidence though.

I used the following change to measure time in search.py:
```
        lat_location += lat_gap_degrees
        lng_location -= calculate_lng_degrees(lat_location)

+       print("Time: {}".format(time.time()))
+
        for direction in range(6):
            for i in range(ring):
```

efiku:
+1 !

AHAAAAAAA:
+1, rebase

hhu94:
What's the status on this PR? I need the updated protos to fix the implementation of lured pokemon in map.js and to create a new webhook for them.

AHAAAAAAA:
Closing this as it's being dealt with in #3106

PR #677, opened 2016-07-20
Adding BasicAuth 

bikrdotnet:
This pull request includes a

- [ ] Bug fix
- [ x ] New feature
- [ ] Translation

- Added BasicAuth 
     Default username and password: pokemon / pokemon 

- Added  Flask-BasicAuth==0.2.0 to requirements.txt

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -12,6 +12,8 @@ from pogom.utils import get_args, insert_mock_data, load_credentials
 from pogom.search import search_loop
 from pogom.models import create_tables, Pokemon
 from pogom.pgoapi.utilities import get_pos_by_name
+from flask_basicauth import BasicAuth
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


runserver.py
```diff
@@ -58,6 +60,13 @@ def start_locator_thread(args):
         insert_mock_data(args.location, 6)
 
     app = Pogom(__name__)
+    
+    app.config['BASIC_AUTH_USERNAME'] = 'pokemon'
+    app.config['BASIC_AUTH_PASSWORD'] = 'pokemon'
+
+    basic_auth = BasicAuth(app)
+    app.config['BASIC_AUTH_FORCE'] = True
```
ellisio: This is enabling this by default, isn't it? Probably should set this to `False` and then add a flag on the args.

Ashex1:
Seems if authentication is going to be implemented that this should be configurable either as an argument or from a file. Hardcoded credentials basically prevents people from using the project and pulling changes in easily.

talentedmrripley:
Perhaps it could get the credentials from credentials.json.  And I agree it should default to false with a flag to enable.

bikrdotnet:
Thanks for the comments.  Going to dig in and see how the parameters work and relate to the py.

bikrdotnet:
New PR Coming.

PR #682, opened 2016-07-20
develop: added heroku support using Procfile

Chlodochar:
This adds minimal Heroku support. Documentation should go into the wiki once this is in master.

### Documentation

- Create an app on Heroku.
- Clone PokemonGo-Map on your PC:
```sh
$ git clone https://github.com/AHAAAAAAA/PokemonGo-Map
$ cd PokemonGo-Map
$ git checkout develop
```
- Copy `credentials.example.json` to `credentials.json` and edit its contents.
- Commit `credentials.json`
```sh
$ git add -f credentials.json
$ git commit -m "added credentials.json"
```
- Push PokemonGo-Map to Heroku:
```sh
$ git push https://git.heroku.com/yourapp.git develop:master
```
- On Heroku add the following environment variables: `USERNAME`, `PASSWORD`, `LOCATION` and `STEP_COUNT`.
- On Heroku start and open the app.

Optionally you can add your own parameters using the `EXTRA_ARGS` environment variable. For example you can use `-dp -dg -ar 60` for `EXTRA_ARGS` to get pokestops, gyms and auto-refresh.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: python runserver.py -u "$USERNAME" -p "$PASSWORD" -l "$LOCATION" -st $STEP_COUNT $EXTRA_ARGS -H 0.0.0.0 -P $PORT


gsabran:
Nice PR :)

Ayysir:
-ar 60 is in seconds or minutes?

AHAAAAAAA:
Can you add this to the Wiki before we merge?

gsabran:
-ar is in seconds, so -ar 60 is auto refresh once per minute

Chlodochar:
I do not have permission to write to the wiki.

The steps are a bit more involved now that `credentials.json` isn't provided anymore. A good alternative is to have Google Map credentials passed as an environment variable or commandline argument, but that can be done later.

@gsabran If there is documentation on the different command-line arguments, then it makes sense to refer to that doc from the Heroku doc.

DanielVolz:
Maybe I'm missing something, but where is the Procfile?

SilverC:
@Chlodochar I think you should also add support for pulling in the command line arguments via config vars so that username and password don't have to be source controlled in the procfile

Chlodochar:
@DanielVolz I guess you missed: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/682/files ?

@SilverC That's already possible with `EXTRA_ARGS` I think?

punkstar:
Great work @Chlodochar.  I've tested this and it works fine for me.  I've opened #946 that takes this a step further and adds an `app.json` and a deploy button.

MonkeyLeeT:
Document added for you. Great job guys.

PR #702, opened 2016-07-21
Add -dg -dp -dl back in and stamped calls

invisiblek:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

-fix parameters for showing/hiding markers
-fix calling routes without a timestamp (many don't need it)

https://github.com/AHAAAAAAA/PokemonGo-Map/issues/705

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -16,8 +16,9 @@ class Pogom(Flask):
         self.json_encoder = CustomJSONEncoder
         self.route("/", methods=['GET'])(self.fullmap)
         self.route("/pokemons/<stamp>", methods=['GET'])(self.pokemons)
-        self.route("/gyms/<stamp>", methods=['GET'])(self.gyms)
-        self.route("/pokestops/<stamp>", methods=['GET'])(self.pokestops)
+        self.route("/pokemons", methods=['GET'])(self.pokemons_all)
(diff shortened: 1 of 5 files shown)


AHAAAAAAA:
+1 works, code style is good.

medghaim:
This is great except for -dl.

-dp and -dg work separately, and together.

-dl doesn't show any lured pokestops, which i confirmed were there in game, and using -dp :)

invisiblek:
@medghaim seems to be working as expected for me...

Just -dp
dp only

With -dp and -dl
dp and dl

medghaim:
Oh oops. I could've sworn it was -dl alone :O

Eitherway, +1. Good stuff. Sorry about the mixup lol

invisiblek:
well, we could make -dl imply -dp,  suppose that would be intuitive

AHAAAAAAA:
I understand the confusion. Should -dl work standalone to just show lured stops without requiring -dp as well?

-dp: show all
-dl : show lured only
-dp -dl: show lured only

medghaim:
Yeah, imo it should be used stand alone. What do others think?

invisiblek:
yea, that probably makes the most sense, patch incoming

PR #713, opened 2016-07-21
adding os path sep

medghaim:
This pull request includes a

- [x] Bug fix

- Using `os.path.sep` instead of hardcoded '/'
- (Thanks Gaudon <3)


AHAAAAAAA:
+1, tested works.

PR #717, opened 2016-07-21
Basic Auth - HTTP - User / Pass from credentials or line args Enable from line args. 

bikrdotnet:
This pull request includes a

- [ ] Bug fix
- [ x ] New feature
- [ ] Translation

- Added basic_auth
- Added argument --auth-enable , -ae  
- Added argument --auth-user, -au
- Added argument --auth-pass -ap
- Can add basic auth user to credentials.json or line args.  
- Added prerequisites to requirements.txt

### DIFF
--- a/credentials.example.json
+++ b/credentials.example.json
@@ -1,3 +1,5 @@
 {
 	"gmaps_key" : ""
+	"auth_user" : ""
+	"auth_pass" : ""
 }
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -23,6 +23,8 @@ def get_args():
     # fuck PEP8
     parser = argparse.ArgumentParser()
     parser.add_argument('-ae', '--auth_enable', help ='Basic Auth', action='store_true', default=False)
+    parser.add_argument('-au', '--auth_user', type=str, help='Auth User')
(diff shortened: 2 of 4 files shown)


bikrdotnet:
Took suggestions on not mandating and static U/P .

MonkeyLeeT:
@bikrdotnet we can merge once wiki is updated with Heroku info. Could you help with that?

Bitti09:
isn't a reverse proxy like caddy ( verry small) a better solution for this ?

bikrdotnet:
No? Basic auth is a pretty standardized concept across web servers and should be optionally but into any candidate running flask imo. Forward facing install on aws right now and the ease of basic auth is very attractive.

Justinian6:
I found that adding html authentication directly into the app.py script worked well:
screenshot 46

bikrdotnet:
Def needs to be user enabled / disabled.

Justinian6:
I'm a Java developer and my experience in Python is EXTREMELY limited just based on what I've read through these scripts... However I assume you can just implement that and then replace static passwords with a variable that can be defined in the command line. Else if no variable is present assume yes in the requires_auth() function

talentedmrripley:
Why did this get closed?  I was really hoping to see this get merged.

Issue #700, opened 2016-07-21
Gym icons don't despawn and overlap


Skullmonkey:
### Environment info
Operating System:
Win10
Installed version of Python:
2.7.12

### Steps to reproduce
1. Run script as normal

### What have you tried?
1. Nothing. I'm sure a restart would fix it.

### Logs or other output that would be helpful
http://imgur.com/QMnYGnC

Skullmonkey:
More info:
I also noticed that Pokemon were also staying on the map past time. I did a refresh on the map (not the script) and they are gone.

invisiblek:
yea, this is due to the auto_refresh stuff, i'll look at fixing it

invisiblek:
@Lunat1q this seem to be broken even if I hard reset back to https://github.com/AHAAAAAAA/PokemonGo-Map/commit/74b460923de16f997a12020cb8f67684924abc09

Something you want to take a look at?

PR #732, opened 2016-07-21
fix for Gone! pokemon not being removed from map

bwalendz:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- No longer need the timezone correction when checking for disappeared pokemon to remove from map. All times are now UTC.

https://github.com/AHAAAAAAA/PokemonGo-Map/issues/700

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -145,7 +145,7 @@ GetNewPokemons = function(stamp) {
     lastStamp = dObj.getTime();
     
     $.each(markers, function(i, item){
-        if (item.disapear <= lastStamp - (dObj.getTimezoneOffset() * 60000))        
+        if (item.disapear <= lastStamp)        
             item.m.setMap(null);        
     });
 };


benemorius:
:+1: works for me

invisiblek:
Well, it fixes the cleanup portion of this whole thing, but another (bigger) issue here is with each iteration of the loop, its dropping down another duplicate marker. sure this cleans them all up later (for pokemon at least, not gyms or pokestops), but it definitely likes to chew up memory before it gets to that point

My patch, although maybe a bit long-handed, ensures that only one instance of a particular marker will exist, ever. I also, if I encountered the same pokemon_id/gym_id/pokestop_id, threw a bit of logic together to determine if it was worth deleting the old marker and putting a new on in its place.

I think with that approach, we could actually strip out a lot of this messy timestamp business.

https://github.com/AHAAAAAAA/PokemonGo-Map/pull/736

Ayysir:
@invisiblek just tried your patch on my server interation however, with -ar set at 3 .... pokemon still sticks even if its declared as 'gone'

invisiblek:
Yea, I just noticed that. I see what's up. I'm just removing from the array, not the DOM on those cases.

Ayysir:
@invisiblek no problem, im here to test if anything

JZ6:
who tested with good results?

hhu94:
I applied the same fix on my build and it does indeed fix it. But I agree with @invisiblek, there are bigger underlying issues with the markers.

invisiblek:
hold off on this for a little bit please

MikeDX:
This does not fix the real issue, which is the stored time not being utcnow. Please see #751 for a complete fix

invisiblek:
Getting close: https://github.com/invisiblek/PokemonGo-Map/commit/9ef49cf327967ca7793be93242570a33ad2da14d

DerScreever:
@invisiblek 
when you finished your solution pull request pls. then i can close this one and we test your pull request.

invisiblek:
@DerScreever someone else will have to look at this. For whatever reason it was decided to completely rewrite half of this app in one giant, ugly commit. I'm not spending the time to figure out why this is broken now.

DerScreever:
@invisiblek no problem ^^ i know how you feel :-) 
we will do this for you :)

sykhro:
@invisiblek I honestly don't know why it was all pushed into a single commit, because I had checked and written different commit messages. Also, no one said you have to figure it out :) issue was fixed as soon as it was reported

Issue #714, opened 2016-07-21
Pokemon arent removed until page refresh


Bakies:
on dev branch
They are marked Gone! when you hover over them but never get removed from the map
commit 73c151d1d56a342925b2be643b3e7f4fd8476dee

Skullmonkey:
#704 and #705 duplicate.

Bakies:
closed: goto dupes

Bakies:
those dupes seem like they are different issues so I am re-opening this issue.

Skullmonkey:
My bad, I meant to reference #700

Bakies:
@DerScreever this is still an issue, should not be closed

PR #736, opened 2016-07-21
Fix duplicate markers showing up on map

invisiblek:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

-use arrays in the DOM to check if the marker already exists and react accordingly

https://github.com/AHAAAAAAA/PokemonGo-Map/issues/714
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/700

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -3,7 +3,10 @@ var map,
     lastStamp = 0,
     requestInterval = 10000;
 
-var markers = [];
+var pokemons = [];
+var gyms = [];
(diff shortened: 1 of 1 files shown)


Ayysir:
@invisiblek just tried your patch on my server interation however, with -ar set at 3 .... pokemon still sticks even if its declared as 'gone'

invisiblek:
this needs more work

Issue #596, opened 2016-07-20
Development branch finding many less pokemon than Master branch


talentedmrripley:
I'm running both the master and dev branch with the same parameters on the same machine.  The master branch is showing about double the amount of Pokemon in my area as the dev branch.

docrockulus:
I have the same issue, I am also unable to see pokestops/gyms on the map in the dev branch.

noaqzz:
http://imgur.com/60HT2SG

python runserver.py -a PTC -u XXX -p YYY -l "ZZZ" -st 6 -H 0.0.0.0 -P 12345 -k AAAAAA

python example.py -a PTC -u XXX -p YYY -l "ZZZ" -st 6 -H 0.0.0.0 -P 12345

docrockulus:
python runserver.py -a google -u xxx -p xxx -l "42.315035 -83.037306" -st 5 -H 0.0.0.0 -P 5000 -ar 30 -i 96

I also have two more running on 5001 and 5002, I see all the pokemon data they have each collected on the 5000 map, but the number of mons is signifigantly less than that of the master branch.

daruda87:
Same issue here...

talentedmrripley:
python example.py -u ***** -p ***** -st 10 -i "Pidgey,Rattata,Paras,Pidgeotto,Geodude,Mankey,Abra,Ponyta,Spearow,Drowzee,Venonat,Zubat" -l "*****,******" -H 0.0.0.0 -P 5002 -dg -ar 3

python runserver.py -u ***** -p ***** -st 10 -i "Pidgey,Rattata,Paras,Pidgeotto,Geodude,Mankey,Abra,Ponyta,Spearow,Drowzee,Venonat,Zubat" -l "*****,*****" -H 0.0.0.0 -dg -ar 3

nletourneau:
This is probably because of the fixing of the spiral logic. The old traversal was doing a lot of duplicate path finding across the map and it has been made a ton better... but the distances between scans is now greater as a side effect of this cleanup. 

Old style (master branch)
![image](https://cloud.githubusercontent.com/assets/225737/16997026/8590aa32-4e68-11e6-88b7-d61f28bf8391.png)

new style (develop branch)
![image](https://cloud.githubusercontent.com/assets/225737/16997034/8fce745c-4e68-11e6-947d-ffdf59d2beeb.png)

Likely need to just lower the 0.0025 in search.py to something like 0.0005 and increase the steps.

[...]

PR #755, opened 2016-07-21
Fix the issue of less pokemon appearing during scan

AHAAAAAAA:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

Fixed issue. The problem was simply that the requests happened too rapidly and we weren't getting the same information back.

Old develop (1 second between requests):

New develop (2 seconds between requests):

Current develop branch:
screenshot 2016-07-20 23 42 24
New develop branch:
screenshot 2016-07-20 23 50 46
Master:
screenshot 2016-07-20 23 45 38


AHAAAAAAA:
Solves #596

DerScreever:
Aprroved +1, Someone test pls

stereoa:
develop
image
fixscan
image

1 25 step loop done on each.. Looks identical to me. I could try a more populated area.

Ayysir:
no improvement, tested for central park, ny area

fredgido:
Still less pokemon here.

g3r4n:
Still less pokemon

fredgido:
@AHAAAAAAA  in line 37 of search.py if you replace a step of 0.0025 to a shorter one, like 0.00125 it works like master.

g3r4n:
It's because of map projection distorsion. Depending of distance betwween the equator and the area of scan.
I think we need a GIS expert and i'm not a GIS expert :smile:

nletourneau:
I've done a bit of investigation on scanning, putting my findings here. I might have time to get a PR up but if someone beats me to it and uses this info all the better.

# Cellids
list that gets sent is the client telling the server "here are the cell ids I care about, so only fetch data from these cell ids, the rest of the request will be a refinement on the data contained within these cellids"

The reason for this is that cellids are polygons, easy to convert to lat/lng (with libraries already included in this proj) and they are constant time comparison.

You need to make sure your cellid list overlaps the area in which your searching, this is not always intuitive (the plus/minus 10 doesn't work right in all circumstances)

# circle searching
once you have your cellid list properly generated pokemon show up in a 100m radius (harvesine distance) so to have an efficient search mechanism you'll need to do some overlapping like [this](http://stackoverflow.com/questions/7716460/fully-cover-a-rectangle-with-minimum-amount-of-fixed-radius-circles)

# example

this map illustrates what i'm talking about. The blue polygons are cells, the red line in the polygons is the sorted sequence of cells, The center of the red circle is a found pokemon, and the radiating red lines are traversals I did incrementing lat/lng until the pokemon disappeared. The red circle is 100m in radius.

You can see for two of the search branches (north and northeast) get cut short because they reach the limit of the cellid overlap.

image


Would this be better in the wiki?

JZ6:
@AHAAAAAAA @sykhro

Rhynocerous:
@nletourneau Scanning directly along the grid still produces missing pokemon

Shugabuga:
Was this ever fixed?

PR #757, opened 2016-07-21
Small modification to runserver.py to allow it to being packed via setup.py

Darkade:
This pull request adds a `main()` function to runserver.py The `main()` function can be used as an entry_point in a `setup.py`

It also adds more python ignores to `.gitignore` because of the files setuptools makes.

-------------

I've done an small version of a pip installer here https://github.com/Darkade/pokemongo-map-pip
This repo, `PokemonGo-Map` is a submodule of mine

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-*.pyc
 *.idea
 *.db
 *.db-journal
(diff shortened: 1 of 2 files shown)


MonkeyLeeT:
lgtm, anyone else testing if this is working fine?

PR #826, opened 2016-07-21
Updating error message in hopes of less people opening issues...

medghaim:
This pull request includes a

- [ ] Bug fix
- [x] New string!
- [ ] Translation

- Updated exception message - few people opening issues with that exception, I hope this makes it clearer..


DerScreever:
+1

marauder37:
Why don't we put the actual URL of the wiki page in the message?

medghaim:
I tried, it looks pretty ugly in the console window.

PR #862, opened 2016-07-21
Removed unnecessary google map customizations

TGM:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [X ] Improvement


sykhro:
This is up to the frontend team. Design was agreed on dark colors, it won't be changed. We would accept a PR for a switch between the two, but no replacements.

PR #871, opened 2016-07-21
Fix ES6 template string

DeWenS:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Fix ES6 template string


DerScreever:
@DeWenS 
you are done with this? or you want to add more?

DeWenS:
Now think its ok

DerScreever:
@DeWenS 
then clean it up pls....
i will merge when the code is okay.

benwaffle:
You could just add babel to grunt

BillyWM:
This backslash syntax is not part of the ECMAScript standard. Also you're mixing concatenation - "str" + "another str" - with the backslash multiline trick, which defeats the purpose. Might as well just do

"line" +
"next line" +
"another line"

The safest and cleanest thing IMO is a string replace, and use normal concatenation to handle multiline.

var str = "$poke is the best Pokemon" +
"like no one" +
"ever was";

str = str.replace("$poke", pokemon_name)

--------------

Babel is cool and I use it in my personal projects, but it can cause friction when using a debugger (You'll be stepping through transpiled output, and have to generate sourcemaps to get a semblance of the original back)

PR #882, opened 2016-07-21
Add ability to delay next scan by X seconds

FoxxMD:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

* Adds an optional `--scan-delay` (`-sd`) argument that delays `search_loop` execution by the specified number of seconds
* Default delay is 1 second (unchanged)
* Outputs the delay time after a completed search if delay time is not default

___

Nobody likes argument hell but this seems like a simple no brainer. 

For people like me who want to have PokemonGo-Map running in the background and check on it only every once in a while starting a scan immediately after the previous one isn't necessary (since I'm only looking at it every 5-10 minutes) and only contributes to strain on niantic servers without any real benefit to myself. This PR makes it easy to spread out time between scans.

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -93,4 +93,6 @@ def search_loop(args):
     while True:
         search(args)
         log.info("Scanning complete.")
-        time.sleep(1)
+        if args.scan_delay > 1:
+            log.info('Waiting {:d} seconds before beginning new scan.'.format(args.scan_delay))
+        time.sleep(args.scan_delay)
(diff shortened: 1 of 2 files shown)


pogom/search.py
```diff
@@ -93,4 +93,6 @@ def search_loop(args):
     while True:
         search(args)
         log.info("Scanning complete.")
-        time.sleep(1)
+        if args.scan_delay > 1:
+            log.info('Waiting {:d} seconds before beginning new scan.'.format(args.scan_delay))
+        time.sleep(args.scan_delay)
```
MonkeyLeeT: What if there's no -sd argument?

FoxxMD: I'm not too familiar with how `parser.add_argument` works but the argument has a default value of 1 -- will it not always exist even when not explicitly added?

MonkeyLeeT: Can you do a quick testing without  -sd flag (and manually change default to something like 5) ?

MichaelErmer:
+1, also works for me

PR #904, opened 2016-07-21
Add an initial Dockerfile

manuelmazzuola:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Add Dockerfile
- Modify README.md file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:2-onbuild
+
+ENV AUTH_SERVICE ptc
+ENV USERNAME ""
+ENV PASSWORD ""
+ENV LOCATION ""
+ENV ST 5
+ENV HOST "0.0.0.0"
+ENV PORT 5000
+ENV GMAP_KEY ""
+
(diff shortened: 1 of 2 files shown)


jangrewe:
Why not just set the entrypoint and the let the user pass the usual parameters, instead of all that ENV overhead?

manuelmazzuola:
The env overhead as you say is necessary because if I wanted to use an external service that "eat" my image and deploy it somewhere in the world I have to use the environments variables.

ellisio:
^ Such as Kubernetes. However, you should do an `entrypoint.sh` file instead like suggested. Then if you need defaults your entrypoint handles it. For example:

```
#!/bin/bash

if [[ -z "$AUTH_SERVICE" ]]; then
  AUTH_SERVICE=ptc
fi

exec "$@"
```

ultrafez:
There's already a discussion going on about adding Docker support in #599

MonkeyLeeT:
Yeah, #599 .

PR #907, opened 2016-07-21
Adding docker compose configuration

fyaconiello:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Added docker configuration to help spin up new environments without worrying about python/pip/etc versions. 

related to #904 #599 both of which use a Dockerfile, this solution utilizes the docker compose tool instead.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,12 @@
+flask:
+    image: python:2.7-onbuild
+    ports:
+        - "5000:5000"
+    volumes:
+        - .:/usr/src/app
+    working_dir: /usr/src/app
+    entrypoint: |
+        bash -c 'bash -s <<EOF
+        pip install -r requirements.txt
+        python example.py -a google -u "XXXXXXX" -p "YYYYYY" -l "ZZZZZZZZ" -st 10 --host 0.0.0.0
+        EOF'


fyaconiello:
run `docker-compose up --build` to initialize new map on port 5000 of your docker host.

ultrafez:
What benefits does Docker Compose give us over using plain Docker? The purpose of Docker Compose is to allow building applications that use multiple Docker images, which this application doesn't need.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+    image: python:2.7-onbuild
+    ports:
+        - "5000:5000"
+    volumes:
+        - .:/usr/src/app
+    working_dir: /usr/src/app
+    entrypoint: |
+        bash -c 'bash -s <<EOF
+        pip install -r requirements.txt
+        python example.py -a google -u "XXXXXXX" -p "YYYYYY" -l "ZZZZZZZZ" -st 10 --host 0.0.0.0
```
carloscheddar: I think that it may be more clear if you use environment variables here instead of `XXX`, `YYY` or `ZZZ`.

carloscheddar:
@ultrafez On this particular implementation if we use only Docker instead of Compose executing the `volumes`, `port forwarding` and `working_dir` would have to be done manually in the `docker run` command. With Compose we just do `docker-compose up` and it gets the configuration from the `docker-compose.yml` file.

rdamron:
I agree with @ultrafez, why do we need docker orchestration when we only need a single docker image? Seems needlessly complicated at this point.

rdamron:
Also this is meant for the master branch? The entrypoint seems to be "python example.py" (old) this would need updating to be considered.

carloscheddar:
I just saw #599, it looks very robust and it eliminates the need for `volumes` and `working_dir` that this branch introduces. So with #599 I don't see a need for this one either.

MonkeyLeeT:
#599 will be the starting point for all Docker support.

Issue #909, opened 2016-07-21
Use localstore to cache selection options


JakeCooper:
Add in a localstore object to cache the selections for gyms, pokemon, and pokestops.

JakeCooper:
Resolved in PR #929

PR #929, opened 2016-07-21
Implement Caching of Options

JakeCooper:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added caching in localStorage so that options can be preserved between refresh.


JakeCooper:
Resolves issue #909

PR #936, opened 2016-07-21
Map style switch option

JKapitein:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added a select box to the menu where the Google Maps theme can be chosen
- Currently has the dark theme (default) and the standard Google Maps theme.  
- Adding new map themes is done by creating a new entry in mapstyles.js, and adding the option to the select in the menu.

The selected theme is saved in a cookie.

Credits: These gists for the get/set cookie functions
https://gist.github.com/shvedchenko/9387318
https://gist.github.com/wpsmith/6cf23551dd140fb72ae7

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -1,124 +1,18 @@
 var map;
 
+var mapStyleCookie = 'map-style';
+var defaultMapStyle = 'dark';
+
 function initMap() {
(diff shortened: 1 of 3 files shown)


JKapitein:
I hate markdown by the way.

PR #940, opened 2016-07-21
Yaml parsing

kgurnani96:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

- Replaced cli args parsing with yaml config file

If this is related to an existing ticket, include a link to it as well.

PR #233, opened 2016-07-19
Created a config file using YAML

MelonSmasher:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

The following changes were made

- The `get_args` function looks for a config file at `conf/settings.yaml`. If it finds the file, a custom class converts the config to an object and that is returned and used exactly the same way as the `parse_args` object. If a config file is not found arguments are loaded from the command line.
- Added PyYAML 3.11 to the requirements.
- Added example settings file
- Updated gitignore to ignore `conf/settings.yaml`

PR #160, opened 2016-07-18
Settings File

JZ6:
This pull request includes a

- [ ] Bug fix
- [x ] New feature
- [ ] Translation

The following changes were made

-added args_from_file
added settings.txt

Need testing, wrote on mobile.

PR #968, opened 2016-07-21
Yaml config file

MelonSmasher:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Created a yaml config file that can be used instead of command arguments
- Added a switch to the program called `-ic` `--ignore-config` that will ignore the config file, and makes the program expect command line arguments.
- If there is no config file the program expects command line arguments.
- Moved google maps api key into config file.
- Removed `load_credentials()` function and call. This is now taken care of by `get_args()`
- Organized a bit of code related to the api key in `runserver.py`.
- The config file is located in `config/config.yml`.
- Created example config file.
- Added `config/config.yml` to gitignore

Related to: #940, #233, #160 ,#142

~~In a previous PR (#233), I was going to implement a YAML config file, I decided against this since the project is already including the JSON library. It seemed silly to bring another dependency into the project for a task that could be handled by something that was already in the project.~~

Sorry about the commit chain, I had to rebase.

Edit: changed JSON to YAML.


MelonSmasher:
@DerScreever this commit should replace #233.

Ashex1:
What's the reason for using json over yaml?

MelonSmasher:
JSON was already used for a credentials.json file, and the JSON library is already used. I did not see a need to create another external dependency for this project.

MelonSmasher:
@Ashex1  @DerScreever I can switch this over to YAML in a jiff, please hold.

MelonSmasher:
All set with b30a283. The program now reads in a YAML file.

Ashex1:
It would be nice to have the pgo auth nested but looking at how you're parsing the config we should leave it as is. 

Would this method be able to handle nested entries?

MelonSmasher:
@Ashex1 I agree, right now this config is not exactly using YAML to it's full potential. What do you guys think about this: https://pypi.python.org/pypi/yconf

It might help kill two birds with one stone and keep things clean down the line as more options get added.

Ashex1:
@MelonSmasher I like the idea of using yconf to extend this further for the desired result. 

I'd propose we merge this as is to get everyone used to YAML config then submit a new PR with yconf since you'll need to jiggle the utils module to be in a class. If you see this change as an easy one then we could get it in now.

AHAAAAAAA:
+1

MelonSmasher:
@AHAAAAAAA, just wondering why this commit was reverted?

JZ6:
@MelonSmasher broke too many things, i ll look for a fix

MelonSmasher:
Let me know if I can do anything to help.

PR #976, opened 2016-07-21
Resolve bug with lot of expires pokemons

realmyst:
This pull request includes a

- [ ] Bug fix

Remove excess params from pokemonLabel function. It resolve bug with disappears time at map. See function definition: https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/static/map.js#L58


bocaci:
+1, thank you!

sykhro:
Already fixed

PR #1000, opened 2016-07-21
Added No-Server mode

LTHeaven:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

-  Added No-Server mode (flag -ns). Effect: Only starts the searcher but not the webserver

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -3,6 +3,7 @@
 
 import os
 import logging
+import time
 
 from threading import Thread
 
@@ -60,4 +61,8 @@ if __name__ == '__main__':
         config['GMAPS_KEY'] = args.gmaps_key
     else:
         config['GMAPS_KEY'] = load_credentials(os.path.dirname(os.path.realpath(__file__)))['gmaps_key']
-    app.run(threaded=True, debug=args.debug, host=args.host, port=args.port)
+    if args.no_server:
+        while True:
+            time.sleep(1000)
+    else:
+        app.run(threaded=True, debug=args.debug, host=args.host, port=args.port)
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -34,6 +34,7 @@ def get_args():
     parser.add_argument('-c', '--china', help='Coordinates transformer for China', action='store_true')
     parser.add_argument('-d', '--debug', help='Debug Mode', action='store_true')
     parser.add_argument('-m', '--mock', help='Mock mode. Starts the web server but not the background thread.', action='store_true', default=False)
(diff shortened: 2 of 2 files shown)


JZ6:
Woo pr #1000!

JZ6:
is there a better way of implmentation without using time.sleep?

MonkeyLeeT:
Why not just
`If not args.no_server:  
  app.run()`
?

LTHeaven:
i have to keep the main thread alive thatswhy the while loop.

MonkeyLeeT:
I'm not most familiar with python. Is that how to keep daemon thread alive, by sleep(1000)?

LTHeaven:
the problem is when the main thread ends the search thread also ends. so i make a loop and use time.sleep to reduce the interpreter stress

invisiblek:
what about this? https://gist.github.com/invisiblek/e6f6cec123ec4431ee48475c48d20aa6

devinsba:
@LTHeaven is your plan to share a database on one web server with a bunch of searchers?

kingvibs:
@LTHeaven @invisiblek if that is the case then why not implement use of DHT (distributed hash table) or similar, so nodes can pop in and out and reduce your server loads

devinsba:
@kingvibs That seems like not a good fit for the problem. Could easily be done with a simple change in database type. Ie: switch to not sqlite (mysql, postgres, cassandra, etc.)

runserver.py
```diff
@@ -60,4 +61,8 @@ def start_locator_thread(args):
         config['GMAPS_KEY'] = args.gmaps_key
     else:
         config['GMAPS_KEY'] = load_credentials(os.path.dirname(os.path.realpath(__file__)))['gmaps_key']
-    app.run(threaded=True, debug=args.debug, host=args.host, port=args.port)
+    if args.no_server:
+        while True:
+            time.sleep(1000)
+    else:
+        app.run(threaded=True, debug=args.debug, host=args.host, port=args.port)
```
devinsba: I think the solution posted by @invisiblek is really how you want to do this

https://gist.github.com/invisiblek/e6f6cec123ec4431ee48475c48d20aa6

JZ6:
@invisiblek could u do a pr with that

LTHeaven:
@kingvibs @invisiblek today is the first time writing python for me and using git so you guys can take it on from here. im going to bed soon^^

MonkeyLeeT:
@invisiblek We'll be ready to take your changes.

@LTHeaven Thanks for understanding bud :D

LTHeaven:
no problem. i just saw this missing and implemented it JZ6 then said i should commit it

PR #1012, opened 2016-07-21
Add Dockerfile

jnovack:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Added Dockerfile

For developers, Docker is love, Docker is life.  Embrace, Adapt, Conform.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM python:2-alpine
+
+# Install dependencies
+RUN apk --update add --virtual build-deps openssl ca-certificates python-dev build-base gcc
+
+# Copy and install requirements.txt only to minimize image rebuilding on updates
+COPY requirements.txt /myapp/requirements.txt
+WORKDIR /myapp
+RUN pip install --no-cache-dir -r requirements.txt
+
+# Remove unneeded caches
+RUN apk del build-deps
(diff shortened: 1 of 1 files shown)


jnovack:
Two discussion points:

* Perhaps the README should be updated as well?
* Can we use environment variables AS WELL AS command line arguments?

# (Re-)Building
    git pull
    docker build -t pogomap .

# Running
    docker run -it --rm pogomap python runserver.py -u USERNAME -p PASSWORD -a ptc -l "LOCATION" -st STEPS

DanielVolz:
Can you explain me why docker is so good? What advantages does it over the "normal" clone git? Is it about dependencies?

jnovack:
The purpose of docker is to run it the container so that I do not have to pollute my drive with python or it's dependencies OR If someone has multiple versions of python, or multiple versions of the requirements.

The point is that it runs isolated in a container.  Not affecting the host OS at all.

The only change that would NEED to be made for the upstream project is binding to the network card rather than localhost (or allow the OPTION of binding to network interface rather than localhost).

DanielVolz:
Ahh I understand. Yes this seems reasonable. Thanks for the explanation!

MonkeyLeeT:
#599 will be the starting point for all Docker support.

PR #90, opened 2016-07-18
Use in game sprites for map markers

hhu94:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

- Now uses sprites extracted from the game assets for the map markers.
- Updated screenshot in README accordingly.

Please tell me if you like the size of the sprites. They were originally 128x128 but I resized them to ~~50x50~~ 65x65 because I thought they were too big. I still have the original 128x128 sprites.

cover

PR #1016, opened 2016-07-21
Change some icons more clearer

qfdk:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Change some icons more clearer

Change some icons more clearer


qfdk:
I put the original file on format PNG in my site 
[Download here](http://qfdk.me/imgs.zip)
qq20160721-0 2x

hhu94:
I had already done something similar to this (see #90), then for some reason people thought the smaller (current) version looks better (see #166).

I much prefer having the black outline I added to the icons, as it makes the pokemon easier to distinguish on the map.

PR #1024, opened 2016-07-21
Add no-webserver mode

invisiblek:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

-
-
-

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -36,6 +36,7 @@ def get_args():
     parser.add_argument('-c', '--china', help='Coordinates transformer for China', action='store_true')
     parser.add_argument('-d', '--debug', help='Debug Mode', action='store_true')
     parser.add_argument('-m', '--mock', help='Mock mode. Starts the web server but not the background thread.', action='store_true', default=False)
+    parser.add_argument('-ns', '--no-server', help='No-Server Mode. Starts the searcher but not the Webserver.', action='store_true', default=False, dest='no_server')
     parser.add_argument('-k', '--google-maps-key', help='Google Maps Javascript API Key', default=None, dest='gmaps_key')
     parser.set_defaults(DEBUG=False)
     args = parser.parse_args()
--- a/runserver.py
+++ b/runserver.py
@@ -3,6 +3,7 @@
 
 import os
 import logging
+import time
 
 from threading import Thread
 
@@ -16,6 +17,7 @@ from pogom.pgoapi.utilities import get_pos_by_name
 
 log = logging.getLogger(__name__)
 
+search_thread = Thread()
 
 def start_locator_thread(args):
     search_thread = Thread(target=search_loop, args=(args,))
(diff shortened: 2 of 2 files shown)


invisiblek:
@JZ6

MonkeyLeeT:
Tested, works for me and backend db is being updated.

One more collab review please.

Bitti09:
How to show the results?
Just running a second instance without the flag in same dir?

jangrewe:
Does this mean it's possible to have multiple searchers update the same db?
E.g. run one app as searcher & webserver, and X nodes only as searchers, writing to the same file?
Or are we getting MySQL/MongoDB/whatever as a backend soon? :grin:

invisiblek:
@Bitti09 - you could monitor the database with sqlite3 and see things change

@jangrewe - in the long run yes, i think that's the plan here. but as far as i know there isn't anyone working on a replacement back end (yet)

MonkeyLeeT:
@jangrewe For now we don't wanna break the backend without db adapter for it. If you wanna work on a adapter and make it generic, it'd be wonderful.

jangrewe:
@MonkeyLeeT my Python skills are a bit lacking, not sure if it's a good idea to break other people's shit :grin:

euphwes:
@jangrewe It appears that way. I have 3 instances of this app running on my home PC (centered on my home, my work, and a nearby park). They are all running out of the same directory (the repository root), and therefore they are all updating the same local database file. When I connect to the web app for one of those instances, I can see pokemon/stops/etc for all 3 locations on the single map. It's pretty snazzy. I'm not even sure it was intended to work that way, but it certainly *does* work.

jangrewe:
@euphwes Yea, that happened to me, too - by accident, i swear! ;-)
I'm just not sure SQLite will cope with more than 2-3 concurrent writes, that's what RDBMS or at least some NoSQL DB is for :grin:

LTHeaven:
Im currently running this 32 times on my server http://heavensgat.es:5000/
zoom out a bit im mapping to citys and a little area

callmebackdraft:
Whenever i try to run with the -ns argument, it returns the error below, is this something i am doing wrong ? running it without the -ns flag runs as expected.


`Traceback (most recent call last):
  File "runserver.py", line 66, in <module>
    search_thread.join()
  File "C:\Python27\lib\threading.py", line 929, in join
2016-07-22 02:08:02,618 [   auth_ptc] [   INFO] PTC login for: ########
    raise RuntimeError("cannot join thread before it is started")
RuntimeError: cannot join thread before it is started`

AHAAAAAAA:
I'm also running into the same issues as @callmebackdraft

invisiblek:
added a loop if we enable no-webserver mode, will loop until search_thread is started before attempting to join the thread. seems to have fixed it on my end (must be racy)

AHAAAAAAA:
Perfect, will be using this during the weekend to collect data. Original reason I made this was so I could play with the data and do some data science on it.

PR #1068, opened 2016-07-21
Don't kick the server while it's down

hgentry:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- When the search thread detects that the server is down, wait a bit before trying again. There are so many people running this that we are now probably a not insignificant chunk of Niantic traffic; we don't need to keep punching the servers repeatedly while they are trying to come back up.

This will also shut up the dozens of issues about the response dictionary key error. ;)

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -14,6 +14,7 @@ log = logging.getLogger(__name__)
 
 TIMESTAMP = '\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000'
 REQ_SLEEP = 1
+failed_consecutive = 0
 api = PGoApi()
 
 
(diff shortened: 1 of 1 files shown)


JZ6:
Nice idea! Although 15 seconds might be too long of a retry period, maybe 5 seconds would be better

JZ6:
+1 ready to merge

JZ6:
One last change since wit time is decreased we can decrease failed times too
5 times should be enough 
Sorry promise last change!

Maxincredible52:
this is a good idea! +1

CarlTaylor1989:
+1

MonkeyLeeT:
Good one. Merged.

PR #1078, opened 2016-07-21
Added if statement to avoid UnboundLocalError

SpidermanAlliance:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Added if statement to avoid UnboundLocalError: local variable 'failed_consecutive' referenced before assignment

Based off of: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1068

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -84,8 +84,8 @@ def search(args):
             parse_map(response_dict)
         except KeyError:
             log.error('Scan step failed. Response dictionary key error.')
-            if not (failed_consecutive):
-                failed_consecutive=0
+    
+            global failed_consecutive
             failed_consecutive += 1
             if(failed_consecutive >= 5):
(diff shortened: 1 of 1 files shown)


MonkeyLeeT:
http://stackoverflow.com/questions/9264763/unboundlocalerror-in-python

From this repo, it seems that failed_consecutive outside the function will simply not be used. So either we should remove that outside declaration and initiate it inside, or we should use global keyword.

HenrySia:
Even after your fix, this issue still repros. I think the right fix is to insert:
`global failed_consecutive`
before
`failed_consecutive += 1`
See: http://stackoverflow.com/questions/855493/referenced-before-assignment-error-in-python

SpidermanAlliance:
@HenrySia Updated. Seems to be running okay for me now.

HenrySia:
@SpidermanAlliance thank you for the fix!

PR #1052, opened 2016-07-21
Optional larger icons & icons in info popover

parkershepherd:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

- Added large icons
- Added toggle to change (pokemon) icon size
- Updated localstorage code so pokemon, pokestops, gyms defaulted to 'true', and large icons defaults to 'false'
- Added icons to pokemon info popover

image

If this is related to an existing ticket, include a link to it as well.

PR #1091, opened 2016-07-22
Fix localStorage options always defaulting to `true`

billyvg:
This pull request includes a

- [X ] Bug fix
- [ ] New feature
- [ ] Translation

Check if `localStorage` var is explicitly undefined, otherwise when you turn off switch, it will always default to true

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -234,10 +234,9 @@ function clearStaleMarkers() {
 };
 
 function updateMap() {
-    
-    localStorage.showPokemon = localStorage.showPokemon || true;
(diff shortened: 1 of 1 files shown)


JakeCooper:
localStorage does not actually store boolean values, only strings. A guard can be efficiently utilized to check for null, because localStorage["value"] will never equal false, only "false". Example below.
image

parkershepherd:
Just fyi, https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1052 also accounts for this

billyvg:
@JakeCooper Ah makes sense, closing.

PR #1093, opened 2016-07-22
Reduces search step length to find all Pokémons.

aedm:
This pull request includes a
- Bug fix

- Reduces search step distance from 0.0025 to 0.002. Previous value skipped some Pokémons on the map.


MonkeyLeeT:
People are working on the new (and hopefully final) scanning algorithm. Feel free to throw a quick adjustment of step length after that.

PR #1098, opened 2016-07-22
Rename credentials.example.json to credentials.json

FrostTheFox:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Changed credentials.example.json to credentials.json
-
-

Allows for APIKeyException to happen, giving a more helpful message telling user to go check the wiki to get an GMaps API key.


algoni:
Now credentials.json is both in .gitignore and in the repository, and pulling this wiped my API key from the file.

Maybe resolve this in a different way, perhaps the python script could create credentials.json if it does not exist, or create a new exception type for missing creds file, with a user-friendly message?

toumaz:
Agreeing with @algoni. This seems like a weird way to attack the problem, and one that makes it actively harder to pull in new changes for others, too.

@FrostTheFox I've opened PR #1238 as an alternative solution. Do you think that could work?

PR #1073, opened 2016-07-21
New feature: Email when a rare pokemon is found

mikehow1984:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

-added method to models that lets you use your GMail to mail yourself when a rare Pokemon is found
-the credentials json now includes gmail name, gmail password, and rare pokemon fields (a list of pokemon you want notifications for)
-made it so that credentials now load correctly in the load_credentials method

PR #1106, opened 2016-07-22
Email Alerting for Rare Pokemon

patrickread:
H/T to @mikehow1984 [for the SMTP code](https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1073)! Future contributions welcome from anybody.

This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added an `alerts.example.json` config file, where you designate ID's of rare Pokemon on which you'd like alerts. I used ID's to avoid complexities with localities, though using names might be less obnoxious.
- Add Gmail username and password (preferrably, use an [app specific password](https://support.google.com/accounts/answer/185833?hl=en)) here. A better way to do this might be to use an env var, so it's more deployable.
- The alerts use a different thread from the searching and web server. This allows the others to run uninhibited by any emailing efforts
- This also uses a new DB field on the `Pokemon` model called `alerted`. As the name suggests, it'll be True if you've already received an alert on this particular encounter.


AHAAAAAAA:
Putting this on hold till the develop branch is stable and complete enough.

patrickread:
@AHAAAAAAA Any interest in revisiting this?

PR #1109, opened 2016-07-22
Fix issue with changing non global variable when Niantic servers fail

gsabran:
This pull request includes a

- [ x] Bug fix

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -56,6 +56,7 @@ def login(args, position):
 
 
 def search(args):
+    global failed_consecutive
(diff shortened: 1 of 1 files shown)


hgentry:
How is this functionally different from what's already there? failed_consecutive is created at global scope on line 17

colinchilds:
@hgentry, without this, failed_consecutive is local scope within this function and throws an error

MonkeyLeeT:
Done in #1078. Thanks for PR.

PR #1118, opened 2016-07-22
Added notification system to notify people about new pokemons.

PiaFraus:

Added notification feature. Uses Notification API, so it should work in all modern browsers.
 Even when page is in background this will notify user about new pokemons. Has switch to disable it.


Whoknowsit:
+1

JoshuaSoileau:
+1

AHAAAAAAA:
#1149 solved this issue

PR #1119, opened 2016-07-22
Use Babel

arao456:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Add babel to grunt task
- Use babel to transform app.js and map.js

Someone in Discord was mentioning broken string templating. I think most browsers support this but not all do. Plus you can start using arrow functions and all that good stuff :)


BillyWM:
It's IE11 and the stock Android browser that break because of this particular feature. Related PR: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/871

No need to rewrite that bit if you're using Babel.

nekitk:
maybe also use babel-polyfill?

m1so:
Could this be merged after resolving conflicts? Transpilation to ES5 would fix issues on mobile devices that don't support ES6 like Safari on iOS.

MonkeyLeeT:
#1119 #1140 #1156 

ravishivt:
I'm fine with this one.  My #1140 also fixes jsHint reported issues but I can put those in a separate request.

BillyWM:
Please also add babel-polyfill per @nektik's suggestion to whichever one is merged. This gives you the highest browser compatibility by ensuring certain methods exist as well as bigger things like Promises.

arao456:
i think we should go with ravishivt at #1140. It's cleaner to use the IIFE than using my "es2015-without-strict". I'm going to close this one :+1:

PR #1133, opened 2016-07-22
Update search.py

t3r3nc3101:
Fiddling with settings


MonkeyLeeT:
Why this change?

macgngsta:
PTC is actually the one that has an expiry on its auth token. Google doesn't care, but there is a limit of 25.

https://developers.google.com/identity/protocols/OAuth2

not sure what the expiry time is without intercepting the call to `https://sso.pokemon.com/sso/oauth2.0/accessToken` , but I assume that the 1000 was there to convert from ms to seconds.

PR #1140, opened 2016-07-22
Use grunt-babel and modified jshint to correctly target JS source.

ravishivt:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

- Added grunt-babel to the project, which is used to transpile ES6/ES2015 code to ES5.  This fixes issues with older browsers that don't support ES6 (such as iOS 8 users).
- Moved map.js from static/ to static/js as that is where app.js lives.
- Updated gruntfile to run jshint against the source code.  Previously not checking anything.
- In jshint task, use esversion: 6 to check ES6 syntax.
- Cleaned up all the jshint reported errors with app.js and map.js.
- Grunt now minifies (uglifies) map.js as well.


BillyWM:
re:map.js - Just my 0.02 but as long as you're using Babel, might as well use a Class instead of an IIFE.

ravishivt:
@BillyWM I'm just trying to introduce ES6 transpilation with minimal changes to the existing source as there were already references to ES6 template strings that were breaking in older browsers.  In the future, I agree we could make a lot of ES6 related improvements to the source.

MonkeyLeeT:
#1119 #1140 #1156 

BillyWM:
Probably best to pull this one, and have @ravishivt add babel-polyfill as @nekitk suggested in #1119.

Babel transforms syntax, but the polyfill is needed in addition for the widest browser compatibility.

#1156 can be optional depending on whether it merges cleanly on top. Autoprefixing is fine but useless prefixes don't hurt anything (they're ignored by browsers that don't understand them). More important is getting ES6 transpilation set up.

MonkeyLeeT:
@ravishivt Can you rebase? We'll take this one then.

ravishivt:
Just opened a new pull request without the jsHint changes, #1675.

nivenj:
Tried the changes, fixed my issue of not being able to render the map on older browsers, but had to revert as change location and pokemon notify where not visible on filter.

PR #1148, opened 2016-07-22
Modified Load Credentials to Avoid Overwriting API Key

krazedkat:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

-Modified utils.py to allow for credentials.example.json to be the name of the file while still getting the favourable APIKeyException error.
-Renamed credentials.json to credentials.example.json

This change allows for credentials.json to not be overwritten by pulling updates.


quagliero:
+1

I don't know why credentials.json isn't .gitignore'd anyway?

quagliero:
Isn't credentials.json in the .gitignore anyway? I can see it in there yet any changes I now make to it are tracked in the working directory (on current develop branch)

+1 for this PR as we shouldn't be committing API keys.

Ashex1:
credentials.json is in .gitignore so it shouldn't be a problem. I have no idea why it was changed from credentials.example.json (much less why that pr was merged). 

You're renaming the file so that it will fail, it's more appropriate to handle the IOError exception.

quagliero:
@Ashex1 It seems that since the file has been `git mv`d from credentials.sample.json to credentials.json, that changes to it are now picked up in the working tree. e.g. if I edit the file/add my API key and run a `git status` it shows credentials.json as a modified file to be committed - when it shouldn't show anything. 
screen shot 2016-07-22 at 09 14 57

BaIthamel:
That doesn't solve the original issue that caused the json to be removed, we need to populate the folder with the file for those users who do not understand how to code. This is to prevent devs killing people :P
.gitignore does maybe make this superfluous. Ah yes that was why.

Ashex1:
Sensible/informative error messages are what's needed for users who do not understand how to code. If the file is missing the error should tell them what is wrong and they will rename it. 

Additionally, #968 will do away with this file in favor of a config file for holding all settings.

BaIthamel:
You tell that to everyone who has to man the help chat! People can't follow instructions to save their life! they see an error they panic they don't do what it says or google it.

JZ6:
outdated, credtials.example is too confusing

PR #1149, opened 2016-07-22
Add notifications on Pokemon spawn

andrewda:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Add a notification system
- Add a multi-select to choose Pokémon which, when spawned, will create a notification
- Tested and works in Chrome, Chrome Mobile, and Chromium (including Electron - will fully support Electron ports of PokemonGo-Map)
- Should work on all modern web browsers
- Should work on most mobile browsers (Chrome, Firefox, Android WebView)
- Sound supported in Chrome

This is similar to #1118, but is cleaner and allows custom Pokemon notifications.


JZ6:
+1 need another colab approval then merge

therealshibe:
Howto set sound?

andrewda:
@therealshibe I don't believe notifications come with a sound by default. I can probably add this later, but it's most likely not top-priority, either.

therealshibe:
Okay I cloned your repo and chrome toast notifications are working for me.

ChristianLuxem:
The localization is missing here, the filter will also show only english Pokemons. I think you should add this..

static/map.js
```diff
@@ -1,4 +1,16 @@
...
+
+    if (Notification.permission !== "granted") {
+        Notification.requestPermission();
+    }
+});
+
 var $selectExclude = $("#exclude-pokemon");
+var $selectNotify = $("#notify-pokemon");
 
 $.getJSON("static/locales/pokemon.en.json").done(function(data) {
```
AHAAAAAAA: Can we get a:
$.getJSON("static/locales/pokemon."+locale+".json").done(function(data) {

instead, also appropriate changes to app.py and map.html to make this possible!

AHAAAAAAA:
Merged another PR fixing localization, merging this now.

MatthiasJ:
could you tell me, how to use that feature? as command line param or inside gui?
thx

HenrySia:
@andrewda Hi Andrew, i'm pretty sure this worked before, but I did a git pull recently and I couldn't hear the notification sound anymore. Can you perhaps take a quick look? Thank you very much for the good work!

PR #1156, opened 2016-07-22
Use autoprefixer

nekitk:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Used autoprefixer for vendor prefixing
- Tabified indentation in changed files for consistency with original style


MonkeyLeeT:
#1119 #1140 #1156 

nekitk:
@MonkeyLeeT this pr is about css vendor prefixing, while #1119 and #1140 are for js ES6 support
we can freely use both autoprefixer and babel since they don't interfere in any matter

PR #1161, opened 2016-07-22
New feature + few fixes.

MRokas:
- [x] Bug fix
- [x] New feature
- [ ] Translation

- Added switch for Pokemon Info Window. 
- Fixed `show*` always defaulting to true upon calling `updateMap()` method
- Disabled autoPan for InfoWindows. There won't be random map panning on accidental marker hover, or pans whenever new marker is added if autoPIW is enabled.
- Fixed some badly styled code.


andrewda:
There are a ton of changes here which should probably be separated into separate PRs.

AHAAAAAAA:
It works. A few things:

* It shouldn't default to having the Pokemon infobox shown.
* Clear infoboxes option? Should this happen with toggle?

MRokas:
@andrewda - yeah you might be right, sorry!
@AHAAAAAAA - that was my personal opinion, but both of these can be easily changed. 7 lines change only.

AHAAAAAAA:
Can we separate the show infobox switch from the other stylistic fixes? The autoPan fix and bad style can be added now, but the infobox switch we can put on hold until develop's first release soon.

MRokas:
@AHAAAAAAA - I've resubmited InfoBox switch(with close all of them button) - https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1481

Issue #1059, opened 2016-07-21
Blank map has returned


lnmtriet93:
I just tested the latest pull and it was a blank map at localhost:5000

xijbx:
Yeah I'm getting this issue too - API key tried in both credentials.json and on command line.

DerScreever:
for me work. im on latest commit
http://85.214.145.87:5000/

sykhro:
We are aware of the issue, it's related to local storage. Should be fixed soon.

sykhro:
Please purge your cache, cookies and local storage. It should solve the problem

xijbx:
Can confirm purging cache and local storage fixes this - to anyone unsure you can do it all within Chrome dev tools panel. Hold on the refresh button and you'll have options to hard refresh & empty cache. And you can delete local storage from the resource tab of dev tools.

lnmtriet93:
It worked! I confirm that purging cache and local storage helps fix this bug

PR #1162, opened 2016-07-22
check for 'undefined' value in map_style

### LINKED ISSUE #1154: Google map bug ?
Hi,

I've git pull and correct my gmaps ID, beceause "credentials.json" are replaced. But impossible to load the map. All pokemon/gym/stop are Okay.

Sorry for my bad english x)

skyshard:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

At some point the string 'undefined' was being written to map_style, causing maps to not load. This adds an additional check for that value

Fixes #1154 and may be the cause behind #1059, #1165

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -68,7 +68,9 @@ function initMap() {
         localStorage['map_style'] = this.mapTypeId;
     });
 
-    localStorage['map_style'] = localStorage['map_style'] || 'roadmap';
+    if (!localStorage['map_style'] || localStorage['map_style'] === 'undefined') {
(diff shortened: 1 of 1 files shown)


BillyWM:
IMO the real solution is to tell users to clear their localstorage, but with so many non-technical end users flooding in you'll get 20 tickets anyway....

End users should be kicked out of develop per https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1186 and then this check should be reverted. The string "undefined" isn't getting written to storage anymore, and if it were, this would just hide an underlying issue.

PR #1181, opened 2016-07-22
Fix exception concatenate with str

Mr-Groch:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- fixed exception concatenate with str


medghaim:
+1 - wanted to do this earlier as I ran into that, but totally forgot.

PR #1182, opened 2016-07-22
Change Scan location and SSL support

weissi1994:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

- Added js Function in map.js to perform post requests
- Added py Function in app.py to set new location
- Added Error Handling for json Decode Error
- Added SSL Support


static/map.js
```diff
@@ -93,6 +117,13 @@ function initMap() {
     map.addListener('maptypeid_changed', function(s) {
         localStorage['map_style'] = this.mapTypeId;
     });
+    
+    google.maps.event.addListener(map, 'dblclick', function(event) {
+        console.log(event.latLng);
```
MRokas: `console.log()` shouldn't exist in prod. Forgot to remove I suppose?

medghaim: 'tis a `develop` branch after all!

weissi1994: Yes, forgot to remove it, its gone now

fettlaus:
I don't really get why you created `set_loc(self)` instead of using or improving `next_loc(self)`. As I can see it, they are only slightly different.

weissi1994:
You are right, I just didnt realize this function already existed :/

static/map.js
```diff
@@ -121,7 +121,7 @@ function initMap() {
     google.maps.event.addListener(map, 'dblclick', function(event) {
         var lat = event.latLng.lat();
         var lng = event.latLng.lng();
-        post('/next_loc', {lat: lat, lng: lng});
+        post('/next_loc', {'lat': lat, 'lon': lng});
```
MRokas: Was this really necessary? It also goes against most code-styles.

Ryan-Myers:
+1 for SSL Support. I've got my server set up with a pinned SSL cert, so it's been a pain to workaround it for this app. Thanks!

static/map.js
weissi1994: yes it was^^, i get confused with all the coding languages i use all day

AHAAAAAAA:
Change location implemented in #1225. Can you separate out the SSL functionality and do a PR for that?

Issue #1163, opened 2016-07-22
Suggestion: As a user I'd like to see a list of all nearby Pokemons


Wojozoo:
Hi,

It would be great if we could add a list on the right hand side of all pokemons displayed on the map.
This will make it a lot easier to see what kind of pokemons there are nearby. 
Next it would also be helpfull to be able to filter on this list, so if I click on one pokemon in the list only this pokemon will show up in the map.

MRokas:
I will get it done later today.

MRokas:
Nvm, someone already has done it - https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1184

bytewright:
should be quite easy, but I'm really not the right man for this job :( (only know basic html stuff)

PR #1184, opened 2016-07-22
Mobile mode with min data usage

bytewright:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- added a new 'mobile' mode
- goal was to have a mode, which uses as little mobile data as possible and which just helps to find specific pokemon. Feels less cheaty that way. 
- mobile users visit '/mobile' and get a list of pokemon in the area. only seeing pokemon icon (using the lower size pixel icons), remaining time and distance from scan origin. 
like this: http://i.imgur.com/ZBZNBsf.png
- on android phones, clicking on the icon opens google maps app with a placed marker on the pokemon location


bytewright:
oh i forgot: also chnaging flask logging to only display errors. I didnt wanted to be informed about every get request...

Wojozoo:
@bytewright 
Can you have a look at issue #1163 ?
Can we use the same codes to display a filterable list on the right hand side of the desktop map?

chuyskywalker:
This view would also be really nice for people who are less-sighted and playing with friends, since their screen reader will work with it (where it doesn't at all with the map interface, so I'm told).

PR #1197, opened 2016-07-22
Add an option to runserver.py that enables CORS

msaspence:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Adds a dependency on flask-cors
- Adds a cli flag of `-C, --cors` which if add adds flask cors to the app
- This allows clients running on a different host to access the api

### DIFF
--- a/Easy Setup/requirements.txt
+++ b/Easy Setup/requirements.txt
@@ -13,3 +13,4 @@ requests==2.10.0
 s2sphere==0.2.4
 gpsoauth==0.3.0
 protobuf-to-dict==0.1.0
+flask-cors==2.1.2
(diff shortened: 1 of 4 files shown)


dtphelan:
In case others have trouble, this temporarily broke my install. If you're updating from an older commit, you'll have to run `pip install -r requirements.txt` again after.

kingeddard:
@msaspence what exactly would CORS be needed for? I currently access the server from my phone and i am able to change the location search from there. also have 2 no-server searchers. would this help me?

msaspence:
If you're building a client that is delivered to the browser on a different
port or host. I'm using it to run a web pack development server to build a
client that can still access the Json endpoints.

PR #1212, opened 2016-07-22
Cleanup search

wcastello:
This pull request includes a

- [x] Cleanup

- Some magic numbers dealing with the search such as the time to sleep on heavy load and the usual time to sleep between requests. 

- search() was outputting things like log.error('Niantic servers under heavy load. Waiting before trying again') and then advancing the step nonetheless. Now it keeps trying the step, i.e. sends request, get the dict, try to parse, until the parse_map() is successful. Still does the heavy load sleeps if the parse fails more than config['REQ_MAX_FAILED']).


wcastello:
Dammit, this went to the wrong branch.

PR #1213, opened 2016-07-22
Cleanup search.py

wcastello:
This pull request includes a

- [x] Cleanup

- Move magic numbers dealing with the search such as the time to sleep on heavy load to the config. 

- Gets rid of global var failed_consecutive.  

- search() was outputting things like log.error('Niantic servers under heavy load. Waiting before trying again') and then advancing the step nonetheless. Now it keeps trying the step, i.e. sends request, get the dict, try to parse, until the parse_map() is successful. Still does the heavy load sleeps if the parse fails more than config['REQ_MAX_FAILED']).

### DIFF
--- a/pogom/__init__.py
+++ b/pogom/__init__.py
@@ -8,4 +8,7 @@ config = {
     'ORIGINAL_LATITUDE': None,
     'ORIGINAL_LONGITUDE': None,
     'GMAPS_KEY': None
+    'REQ_SLEEP': 1,
+    'REQ_HEAVY_SLEEP': 5,
+    'REQ_MAX_FAILED': 5
 }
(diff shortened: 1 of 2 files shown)


pogom/__init__.py
```diff
@@ -7,5 +7,8 @@
     'ROOT_PATH': None,
     'ORIGINAL_LATITUDE': None,
     'ORIGINAL_LONGITUDE': None,
-    'GMAPS_KEY': None
+    'GMAPS_KEY': None,
+    'REQ_SLEEP': 1,
+    'REQ_HEAVY_SLEEP': 5,
```
AHAAAAAAA: Heavy sleep should be longer, 30 seconds?

wcastello: 30 sounds reasonable :) I just left whatever it was before.

AHAAAAAAA:
+1

PR #1238, opened 2016-07-22
Revert credentials.json rename and handle missing credentials better

toumaz:
This pull request includes a new feature/bugfix.

Relating to PR #1098 (see the comments made there too), I find the given fix of renaming `credentials.example.json` to `credentials.json` a rather blunt one, since it'll wipe credentials for anybody pulling in changes and generally make it harder for contributors to test out of the same root.

This properly handles the case where `credentials.json` does not exist and defaults the loaded credentials to an empty dictionary instead.

It also uses `get` instead of a direct key lookup to provide a default value to check against, so the more useful `APIKeyException` is raised in the case of an empty (but existing) credentials file, too.


FrostTheFox:
Looks good. I only made that for time time being because it wasn't built to handle credentials.json not existing.

Ashex1:
The use of a default value is clever but this masks the issue of a missing file. The only error they'll get is that there is a problem with the API key.

You could add a message to the IOError handler that the file is missing, this way a person see's a message indicating a missing file. This works but then it looks silly since they first get `missing file` then `bad api key`.

toumaz:
@Ashex1 Agreed. My thought process was that that'd delve a bit too far down the rabbit hole for what I'd classify as a "quick fix". Since the file doesn't *need* to exist in the first place, I felt like a more conservative approach to the problem was best - and better than throwing an even more cryptic `KeyError`.

This also doesn't handle the case of a JSON file that fails parsing - but hey, neither did the original code.

Would you deem this "good enough" for a single unit of work and a merge, then attack the problem further in follow-up PRs?

toumaz:
Or, to elaborate a little (and sorry for the double comment): since the file *only* contains the API key right now, and the error message directs to the wiki which contains instructions for how to create the file, I personally felt it was a toss-up and almost a moot point.

Ashex1:
Yup, that makes sense to me. There's been a _lot_ of activity around the credentials.json recently so this will likely change somehow in the comings days. 

+1

PR #1242, opened 2016-07-22
Utilities, location not found by name and empty location fix

wcastello:
- [x] Bug fix

- Fix exception when position is not found by name
- Don't let utilities.get_pos_by_name() run into an empty location_name or it will raise geopy.exc.GeocoderQueryError: HTTP Error 400: Bad Request

You can reproduce the problems running python runserver.py -u xxx -p "yyy" -l "" -st 5 or python runserver.py -u xxx -p "yyy" -l "this location won't be found" -st 5 for example.


wcastello:
I thought about defaulting to some location but since -l is not optional I thought an empty or not found location should not process either.

AHAAAAAAA:
+1

PR #1225, opened 2016-07-22
Added change location functionality to filters

EgorDm:
This pull request includes a

- [ ] Bug fix
- [x ] New feature
- [ ] Translation

The following changes were made

- Added move to location function to frontend.

If this is related to an existing ticket, include a link to it as well.

PR #1252, opened 2016-07-22
Add master password for next_loc

MichaelErmer:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added optional argument -M/--master, which is required to be passed to /next_log as master= when enabled.
- The UI option is only visible if a master PW is set
-


AHAAAAAAA:
Function? Why do you need to pass in passwords like that?

EgorDm:
Nice idea. Would come in handy if the map is public.

MichaelErmer:
If you share the map with multiple people, you can set a master password so only the person with the master password can alter the location. Adding interface option now.

MichaelErmer:
Waiting for https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1225 to be merged, will ad interface for the password afterwards.

Bakies:
My map is semi-public, right now no one has fucked with it but having the tiny bit of security would be nice

Ashex1:
A reverse proxy is better suited for this use case as you're password protecting a specific uri. There's already a pending PR#717 adding basic auth.  Additionally this password is being passed as a URI parameter which offers no real security since this can pulled from network traffic.

This change is redundant and unnecessary.

MichaelErmer:
@Ashex1 WTF, thats two entirely different uses cases... 

HTTP Auth is fine for preventing public access, this change is to add a password for the coming track geo location and set geo location features. Unless you want to add a fully fledged user management this is a convenient way to control who can change the instances location.

MichaelErmer:
@CarlTaylor1989 Done

MichaelErmer:
@welshboy  its not in there yet, check https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1225 only the server part was merged yet.

welshboy:
@MichaelErmer Ah yeah I can see now. My bad! Was too eager for it

Ashex1:
@MichaelErmer Sorry that comment sounded super bitchy. The primary concern is that there will be little flags for securing the application in different ways. Securing a URI and the application are separate independent settings. Would this setting then apply for any other URIs that need to be secured in the future?

MichaelErmer:
@Ashex1 I think this should be use for any functionality that should only be used by the map owner.

MichaelErmer:
Rebased and tested, https://i.imgur.com/rLMzTOT.png

PR #1020, opened 2016-07-21
FEATURE Use geo api

DanielSiepmann:
This pull request includes a New feature
The following changes were made

* Track user location
* Update map position and marker
* Send request to scan surrounding area

PR #725, opened 2016-07-21
FEATURE: user switchable geolocation w/ map pan fixed.

megahambone:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

The following Features were added:
Use of Geolocation API in most browsers. (Chrome on PC/Chrome on Android won't work unless your page is hosted HTTPS)
As per comments below, this has been Tested in EDGE, Firefox, iOS Chrome, iOS Safari and Dolphin on Android. 
As per comments below, IE11 does not work (one of the PRs merged in used string literals which is too new for IE) and Chrome will not work because it won't allow geolocation using http (unless you're connecting via localhost)
Geolocation can be turned on and off via toggle.
Geolocation will not rehome the map over and over again unless you've moved 20 meters (60 feet ish)

PR #1255, opened 2016-07-22
Add button to set location of map to current location

billyvg:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Added a button in the header to fetch user's current location using browser's geolocation API and set as next location to crawl


billyvg:
Modified #1020 a bit, but added explicit button to set a new location

Rohmin:
Will this also stop the current scan and start a new scan with the given location?

ellisio:
@Rohmin looks like it does. He calls `/next_loc`.

Rohmin:
Just tested, doesn't scan after moving to new location

billyvg:
It scans at the next location after the current scan ends.

talentedmrripley:
This appears to be a duplicate of #725.  That works without needing a button as well.

MonkeyLeeT:
Please send in a new PR with exactly the change you want, considering #725 .

Issue #1240, opened 2016-07-22
Heroku Scan Delay Error


SimonTellsAll:
## Expected Behavior
I have a scan delay of 5 seconds set after each full scan cycle. (-sd 5). After 5 seconds of completing the scan an error occurs.

## Current Behavior
The log shows an error in the code file.

```
2016-07-22T14:34:00.223916+00:00 app[web.1]: 2016-07-22 14:34:00,223 [     search] [   INFO] Waiting 5 seconds before beginning new scan.
2016-07-22T14:34:02.651843+00:00 app[web.1]: 2016-07-22 14:34:02,651 [  _internal] [   INFO] 10.81.154.203 - - [22/Jul/2016 14:34:02] "GET /raw_data?pokemon=true&pokestops=true&gyms=true HTTP/1.1" 200 -
2016-07-22T14:34:02.618090+00:00 heroku[router]: at=info method=GET path="/raw_data?pokemon=true&pokestops=true&gyms=true" host=xxx.herokuapp.com request_id=263b5070-0a17-42a4-b1b0-2e0bda7f81bb fwd="81.99.7.116" dyno=web.1 connect=1ms service=11ms status=200 bytes=17576
2016-07-22T14:34:05.227307+00:00 app[web.1]: Exception in thread search_thread:
2016-07-22T14:34:05.227320+00:00 app[web.1]: Traceback (most recent call last):
2016-07-22T14:34:05.227331+00:00 app[web.1]:   File "/app/pogom/search.py", line 63, in search
2016-07-22T14:34:07.651882+00:00 app[web.1]: 2016-07-22 14:34:07,651 [  _internal] [   INFO] 10.81.154.203 - - [22/Jul/2016 14:34:07] "GET /raw_data?pokemon=true&pokestops=true&gyms=true HTTP/1.1" 200 -
2016-07-22T14:34:07.617978+00:00 heroku[router]: at=info method=GET path="/raw_data?pokemon=true&pokestops=true&gyms=true" host=xxx.herokuapp.com request_id=add244e9-28cf-4314-9e52-5b3ab20442c7 fwd="81.99.7.116" dyno=web.1 connect=1ms service=11ms status=200 bytes=17157

```

## Possible Solution
Correct the part of the code that is causing this bug.

## Steps to Reproduce (for bugs)

1. Deploy on Heroku
2. Set extra args as -sd 5
3. Let full scan complete
4. Will fail

## Context
[...]

Issue #1205, opened 2016-07-22
App crashes on Heroku after checking for logintimer (remaining login time)


DanielVolz:
<img width="745" alt="screenshot 2016-07-22 14 29 25" src="https://cloud.githubusercontent.com/assets/3275994/17057224/fde226b2-5018-11e6-8271-dd6f262ca287.png">

Locally it works flawlessly. I'm on the newest dev commit.
And I used google auth.

birdwell:
I think there have been other issues like this open. I ran into the same problem.

DanielVolz:
Yeah I noticed this to:
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1141
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/965
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1112
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1071

birdwell:
Yeah, I am getting happening over and over again once it completes.
![screen shot 2016-07-22 at 8 26 51 am](https://cloud.githubusercontent.com/assets/3323856/17058463/3c64947a-4fe6-11e6-87ef-7a2d74ebb1f4.png)

MonkeyLeeT:
Closing this issue as it is outdated. If this issue still persists on the `develop` branch, please reopen a new issue and update the information therein to reflect the new changes.

Issue #1141, opened 2016-07-22
auth tickets are a mess! api._auth_provider._ticket_expire is a string


kaysond:
Causes this error to occur, but not consistently. I imagine the api is sometimes setting something unexpected in the _ticket_expire variable

2016-07-22T06:22:22.627426+00:00 app[web.1]: Exception in thread search_thread:
2016-07-22T06:22:22.627456+00:00 app[web.1]: Traceback (most recent call last):
2016-07-22T06:22:22.627459+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/threading.py", line 801, in __bootstrap_inner
2016-07-22T06:22:22.627479+00:00 app[web.1]:     self.run()
2016-07-22T06:22:22.627426+00:00 app[web.1]: Exception in thread search_thread:
2016-07-22T06:22:22.627456+00:00 app[web.1]: Traceback (most recent call last):
2016-07-22T06:22:22.627482+00:00 app[web.1]:     search(args)
2016-07-22T06:22:22.627483+00:00 app[web.1]:   File "/app/pogom/search.py", line 63, in search
2016-07-22T06:22:22.627488+00:00 app[web.1]: 
2016-07-22T06:22:22.627426+00:00 app[web.1]: Exception in thread search_thread:
2016-07-22T06:22:22.627482+00:00 app[web.1]:     search(args)
2016-07-22T06:22:22.627484+00:00 app[web.1]:     remaining_time = api._auth_provider._ticket_expire/1000 - time.time()
2016-07-22T06:22:22.627488+00:00 app[web.1]: 
2016-07-22T06:22:22.627484+00:00 app[web.1]: TypeError: unsupported operand type(s) for /: 'str' and 'int'

After some debugging - the first rpc POST request returns only two auth_ticket values. The first is a long alphanumeric/punctuation. The second is an integer.

The search thread then makes a POST request to /plfe/rpc instead of the given endpoint from before, which generates a third, different auth_ticket (another alphanumeric/punctuation), though this doesn't seem to overwrite the alphanumeric.

The first search starts before auth_tickets get populated, so it runs fine. On the second loop, it crashes because of this error.

kaysond:
Looks like _ticket_expire is set to: "y3FPIf7/IHXaIDLEQJLkVg==" in my case.

mhelmetag:
Yeah I see this as well on my heroku instance. Can't seem to get passed one search loop.

[...]

Issue #1071, opened 2016-07-21
Pokemon not showing on heroku


famendola1:
I get this error in the heroku logs:

[  ERROR] Scan step failed. Response dictionary key error.

I don't know where the dictionary is being referenced

Also, at the end I get this: 

TypeError: unsupported operand type(s) for /: 'str' and 'int'

from the call:

remaining_time = api._auth_provider._ticket_expire/1000 - time.time()

whippythought:
Duplicate of #965 which was closed with no resolution.

birdwell:
I am getting the same error. The api._auth_provider._ticket_expire isn't an integer and you can't convert it to an int either. Might be tied to a certain version of python.

macgngsta:
Response dictionary key error was related to the pokemon go servers being down. iirc

birdwell:
The code should still fail gracefully and not stop everything.

birdwell:
Right now, I'm not getting it to go through another scan after the first one.

c133:
Getting the same unsupported operand type error on heroku instance. Is there a fix for this yet?

medghaim:
Closing this issue as it is outdated. If this issue still persists on the `develop` branch, please reopen a new issue and update the information therein to reflect the new changes.

Please direct support questions to: https://discordapp.com/invite/Kzp4yky

PR #1272, opened 2016-07-22
Fixed auth ticket

Frugghi:
This pull request includes a

- [X] Bug fix
- [ ] New feature
- [ ] Translation

- Fixed the order of params when calling set_ticket

Running locally the order of params is always (?) correct but when running on Heroku most of the times the order is wrong. This PR should fix the issue.
I'm not a Python guy so I don't know if there is a better way to do what I did.

#965
#1141
#1071
#1205 
#1240

### DIFF
--- a/pogom/pgoapi/pgoapi.py
+++ b/pogom/pgoapi/pgoapi.py
@@ -168,7 +168,8 @@ class PGoApi:
             return False
         
         if 'auth_ticket' in response:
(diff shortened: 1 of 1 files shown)


kaysond:
Thank you!

mhelmetag:
@Frugghi , nice! Worked for me. Was able to pass the second search loop with no problem.

kaysond:
Works for me too.

PR #1279, opened 2016-07-22
Added option to load settings from config.ini

JZ6:
This pull request includes a

- [ ] Bug fix
- [x ] New feature
- [ ] Translation

- add new -se command
- loads settings from config.ini
-

### DIFF
--- a/config.ini
+++ b/config.ini
@@ -1,0 +1,8 @@
+[Required]
+username: 
+password: 
+steps: 12
+location: 
+
+[Others]
+#run with console commands
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -8,6 +8,7 @@ import re
 import uuid
 import os
 import json
+import ConfigParser
 from datetime import datetime, timedelta
 
 from . import config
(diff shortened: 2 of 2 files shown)


config.ini
```diff
@@ -0,0 +1,8 @@
+[Required]
```
Ashex1: Using sections like Required and Others really limits how extensible this will be. Would make more sense to break it into sections that indicate their purpose and add comments indicated which sections are required.

```ini
#Provide Niantic Credentials here
[Authentication]
type: google # Sane values are google or ptc
user:
password:

# Settings that configure search
[Search]
steps: 
location:
auto_refresh:
```

pogom/utils.py
```diff
@@ -1,118 +1,126 @@
...
+from exceptions import APIKeyException
+
+
+def parse_unicode(bytestring):
+    decoded_string = bytestring.decode(sys.getfilesystemencoding())
+    return decoded_string
+
+def parse_config(args):
+    Config = ConfigParser.ConfigParser()
+    Config.read("config.ini")
```
Ashex1: We should be able to specify the location for the config file.

pogom/utils.py
```diff
@@ -1,118 +1,126 @@
...
+            config['LOCALES_DIR'],
+            'pokemon.{}.json'.format(config['LOCALE']))
+
+        with open(file_path, 'r') as f:
+            get_pokemon_name.names = json.loads(f.read())
+
+    return get_pokemon_name.names[str(pokemon_id)]
+
+def load_credentials(filepath):
+    with open(filepath+os.path.sep+'credentials.json') as file:
```
Ashex1: Doesn't this replace the function of credentials.json?

config.ini
JZ6: you don't get to chose whether to run each section, all data in the file will overwrite console commands when you use -se

pogom/utils.py
JZ6: i lladd an os path later

pogom/utils.py
JZ6: nope keeping credentials.json to avoid confusion

PR #1280, opened 2016-07-22
Faster pokemon search

medghaim:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Ignores pokestops and gyms on first iteration to serve pokemon to the user faster.

Related Issues: a whooole bunch.

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -75,7 +75,7 @@ class Gym(BaseModel):
     last_modified = DateTimeField()
 
 
-def parse_map(map_dict):
+def parse_map(map_dict, iteration_num):
     pokemons = {}
     pokestops = {}
     gyms = {}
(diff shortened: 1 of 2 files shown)


AHAAAAAAA:
+1, good job.

pogom/models.py
```diff
@@ -96,38 +96,39 @@ def parse_map(map_dict):
...
-                    if 'lure_info' in f:
-                        lure_expiration = datetime.utcfromtimestamp(
-                            f['lure_info']['lure_expires_timestamp_ms'] / 1000.0)
-                        active_pokemon_id = f['lure_info']['active_pokemon_id']
-                    else:
-                        lure_expiration, active_pokemon_id = None, None
-
-                    pokestops[f['id']] = {
-                        'pokestop_id': f['id'],
+        if iteration_num > 0 or step > 50:
```
jangrewe: why the 50?

medghaim: That was discussed. If `-st` is too high, the user will end up going on too long without any pokestops/gyms being displayed. This way it's the best of both worlds - they get a quick search, but don't go too long without not seeing stops/gyms.

What does everyone think about the magic 50? should it exist? be higher? lower?

reyqn: Seems arbitrary. So that you don't have to wait too much to get info on gyms or pokestops if you used a big -st.
>=49 (7^2) would probably more logical. And >= 37 or 61 once hexsearch is implemented

jangrewe: Okay, so it's just a "at some point we eventually gotta do this" number, right? On the other hand, if you ran it at least once before, all the gyms and stops should be in the DB already, so you could also do it either after the first iteration or right away...

Maybe also do it only every n-th iteration, as they don't change as often as the Pokemon?

Lutcikaur:
testing shows that this saves .002 seconds per search (for the first 50 searches, so a savings of  0.1s after 50+ seconds). Considering searches force a sleep of 1 second, and a slight time delay on overhead and the api call I don't think this is worth the time savings.

PR #1285, opened 2016-07-22
Fix: Localization should change Pokedex Link

blankse:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Pokedex Link with right locale

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -144,16 +144,28 @@ function initSidebar() {
 
 var pad = function (number) { return number <= 99 ? ("0" + number).slice(-2) : number; }
 
+function pokemonLang(locale) {
+    switch(locale.slice(-2)) {
+        case 'br':
+        case 'de':
+        case 'fr':
+        case 'ru':
+            return locale;
+        default:
+            return 'us';
+    }
+}
 
 function pokemonLabel(name, disappear_time, id, latitude, longitude) {
     disappear_date = new Date(disappear_time)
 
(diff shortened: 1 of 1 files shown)


MonkeyLeeT:
If you take a careful look at other language's podex, you'll see that they're poorly supported for now. 

Also {{lang}} is dominating the language setting, and it's currently mapping to locale file names, which is again different from expected "us" or "jp".

So this PR is not really working as you would expect.

blankse:
You are right. I add a switch statement before when I came to my pc ;)

MonkeyLeeT:
So would you mind close this one and reimplement another version later after our first release (soon)?

Default en pokedex should be enough for first release.

blankse:
Why close? I can rebase if needed.

blankse:
@MonkeyLeeT I changed it

MonkeyLeeT:
@blankse than a simple "-locale ja" will break it. I don't really think it's necessary to add this feature now cuz basically you need to manually map niantic's locale (us, jp) names to our locale folder's filenames (en, ja).

blankse:
Why should "-locale ja" break it? The pokemonLang function returns "us"

MonkeyLeeT:
But technically you should redirect them to Japanese version. We should not have incomplete codes like the switch clauses you have there.

JZ6:
too little benefit, too much room for error, sorry.

PR #1006, opened 2016-07-21
Fix dynamic displaying of pokestops based on their lure module status

invisiblek:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

-Set up a label for poke stops
-Dynamically update the pokestop marker and label based on lure_expiration vs current time and previous reading

If this is related to an existing ticket, include a link to it as well.

PR #1287, opened 2016-07-22
Add accurate Pokestop labels

andrewda:
This pull request includes a

- [X] Bug fix
- [X] New feature
- [ ] Translation

- Show expiration time for Lured Pokéstops
- Show directions from current location to Pokéstop
- Show Pokémon lured by that Pokéstop

Similar to #1006, but instead of using `lure_expiration`, which doesn't seem to be working properly, it uses `current_time - last_lured` to get the exact time until it expires.

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -12,11 +12,14 @@ document.addEventListener("DOMContentLoaded", function () {
 var $selectExclude = $("#exclude-pokemon");
 var $selectNotify = $("#notify-pokemon");
 
+var idToPokemon = {};
+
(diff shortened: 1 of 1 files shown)


JellyBellyFred:
:+1: looks good!

talentedmrripley:
Does this also show the name or description of the stop?  That would be useful info to display.

andrewda:
@talentedmrripley I haven't seen that in any server responses, but I'll look into a way for getting those. You're right, they would indeed be handy to have.

JZ6:
+1

AHAAAAAAA:
Great work! merging this

PR #1289, opened 2016-07-22
python 2.7 batch

rasbt:
This pull request includes a

- [ ] Bug fix
- [ ] New feature
- [ ] Translation
- [x] Documentation update

- Added a Python 2.7 batch to the README.md file to indicate that only Python 2.7 is supported at the moment

(PS: If there's interest, I can look into making it both Python 2.7.x and Python 3.x compatible; let me know)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 # PokemonGo Map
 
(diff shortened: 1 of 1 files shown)


JZ6:
approved

AHAAAAAAA:
@rasbt big fan here! thank you for this! We're getting a new release out soon, so I'd hold off on the compatibility!

rasbt:
> @rasbt big fan here! thank you for this! We're getting a new release out soon, so I'd hold off on the compatibility!

Wow, thanks! And yeah, waiting for the next major release before looking into this further is a good idea, I agree :) I am just back from SciPy 2016, where we had extensive Py27.x vs. Py 3.x discussions (over several beers :P), so I just thought bringing it up may not be a bad idea ;)

PR #1299, opened 2016-07-22
added pokemon icons to select

richipargo:
This pull request includes a
- [x] New feature

- added icons to select options

selection_001


ximex:
if think this is getting to large. maybe only the icons and the names on hover?

MonkeyLeeT:
You will need to rework your PR after today's change to use collection of sprites in one picture.

JZ6:
might be useful in implmenting checkboxes in the future

wudifeixue:
I love this idea! Helps so much on filtering the random names that I can never remember.

abdulsettarova:
gived me  job maps plzzzzz

PR #1316, opened 2016-07-22
Added option to load settings from config.ini

JZ6:
This pull request includes a

- [ ] Bug fix
- [x ] New feature
- [ ] Translation

- Added parser load from config.ini.
-moved credentials to /config
-


config/config.ini
```diff
@@ -0,0 +1,19 @@
+#Provide Pokemon Go Credentials here. This section must be filled out!
+
+[Athentication]
```
Ashex1: Should be `Authentication`, would also be good to change `Auth_Service` to just `Service` as Auth is a bit redundant.

JZ6: on it

npeep:
Is there a way to tell the runserver.py to use this config?  I get the following after filling out the Config.ini and running the runserver.py with no arguments:

runserver.py: error: argument -u/--username is required

npeep:
I see the -se option that was added now, thanks!!

hhu94:
Why was this merged? It doesn't work properly. Some of the arguments are still required even when using the -se argument.

PR #1322, opened 2016-07-22
show unhidden pokemons notifications in browser

SJ119:
This pull request includes a

- [Browser notifications of unhidden Pokemon ] New feature

- Compare pokemon found against excluded Pokemon name list (found in sidebar), if Pokemon found does not exist in excluded list, send out notification to browser
- Toggle for this located in sidebar

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -120,6 +120,7 @@ function initSidebar() {
     $('#gyms-switch').prop('checked', localStorage.showGyms === 'true');
     $('#pokemon-switch').prop('checked', localStorage.showPokemon === 'true');
     $('#pokestops-switch').prop('checked', localStorage.showPokestops === 'true');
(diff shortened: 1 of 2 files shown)


andrewda:
If this options is enabled, you will be spammed to hell with Pokémon notifications when the map first loads. After that, this is a great addition!

SJ119:
Yeah, good point, I already had a list of excluded pokemon, and found that adding pokemon to the notification list one by one was kind of tedious. I do agree that the notifications at the beginning are hella annoying tho

PiaFraus:
And this is why in my solution - I accumulated new pokemons first and only then issued notifications for all new pokemons and not one by one.

AHAAAAAAA:
This has the function of NotifyAllExceptExcluded which is the same as #1415. Like I commented there, I think the proper approach would be to include a 'All' selector in Notify pokemon, that takes into account excluded pokemon.

PR #1336, opened 2016-07-22
Update map.js for optional URL passed locations

wbechard:
This pull request includes a

- [1 ] New feature

- Enabled optional URL parameters for geo positioning.
- Allows Location Aware apps to load the URL dynamically
- Does not change the default behaviour
- Uses lat and lang querystring parameters (?lat=42.3152924&long=-83.0367337)

I've build a location aware Android app that can pass the location into the map.js to allow for live map location detection (without HTTPS on the server end).  This change doesn't change the default functionality, merely adds optional URL lat & long params.

Added function getParameterByName(), and a few lines in initMap().

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -63,6 +63,18 @@ var selectedStyle = 'light';
 
 function initMap() {
 
+    // Add support for URL passed locations.
+    var latitude = getParameterByName('lat');
+    var longitude = getParameterByName('long');
+
+    if (latitude) {
+	var pos = {
+	    lat: latitude,
+	    lng: longitude
+	};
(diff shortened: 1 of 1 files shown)


vinsilsim:
I tried to add this feature but even when the step parameters nothing happens

check:

http://pokemap.com.br/?lat=47.60754285227&long=-122.34254118148161

wbechard:
Yes, some changes just broke some of it.  You probably see the position marker in the correct location, but the map isn't centering.  I'll resolve it tonight and commit.

vinsilsim:
@wbechard thank you so much!

Issue #1105, opened 2016-07-22
Possible gaps in search pattern


hazanjon:
I have raised a PR #1102 which shows visually on the map a 100m circle around the coords that have been loaded from the API. This seems to show that the current search step of 0.0025 Is actually too large for Latitude (And partially Longitude). This can be seen in the image below:

![scanned locations](https://cloud.githubusercontent.com/assets/1560723/17044439/fa01f89e-4fb7-11e6-8bc4-3e752c299333.png)

The current search is leaving large horizontal strips of the map unsearched.

## Possible Solution

Changing the scan steps to 0.002 for Longitude and 0.0012 for Latitude provides a much more complete coverage:

![tight locations](https://cloud.githubusercontent.com/assets/1560723/17044527/a7e1424e-4fb8-11e6-9746-25f09bf5b410.png)

This does lead to a duplication of search area, to optimise this I would think running the search as two passes would be beneficial. Each pass taking alternating locations.

reyqn:
#776 will fix it. But yeah you can do that at the moment. 0.00126 should be better

FrostTheFox:
We're working on all of this, please standby.

arianizadi:
Hey i am testing to see if your solution is working. So far it is! @hazanjon

ericpadilla:
where could you change the steps @haranjon? I would love to try it out

arianizadi:
hey @ericpadilla i made a video how to https://www.youtube.com/watch?v=y5qvVBFW-B0 its not the best video but its something.

hazanjon:
@ericpadilla You can change the values in search.py line 38

`yield (x * 0.0012 + initial_location[0], y * 0.0020 + initial_location[1], 0)`

hazanjon:
Actually it may be better to change the pattern entirely to be a Hex pattern

![pattern](http://i.stack.imgur.com/XI4um.png)

hazanjon:
@reyqn Sorry missed your link to #776. You right that will fix this completely :)

talentedmrripley:
Can we get #776 merged soon?

hazanjon:
Closing as this is now fixed

PR #1102, opened 2016-07-22
Add Scanned Area markers onto the map

hazanjon:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

This adds a new option on the map that shows what areas have been scanned (100m Radius around the requested LatLng sent to the API)

The circles are colored between Green and Red depending on how recent the area was updated. Green for 0 mins and Red for >= 15 mins

scanned locations

I will raise a separate issue but this has shown up a possible problem with the search stepping of 0.0025 in search.py as can be seen in the image above.

Edit - A better image now we are using the Hex grid:

big hex

Issue #522, opened 2016-07-20
[Feature request]"Fog of war"


Lunat1q:
I don't know  is it possible to do on google map, but here is it:
Add feature to see: 
1. What zone is currently in detection radius
2. Which "squares" was checked, and which wasn't
Imo it should looks like this:
http://i.imgur.com/0HTZTt3.png
where noColor - zone out of radius
blurred - not checked zone yet.

ofc this is just a picture, but idea is just showing those zones, no matter how :)

KevakaBuel:
I would prefer a less drastical change, something like the range indicator you have inside the actual game. Feature itself would be pretty nice. #494 has a similar idea

sykhro:
The blurring is a good design solution, we will make a few tests

frdmn:
I totally agree with @KevakaBuel, this looks too prominent - Why not a simple translucent radius like this: 

![bildschirmfoto 2016-07-20 um 14 21 38](https://cloud.githubusercontent.com/assets/944459/16986127/4957303c-4e85-11e6-9b67-b57485079da9.png)

(via http://obeattie.github.io/gmaps-radius/?lat=51.500358&lng=-0.125506&z=10&u=mi&r=5)

Lunat1q:
i didn't say it should be exactly as i draw. There should be 2 options: displaying detection zone, and displaying already scanned area INSIDE that zone.

KevakaBuel:
I guess it cant be a circle because we scan in squares but i like the style of it =)

DerScreever:
like @KevakaBuel  say, we scan in squares so a radius would confuse the user :D

frdmn:
@DerScreever @KevakaBuel Oh, I was not aware of that. In this case we definitely should show a rectangle shape as well. 👍

MikeDX:
We scan in squares but the radius of the scan is whatever niantic set, (distance from player) which I'm fairly sure is not a square box surrounding the player

HinoReo:
@DerScreever I think that "we scan in squares" meant that the figure made of circles is square(ish), Though with implementation of hexagonal search pattern it will be more round, rendering a circle more appropriate, imo.

PR #1338, opened 2016-07-22
Change search pattern to use a hex grid

hazanjon:
This pull request includes a

- [x] Bug fix
- [x] New feature
- [ ] Translation

This changes the search pattern to use a hexagonal grid. 

hex grid

This fixes Issue #1105

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -15,6 +15,19 @@ log = logging.getLogger(__name__)
 TIMESTAMP = '\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000'
 api = PGoApi()
 
+#Constants for Hex Grid
+#Gap between vertical and horzonal "rows"
+lat_gap_meters = 150
+lng_gap_meters = 86.6
+
+#111111m is approx 1 degree Lat, which is close enough for this
+meters_per_degree = 111111
+lat_gap_degrees = float(lat_gap_meters) / meters_per_degree
+
+def calculate_lng_degrees(lat):
+    return float(lng_gap_meters) / (meters_per_degree * math.cos(math.radians(lat)))
+
+
 def send_map_request(api, position):
     try:
         api.set_position(*position)
@@ -29,15 +42,46 @@ def send_map_request(api, position):
 
 
 def generate_location_steps(initial_location, num_steps):
-    pos, x, y, dx, dy = 1, 0, 0, 0, -1
 
-    while -num_steps / 2 < x <= num_steps / 2 and -num_steps / 2 < y <= num_steps / 2:
-        yield (x * 0.00125 + initial_location[0], y * 0.00175 + initial_location[1], 0)
+    ring = 1 #Which ring are we on, 0 = center
+    lat_location = initial_location[0]
+    lng_location = initial_location[1]
 
-        if x == y or (x < 0 and x == -y) or (x > 0 and x == 1 - y):
-            dx, dy = -dy, dx
+    yield (initial_location[0],initial_location[1], 0) #Middle circle
 
-        x, y = x + dx, y + dy
+    while ring < num_steps:
+        #Move the location diagonally to top left spot, then start the circle which will end up back here for the next ring 
+        #Move Lat north first
+        lat_location += lat_gap_degrees
+        lng_location -= calculate_lng_degrees(lat_location)
+
+        for direction in range(6):
+            for i in range(ring):
+                if direction == 0: #Right
+                    lng_location += calculate_lng_degrees(lat_location) * 2
+
+                if direction == 1: #Right Down
+                    lat_location -= lat_gap_degrees
+                    lng_location += calculate_lng_degrees(lat_location)
+
+                if direction == 2: #Left Down
+                    lat_location -= lat_gap_degrees
+                    lng_location -= calculate_lng_degrees(lat_location)
+
+                if direction == 3: #Left
+                    lng_location -= calculate_lng_degrees(lat_location) * 2
+
+                if direction == 4: #Left Up
+                    lat_location += lat_gap_degrees
+                    lng_location -= calculate_lng_degrees(lat_location)
+
+                if direction == 5: #Right Up
+                    lat_location += lat_gap_degrees
+                    lng_location += calculate_lng_degrees(lat_location)
+
+                yield (lat_location, lng_location, 0) #Middle circle
+
+        ring += 1
 
 
 def login(args, position):
@@ -54,6 +98,7 @@ def login(args, position):
 
 def search(args, i):
     num_steps = args.step_limit
+    total_steps = (3 * (num_steps**2)) - (3 * num_steps) + 1
     position = (config['ORIGINAL_LATITUDE'], config['ORIGINAL_LONGITUDE'], 0)
 
     if api._auth_provider and api._auth_provider._ticket_expire:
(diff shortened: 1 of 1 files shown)


rkowalk:
I was just about to start writing something for this!  Have you thought about extending this to using a polygon grid?

Here is a screenshot of what I was thinking about implementing - starting a series of threaded searches dependent on the geo area provided.

screen shot 2016-07-22 at 2 30 06 pm

rkowalk:
Wait... Misread this.  Do you know if anyone has built something like the above?  A server starting script that can accept coordinates?

hazanjon:
@rkowalk As far as I know no-one has created a PR to fill an area. At the moment all the searching is done outwards for a set starting point.

You could certainly adapt the script to fill inside a area. However it would probably take a long time to search the area in your screenshot. By the time it got halfway through the first part of the map would be out of date. To do what you need you will need to be coordinating multiple workers.

rkowalk:
@hazanjon I have started writing a script that would allow you to have multiple accounts(to avoid throttling), and coordinate the multiple workers(as you stated above).

All it requires is passing in a set of parameters and accounts and it does the rest.

I'm hoping with the update to MySQL it will allow enough synchronous write logic to avoid DB locks.

hazanjon:
You can use my PR #1102 to visualise the new search

AHAAAAAAA:
Another developer was actually just working on this. It's the most efficient way to scan that we found. I looked at the code, and your algorithm is missing bearings. This would work around the equator, but will still have gaps closer to the poles. You need to incorporate bearings, it's not ultra pretty math but look at this [StackOverflow](https://stackoverflow.com/questions/7222382/get-lat-long-given-current-point-distance-and-bearing)

hazanjon:
@AHAAAAAAA I've just tested this using both the most northern and southern cities in the world and also the south pole, But theres no Pokemon showing there, and it worked as expected

Norilsk, Siberia

siberia

Puerto Williams, Chile

puerto williams chile

Rhynocerous:
Equirectangular approximation is accurate within around 3 meters to the poles, it's what I was using in #776 

Without correcting, lat dilates dramatically.  

I closed my PR because it was intended to fix the algorithm missing pokemon, and while it certainly improved it, it still was missing some.

reyqn:
If I'm not wrong, it's less than 0.88% inaccurate. Though I guess using 111694 would solve it everywhere. (A little more overlapping around the equator)
And  @Rhynocerous  I was getting all of them with your PR.

hazanjon:
@Rhynocerous Yeah I saw your PR :) It looked good, only created this one up as you closed yours.

We really should get a hex pattern of some sort merged in as it's far more efficient than the current square pattern. But it's less urgent now the square pattern is covering 100%+

AHAAAAAAA:
You're correct, I tested this and seem to be wrong about my assumptions. Merging

jangrewe:
So what's the step size recommendation if i used 10 until earlier today, when the cell size was ~halved, and used 20 with those smaller cells?

HinoReo:
@jangrewe I believe I saw 31 steps to be of exactly the same coverage.

jangrewe:
@HinoReo thanks, i'll leave it at 20 then for now and will ramp it up later - i just didn't want to pick up too many markers, in case those circles were a lot bigger again.

hazanjon:
The step arg to area covered actually increases. You will get more coverage per circle (api call). But also more circles per step increase.

So you will probably want to drop the step count down per worker

Step Limit | Circles Square Pattern | Circles Hex Pattern
------------ | ------------- | -------------
1 | 1 | 1
2 | 4 | 7
3 | 9 | 19
4 | 16 | 37
5 | 25 | 67
6 | 36 | 103
7 | 49 | 145

hazanjon:
A circle with a 100m radius covers 31415 meters squared

The square search pattern with would cover around 20000 meters squared (63%)

The hex search pattern increases the cover to 26040 meters squared (82%)

jangrewe:
holy sheeez, i just saw that i wanted to do >1400 steps :grin: 
i've set it to 6 now, which will do 91 steps now (at least that's what the log says), and that already says `Completed 205.56% of scan.` at 75 steps ... that may need a bit of fixing ;-)

but huge thanks for that improvement, the hitrate looks very nice!

Faustvii:
I noticed the percentage is messing up - http://i.imgur.com/5d1VgrH.png

HinoReo:
It looks like we are really need #522 to be implemented :)

hazanjon:
Yeah the percentage issue has just been raised in #1370

Looks like I missed updating the log message.

`log.info('Completed {:5.2f}% of scan.'.format(float(step) / num_steps**2*100))`
needs to become
`log.info('Completed {:5.2f}% of scan.'.format(float(step) / total_steps*100))`

Thinking it might be better to get @AHAAAAAAA to change it directly. Or i can make a new PR.

reyqn:
@HinoReo #1102 does it pretty well

jangrewe:
whoa, that would be awesome!

HinoReo:
@reyqn Oh, I missed this one! Thanks. Waiting then :)

hazanjon:
@reyqn @HinoReo It depends on if you want to see the exact areas that have been searched (#1102) or an overall area which would just be a circle of radius (Step Limit * 173.2) * 86.6 meters

jangrewe:
my #1 wish would be seeing how recent the data for a sector is, and #2 then the general area - even though you could figure that out by looking at the sectors from #1 ;-)

hazanjon:
@jangrewe #1102 is what your after for (1) it will highlight the scanned areas in green which then fades to red over 15 mins before dissapearing

jangrewe:
yup, so let's hope somebody deems it worthy to be merged ;-)

HinoReo:
@hazanjon I hope to see theese:
1. Outline for general area that is being searched like in #522 
2. All the circles like you did in #1102 
3. A switch for 1, 2, "Both" and "None" :)

rkowalk:
@hazanjon I'm working on getting my script to disperse threads through a geofenced area.  I can quite get the math down for the size of each hexagon.

What is the diameter of each circle?  And how much overlap?

reyqn:
The diameter of a circle is 200 m. Overlap is 100*(2-sqrt(3)) m.
Hence the width: 2(nb_steps-1)*100sqrt(3)+200 m
The height: 300(nb_steps-1)+200 m for uneven nb_steps
                     300(nb_steps-1)+100 m for even nb_steps

Keep in mind that sqrt(3) has been approximated to 1.732

If you want me to also calculate the real scanned area just say so.

rkowalk:
@reyqn In my attempts to write a geo-fenced script, I am stumped by finding the location of the next script starting point.
Currently I am finding my lat long as such(top left for example):
new_lat = lat_gap_degrees*(steps-1)+lat
new_long = lng-(2*lng_degrees)*steps*1.4

I'm guessing my approximations are just slightly off, but they yield the following result.

screen shot 2016-07-22 at 9 37 02 pm

The goal is to allow a user to pass in a draw path geo, and then generate as many headless scripts as is needed to monitor that area.

Do you have any advice for how to get these formulas more correct?

reyqn:
Your new_lat formula seems good.
new_long = lng-lng_degrees(2steps-3) maybe?

Krys1s:
@rkowalk If you get that working you would be my hero. Was planning to wake up and manually try to fix my script to not overlap so horribly.

hazanjon:
@rkowalk The movement you need to make depends on which side you are moving out from

The "vertical" distance between each row is 150m - lat_gap_degrees variable
The horizontal distance is a bit more fun as one row is offset. but between the center of a circle and one on the row above or below its about~ 86.6m * cos(latitude) - calculate_lng_degrees(lat) function

For diagonal moves (Top/Bot Left/Right) you are changing the Latitude by (Steps * lat_gap_degrees) and Longitude by ((1.5 * Steps) - 0.5) * calculate_lng_degrees(lat)
For Vertical moves its simply (Steps * lat_gap_degrees) * 2

I havent tested the above math so may be completely off, but seems right in my head :)

PR #1341, opened 2016-07-22
Update map.js for optional URL passed locations

wbechard:
This pull request includes a

[1 ] New feature

Enabled optional URL parameters for geo positioning.
Allows Location Aware apps to load the URL dynamically
Does not change the default behaviour
Uses lat and lang querystring parameters (?lat=42.3152924&long=-83.0367337)
I've build a location aware Android app that can pass the location into the map.js to allow for live map location detection (without HTTPS on the server end). This change doesn't change the default functionality, merely adds optional URL lat & long params.

Added function getParameterByName(), and a few lines in initMap().


wbechard:
Updated URL based geo code to live in map.html template rather than the maps.js where all the action is.

MonkeyLeeT:
Well it should still go to map.js isn't it? Others are all following such coding standard.

PR #1356, opened 2016-07-22
Make the /mobile page more accesibility friendly

chuyskywalker:
For a friend with vision issues, this patch will make the mobile page more accessible:

 - pokemon names will have voiceover support
 - columns have headings for clarity
 - "time remaining" is properly pronounce by siri

I also jazzed up the styles for us sighted folks a bit too.

Before:

After:

_I may have "done wrong" by the CSS setup for this though..._

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -74,7 +74,7 @@ class Pogom(Flask):
                 'name': pokemon['pokemon_name'],
                 'card_dir': direction,
                 'distance': int(origin_point.get_distance(pokemon_point).radians * 6366468.241830914),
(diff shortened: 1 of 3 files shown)


chuyskywalker:
Well, those images came out a bit larger than anticipated...

MonkeyLeeT:
@1283 seems to work well. Wanna close this one?

chuyskywalker:
I'd like to reopen this one as it addressed several issues that the other did not. I will rebase on top of the current branch, however, since the styling difference isn't that important. But the other did not include almost any of the work I did which actually makes the page accessibility friendly.

chuyskywalker:
Re-opened as #1696

PR #1372, opened 2016-07-22
Add flags to disable pokemon, pokestop, and/or gym parsing

Daegalus:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Added new arg and config flags
- Added boolean if checks before the parsing sections and the upsert sections to disable them when necessary.
- Changed the else to an elif to properly handle skipping  the first if.

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -10,6 +10,7 @@ from datetime import datetime
 from datetime import timedelta
 from base64 import b64encode
 
+from . import config
 from .utils import get_pokemon_name, load_credentials, get_args
 from .transform import transform_from_wgs_to_gcj
 from .customLog import printPokemon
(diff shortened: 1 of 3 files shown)


niayh:
These parameters (well the opposite really) were already intentionally removed a couple days ago with 51f651228c00a96b86f5c38d1a2d53b32e5d9862

Daegalus:
Well that's stupid. There is a difference between not querying the DB for things to show up on he map and not spending the processing cycles to parse he data and insert it into the DB. It almost doubles he speed at which the search scans and updates Pokémon. And my usecase also has 0 need for pokestops and gyms on the map. Also noticed performce and limits are hit which shows less Pokémon on the map in areas with insane numbers of pokestops like San Francisco. We need these flags serverside for optimizing. Because the efficiency of the parse is already pretty high. I think O(n^3) if I remember the code off the top of my head right.

AHAAAAAAA:
Can you rebase, I agree that these types of optimizations could be useful and not everyone wants to go through GUI.

Daegalus:
All set, rebased onto current develop

PR #1377, opened 2016-07-22
update get_cellid to use s2cell coverage instead of linear path

nletourneau:
This pull request includes a

- [x] Bug fix
- [ ] New feature
- [ ] Translation

As requested by @AHAAAAAAA on twitter after my comment in #755 here are some small GIS improvements. I kept this changeset really small as it it seems there are other PRs to handle the overlapping circles, like #1338 

This updates the cell coverage to only include the 100m radius circle of scanning.

Old style cellid overlay:
old_map

New style cellid overlay (this PR)
map-anim

The strangeness with the random added ones in the third frame of the animation is because for whatever reason the server requires 21 cell ids. I have no idea why.

This needs to be tested in a bunch of different geographies before it gets merged. It is a green green (except for you know... no tests :) ) refactor though and should not change anything other than fixing missing pokemon in cases like the first map image.

### DIFF
--- a/pogom/pgoapi/utilities.py
+++ b/pogom/pgoapi/utilities.py
@@ -27,7 +27,7 @@ import struct
 import re
 
 from importlib import import_module
-from s2sphere import CellId, LatLng
+from s2sphere import LatLng, Cell, RegionCoverer, Cap
 from google.protobuf.internal import encoder
 from geopy.geocoders import GoogleV3
 
@@ -40,7 +40,7 @@ def f2h(float):
 
 def h2f(hex):
   return struct.unpack('<d', struct.pack('<Q', int(hex,16)))[0]
-  
+
 def to_camel_case(value):
   def camelcase():
     while True:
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
This code is beyond my understanding, but the math makes sense to me. It seems like the current develop is capturing more pokemon after a 10 step search, but that's a unthorough crude estimate. We've been getting less pokemon versus v1.0 with our scanning, even though our new search patterns are less redundant. Can you also take a look at pogom/search.py to see what's happening?

Can we also get more testing here, I'd like to get to the root cause of this scan issue soon!

Mandrakia:
I managed to send up to 61 cellIds to the server. After that it will start to truncate data.
The Fort detection radius is way larger than the 21 CellIds people use.

You should also implement the delta mechanism to alleviate the load on the niantic servers.

AHAAAAAAA:
I believe the 21 CellIds radius is the Pokemon detection radius which is the most important imo

Mandrakia:
Best way to test your navigation algorithm is to plot the cellI visited and the position used as origin with a 100m radius around it in real time to check if it doesn't overlap too much.
It will always overlap a little unfortunately but it's better to send some extra query if you just ask for the delta (extremely fast) rather than missing some precious pokemons !

Mandrakia:
Pokemon detection radius is actually way smaller than a Level 15 cell. Tried it :D

AHAAAAAAA:
@Mandrakia that explains a lot of behaviour! I will merge this change for now as it's an improvement on what currently exists, further PRs can further improve on it.

zeryl:
Running develop as my main codebase, and with this, I see MAYBE 10 pokemon in a 40 step radius.  reverting to the previous code, well over 100-150.  Something is still not right on this.

Issue #1192, opened 2016-07-22
Save settings for "Hide common Pokémon"


mrfacu87:
## Expected Behavior
Thank you for the hard work that has been put into this! I would like to suggest an improvement in the "Hide common Pokémon" feature. It would be great if the list of Pokemon added on the filter could be saved so you do not have to add them every time you start a new session. It would also be good if the filter would sync across devices.

## Current Behavior
"Hide common Pokémon" feature allows you to filter Pokemon from the map. Starting a new session on the same computer or different device clears the filter and forces you to re-add the filtered items again.

## Steps to Reproduce (for bugs)
1. Add Pokemons on "Hide common Pokémon"
2. Close browser
3. Open browser and check filter.

## Your Environment

* Version used:
* Environment name and version: Python 2.7
* Operating System and version (desktop or mobile):
Windows 10 
iPhone iOS 9.3.2

craigbaines:
Synced across devices doesnt seem like a good idea if i have understood what you mean. I plan to let my brother / friends access my map from my local machine, all of us i assume will want diff filters, this new way seems much better :)

M4CC:
This would be great!

chuyskywalker:
`localstorage` to the rescue!

chuyskywalker:
Wait, this is already in place. The values are saved in cookies, though, so maybe that's breaking something for you?

chuyskywalker:
@mrfacu87 Try on the latest develop branch now, if localstorage was working for pokestop/gym selection, the pokemon lists will work now

mrfacu87:
Thank you again for this great app. I just updated to the latest develop branch. The list that I had from the previous develop branch is no longer present now. I will continue to use this newer build and check again the next time I update. New build looks great!! I love the Scanned locations and function to enter the address. Thanks.

PR #1390, opened 2016-07-22
Use localstorage for exclude/notify pokemon lists

chuyskywalker:
In #1192 a user is having trouble keeping these lists with cookies, but the localstorage options do work. Since there's no real reason to have these as cookies, I've converted the lists over to localstorage instead.


static/map.js
```diff
@@ -22,19 +22,24 @@ $.getJSON("static/locales/pokemon." + document.documentElement.lang + ".json").d
         idToPokemon[key] = value;
     });
 
-    JSON.parse(readCookie("remember_select_exclude"));
```
chuyskywalker: I love how this was reading the cookie and parsing it...for fun!

andrewda: IDK, looks like it could be useful /s

AHAAAAAAA:
+1 works

PR #1395, opened 2016-07-22
Multithread search

wcastello:
This pull request includes a

- [x] New feature

Alright I'm just going to leave this here. I've tried this before but had some lock problems with sqlite. This time I used a semaphore (this is probably not helping!) around the parse_map() and at least in my tests ~~it's been running pretty smoothly~~ and much faster than the single threaded version:

Run it like: 

`
python runserver.py -u xxxx -p "yyyy" -l "Portland, OR" -st 25 -t 100 -sd 10
`

Took me 2 minutes to cover this area:
screen shot 2016-07-22 at 4 56 11 pm


shivampaw:
Are those Pokestops? If so, where do you live!!

And yeah, I'd like to see multi threading.

wcastello:
@shivampaw they're Pokestops yes, but that's not my location!

pickett2k:
really hope this gets merged if it speeds up search times

EdenGottlieb:
Multithreaded scans with the same user sounds like a source of bans. The game only Soft bans users who cheat so this needs to be checked if it can pass the cheating checks.

"Players caught cheating reported that they could still enter the game but little else. Players could not gather items or even do battle while they are banned from the game. If a banned player throws a Pokeball at a creature, it will leave the area instead."

wcastello:
@EdenGottlieb well, you're not supposed to run any of this with your user, right?

I've been running multithreaded scans since yesterday and so far I'm still waiting for a ban. Anyway, it's optional.

jangrewe:
that looks like a serious speed boost. and even if you're banned and can't play anymore, who cares! as long as the API still provides data ;-)

one thing that would be needed with such a large area: querying the data only for the coordinates within the viewport. otherwise you'd be transferring way too much ...

wcastello:
I'm also considering throwing all the data in memory first before flushing them into sqlite and dealing with db locks, it's really not much stuff. After 4000 steps there is only 2700 pokemons on the map, 886 pokestops and 2840 gyms, that's only 1.3mb on disk.

jangrewe:
or let's wait for the mysql backend PR, that would also help take some strain off of the file

wcastello:
This is after 6200 steps, covered almost the whole city:
screen shot 2016-07-22 at 5 27 36 pm

wcastello:
Oh, yea, mysql would help with this.

Plop:
@wcastello how much time did it take you to generate 6200 steps ?

wcastello:
@Plop, it's been running for around 5-8 minutes

Plop:
@wcastello Amazing :) Would really love to see it merged.

jangrewe:
damn...

i'd probably split that into 4 separate workers, depending of the number of steps needed for Berlin.
how CPU-heavy is it, compared to running single-threaded?

everix1992:
This would be great! I was just thinking about doing something similar myself to speed up search times for a map covering my smallish town. It may be useful to consider adding support for querying with multiple accounts if we're going to be moving to a multi-threaded model

andrewda:
@wcastello Awesome and very fast! Also glad to know we got another Oregon guy working on this, heh.

SimonTellsAll:
Sounds, amazing. I would be interested if we could have multiple user accounts running simultaneously, or even switching between accounts between searches. Might lead to less soft-bans.

wcastello:
@jangrewe it's topping 100% but python only gets one cpu so it doesn't matter much. The browser kind of dies trying to load all those markers though.

joveice:
This is great! please get this merged, it's really use full!

andrewda:
@wcastello 
> The browser kind of dies trying to load all those markers though.

This seems like an incredibly good issue to have.

wcastello:
Something funny with this search algorithm.. since it takes quite a bit to scan and it goes in that sort of circular path the center starts to die when the Pokemons there start to expire, so there's probably a limit on how much this is useful: 
screen shot 2016-07-22 at 5 41 25 pm

joveice:
It's still usefull to refresh the area faster with a smaller area, have a little area myself and the middle dies out for 5 min before it refreshes.

AHAAAAAAA:
rebase and I'll merge, this is perfect.

pogom/utils.py
```diff
@@ -55,6 +55,7 @@ def get_args():
     parser.add_argument('-k', '--google-maps-key', help='Google Maps Javascript API Key', default=None, dest='gmaps_key')
     parser.add_argument('-C', '--cors', help='Enable CORS on web server', action='store_true', default=False)
     parser.add_argument('-D', '--db', help='Database filename', default='pogom.db')
+    parser.add_argument('-t', '--threads', help='Number of search threads', required=False, type=int, default=1, dest='num_threads')
```
AHAAAAAAA: set the default to 5

wcastello: Alright, give me a second.

Liliexx:
, tbh not the fast as you say , and make loads of losing data , need to be fixed

wcastello:
@AHAAAAAAA done.

AHAAAAAAA:
You are perfect

wcastello:
@Liliexx it needs work, help is welcomed, I'm still trying to figure out a solution to the db locks.

medghaim:
@wcastello I'm looking into that as well

medghaim:
Awesome work, btw!

shivampaw:
How many -st can it do in 3 mins?

slonce:
Good afternoon. I did not understand why, when the use does not show a lot of Pokemon? Although the area scanned? It is necessary to use MySql?

jangrewe:
The amount of Pokémon is identical for me, it just is (number-of-threads) times faster.

shivampaw:
Many users faced issues with this. Setting it to 1 thread made all Pokemon show up again. I haven't tested yet but I think something will need to be done.

PR #1396, opened 2016-07-22
Fix scan percentage calculation

EdenGottlieb:
This pull request includes a fix for calculation of completed scan percentage which was broken. 

For some reason it was step/num_step^2*100 (num_step being the raw step limit away from user and not the number of steps required to scan the lattice).

Now it is (step/total_steps)*100 (Brackets only here for clarification, code doesn't need it because of order of operations

- [x] Bug fix
- [ ] New feature
- [ ] Translation

- Changed calculation of percentage
-
-

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -141,7 +141,7 @@ def search(args, i):
             else:
                 log.info('Map Download failed. Trying again.')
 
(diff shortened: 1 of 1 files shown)


shivampaw:
Same as #1374

EdenGottlieb:
You're right, didn't notice that. closing pull request

PR #1401, opened 2016-07-23
Sidebar

joveice:
option to turn on and off select new area in a config file?


joveice:
@clearminds This isent mine, I was going to create a feature request and ended up doing something I dident plan on doing :P

PR #1414, opened 2016-07-23
Overall area marker

hazanjon:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

As referred to / requested in #1338, this feature adds an overlay that shows the current overall search area for any active workers (Last 15 mins) 

This should make it easier for users with multiple workers to check and align their search areas

The option can be toggled independently to the actual scanned areas

worker area
worker area in progress

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -44,12 +44,10 @@ class Pogom(Flask):
             d['scanned'] = ScannedLocation.get_recent()
             
         if request.args.get('location', 'false') == 'true':
-            locations = WorkerLocation.get_all()
-            
+            locations = WorkerLocation.get_recent()
+
             for l in locations:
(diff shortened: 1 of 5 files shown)


hazanjon:
@jangrewe @HinoReo Was this the sort of thing you two were looking for?

HinoReo:
@hazanjon Yes, that's exactly what I meant! :dart:

jangrewe:
@hazanjon pretty much, yes! =)

could it be inverted, though, like a real fog-of-war? so that the markers on the map don't get an overlay (or rather "underlay") and the map style doesn't interfere with it?

EnigmaIsWorking:
Is it possible to have Multithreading for checking Pokemons in each area? If I want to observe 30 kilometers, it takes to much time for refreshing.

Would be really happy that I can give him 5 Accounts and they are splitting the work (faster).... Like a sonar.

PR #1415, opened 2016-07-23
Added a "Notify for all except" filter

ravishivt:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- The notification feature is great but I'd like to be notified of any non-common pokemon.  To do that, I would need to enter 100+ names into the filter.  This change adds a second select2 box that will raise a notification anytime a new pokemon is detected that is not in the list.  The second "exclude" select2 trumps the "include" one if at least one item is entered into it.

Before using existing notification select2, over 100 items selected:
image

After:
image


AHAAAAAAA:
Approach isn't clean. What we could do instead is use both exclude and notify to implement this:

Exclude: x, y, z pokemon
Notify: All (we need to add this all selector)

ravishivt:
I disagree on using the existing exclude to filter out pokeman that shouldn't be notified on.  I want to use the map to inform me of all pokeman in my area but only notify me on all but the most common ones.  If I'm going to walk to a notified slowbro 3 blocks away, I want to see all the other pokemon I could catch along that route including the common ones.  This approach would not be possible with the approach you describe and implemented in #1555.

PiaFraus:
@ravishivt This would require some refactoring, since currently exclude filter filters the incoming data from server. There would be nothing to notify about. But with latest changes we can change it so it will just mark the data as hidden and not create marker for it.

ravishivt:
@PiaFraus That's assuming you are coupling the exclude filter with the notify filter.  My approach does not couple the two.  You wouldn't haven't to use the exclude filter at all.  You let everything come in from the backend and use a "notify exclude" filter to notify on all except.

PR #1422, opened 2016-07-23
removed location as required argument

Roastmaster:
automatically determine the latitude and longitude of user if no location is specified


MonkeyLeeT:
Don't really think it's good idea to rely on something like that.

PR #1427, opened 2016-07-23
Scanned Pokémons/Gyms/PokéStops Statistics

Retalyx:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

-Added scan statistics in the navbar
-You can see stats for all displayed Pokémons, Pokéstops, and gyms

capture
capture


andrewda:
Looks good! Not sure what others prefer or think looks best, but it might be better to have the total on the bottom - that's usually the standard.

Retalyx:
Total on bottom? Why not :) If more people prefer total on bottom, i'll change it ;)

ditchfieldcaleb:
I'd prefer total on top. Would rather not have to scroll all the way down if there are many different pokemons.

JZ6:
this might slow stuff done, perhaps add a toggle

Retalyx:
Why ill this slow stuff down? It's client side and only updated when updateMap() is called :) i tested it multiple Times with hundred of pokémons without any visible slowing down

Shulin:
This isn't populating on mobile.

PR #1433, opened 2016-07-23
Added gps tracking switch, making use of next location code

omarzion:
## Description
Added gps-switch toggle below scanned locations toggle.
Added trackGps to localStorage
create updateGps function that sends /loc?lat= lon= every 3 seconds when toggled on
defaults to off
## Motivation and Context
I think gps tracking would be a great feature to have in this map and is the only feature I really think the project is lacking so far.

The only downsides that I can see, is without using ssl it will not work in chrome since chrome now ignores gps requests without https as stated here https://developers.google.com/web/updates/2016/04/geolocation-on-secure-contexts-only?hl=en

## How Has This Been Tested?
Confirmed working with Dolphin Browser on android and Firefox on linux.
Firefox app is unresponsive (I believe from show scanned)
ran server for 2 hours, no issues
## Types of changes

## Checklist:

	https://developers.google.com/web/updates/2016/04/geolocation-on-secure-contexts-only?hl=en


frdmn:
This should be optional. I use the map with my friends and don't want to update based on GPS.

omarzion:
It has a toggle in the side panel like show scanned region etc and defaults to off

talentedmrripley:
#725 already does this.

georgo:
* It would be nice, if "Track GPS" would disable, when I manually type new address.
* next_loc path shouldn't be absolute (remove root /)

AHAAAAAAA:
Problem I see with this is that the exact point of location could get very confusing. IT can follow you, or be the initial location you entered or be the new location you entered in the sidebar. Very easy to lose track of where it is!

omarzion:
I completely missed #725 from the looks of all the comments that will be the one merged.  Nevertheless I updated to make it compatible with base branch again. 
@georgo thats a good idea! will add feature to turn off gps tracking when a new location is typed in.
@AHAAAAAAA I think using #1484 along with this would fix that issue

MonkeyLeeT:
Thanks for your understanding.

PR #1443, opened 2016-07-23
Map: move fullscreen toggle to left top.

KhasMek:
## Description
Moves the full screen toggle from TOP_RIGHT to LEFT_TOP.

## Motivation and Context
Better UX

## How Has This Been Tested?
Chrome, Firefox, Chrome (Android)

## Types of changes

- [x] UX modification

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -73,6 +73,9 @@ function initMap() {
         },
         zoom: 16,
         fullscreenControl: true,
+        fullscreenControlOptions: {
+                position: google.maps.ControlPosition.LEFT_TOP
+        },
         streetViewControl: false,
 		mapTypeControl: true,
 		mapTypeControlOptions: {


MonkeyLeeT:
Can you describe how left_top is better than default right_top?

KhasMek:
I can, you can also look while you're testing the patch. With the map type selector on right_top and the zoom button on **top_right** it made the map selector really stick out and look out of place. Especially on mobile after the layout was fixed yesterday.


As a side note, I've been submitting to OSS for decades, am the founder of OSS projects that have dealt with hundreds of submissions a day and I have to say that this was one of the most ridiculous questions to come from any project collaborator. Having a proper, clean, and uniform submission process is most critical. However, this isn't a school project, and logic for something so simple doesn't need to be questioned as a CS professor would. Running a project as such will do nothing but turn good developers away due to being overly questioned. All the information pertinent to the commit should *be within the commit message itself* that way the logic for the change is there forever regardless of the platform the code is being viewed on. Being forced into redundancy is not something most developers a favored pastime of developers in my experience either. It's simple - a single width icon at the top of the draw area, proceeded by a double width icon and finally another single width icon **is just bad UX** and any one with merge abilities should know as much. My intention with this comment was not to attack you personally, but again submit my knowledge to the project in hopes of making it better, all around.

AHAAAAAAA:
Good points. Let's do this instead to be more in line with how Google designs their map interfaces:

Full screen top right, selector bottom right below zoom buttons and current location.

KhasMek:
I definitely understand what you mean @AHAAAAAAA,. but in all of Google's documentation they insinuate the map selector should be positioned at left_top, using the default horizontal_bar control style. In my commit I left it as a dropdown menu, as it looks much better on mobile than the horizontal bar with the amount of map types offered. It *could* be moved to bottom_right, or right_bottom. However that would require a change in the actual function of the dropdown menu, as it would be a flyup menu (which is not part of their api TMK and would require more custom code) if we wanted to see any of the controls. ;)

Here's my reference for saying that Google insinuates placing map selector menu at left_top
 - https://developers.google.com/maps/documentation/javascript/examples/control-default

KhasMek:
Here is a screenshot of what it looks like at left_top

pokemap_left_top

JZ6:
merely a cosmetic change, if its not broke why fix it?
plus the toggles bar is on the right, might run into troubles eventually.

KhasMek:
Seriously? It **is** broke though. Do you know what UX is?

PR #1452, opened 2016-07-23
added directions div to gym marker

polyisobutylene:
Directions link was missing on gym markers, was there for pokestops and pokemon.

Added in the same fashion as for the pokemon and pokestops.


Bitti09:
Can you add a Screenshot ?

fraserrj:
gymdirections

PR #1508, opened 2016-07-23
Set default thread count to 1 on Windows until locking problem is sorted out

marauder37:
## Description
Sets default thread count to 1 on Windows until locking problem is sorted out

## How Has This Been Tested?
Tested on Win 7 and OSX

## Types of changes

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -5,7 +5,7 @@ import logging
 import time
 import math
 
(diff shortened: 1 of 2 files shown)


Pedrock:
Multithreading should be disabled when using sqlite.

marauder37:
@Pedrock even with locking around the db write?

PR #1535, opened 2016-07-23
Handle db OperationalError exception and retry InsertQuery()

wcastello:
This deals with the OperationalError exception that comes when the db locks and keep trying to insert that so we don't lose data on the way. 

I ran it with several combinations of threads on OS X and it goes well, gets some lock eventually but with this it retries. Apparently people on Windows are having a harder time. 

Try running a version of sqlite3 >= 3.7.15 for pragma busy_timeout.

I'm having trouble understanding the deal of concurrency with sqlite, so if you're good with please come to save. 

- [x] My code does not follow PEP8.


wcastello:
Hell, hang on.

PR #1537, opened 2016-07-23
Fix multiple password request from console

SilverC:
## Description

Fix request for multiple passwords when -p argument is left out

## Motivation and Context

Currently a password is requested three times in a row due to get_args being called three times and the password not being stored anywhere

## How Has This Been Tested?

Tested locally for google login

## Types of changes

## Checklist:

### DIFF
--- a/pogom/__init__.py
+++ b/pogom/__init__.py
@@ -10,5 +10,6 @@ config = {
     'GMAPS_KEY': None,
     'REQ_SLEEP': 1,
     'REQ_HEAVY_SLEEP': 30,
-    'REQ_MAX_FAILED': 5
+    'REQ_MAX_FAILED': 5,
+    'PASSWORD': None
 }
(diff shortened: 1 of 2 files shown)


gudipudiprashant:
Noob here, can you explain why the function is being called thrice?

gudipudiprashant:
Noob here, can you explain how the function is being called three times?

SilverC:
get_args was being called in three separate locations so if you don't pass the password in as the -p argument then the get_args would ask for the password three times. It's actually called more now, twice in app.py and once in models.py, customLog.py and runserver.py. Which means that if this fix wasn't in it would ask for your password 5 times now

gudipudiprashant:
Didn't notice it being called in different .py files. Thanks :)

gudipudiprashant:
Didn't notice it being called in different .py files. Thanks :)

C-5:
Please

PR #1435, opened 2016-07-23
Fix database lock hang

JohnMachete:
## Description
Just added a try/except in the while loop of bulk_upsert.  Tested for a few hours and instances no longer hang/display database lock message. This should now continue to try to insert the row until successful.

## Motivation and Context
Scripts were hanging when running multiple instances at once due to database locks.

## How Has This Been Tested?
Tested for a few hours via 7 simultaneous instances, no lock errors/no script hangups.

## Types of changes

- x Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Fixes Issues
#1384  #1319  #1028  #998  #744

PR #1538, opened 2016-07-23
handle OperationalError db lock exception, add pragma busy_timeout 30000

wcastello:
This deals with the OperationalError exception that is raised when the db locks and then it keeps trying to insert that data so we don't lose it on the way. Also adds pragma busy_timeout 30000 to sqlite. 

I still get some locks eventually but at least it retries. Apparently people on Windows are having a harder time with this.

Try running a version of sqlite3 >= 3.7.15 for pragma busy_timeout.

I'm having trouble understanding the deal of concurrency with sqlite, so if you're good with this please come to save. 

- [x] My code does not follow PEP8.

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -3,7 +3,8 @@
 
 import logging
 from peewee import Model, SqliteDatabase, InsertQuery, IntegerField,\
-                   CharField, FloatField, BooleanField, DateTimeField
+                   CharField, FloatField, BooleanField, DateTimeField,\
+                   OperationalError
 from datetime import datetime
 from datetime import timedelta
 from base64 import b64encode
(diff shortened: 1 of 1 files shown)


wcastello:
Oh, apparently #1435 did this too.

AHAAAAAAA:
Looks clean, but my Windows machine is on life support. Can we get some testing on this Win people?

pogom/models.py
```diff
@@ -183,12 +186,17 @@ def bulk_upsert(cls, data):
...
-        log.debug("Inserting items {} to {}".format(i, min(i+step, num_rows)))
-        InsertQuery(cls, rows=data.values()[i:min(i+step, num_rows)]).upsert().execute()
+        while True:
+            try:
+                log.debug("Inserting items {} to {}".format(i, min(i+step, num_rows)))
+                InsertQuery(cls, rows=data.values()[i:min(i+step, num_rows)]).upsert().execute()
+                break
+            except OperationalError as e:
+                print('{}... Trying again!'.format(str(e)))
+                time.sleep(1)
```
TomProkop: `import time` is missing

wcastello: My bad, this was fixed by some other PR anyway so I'm closing this one.

Issue #1544, opened 2016-07-23
Feature request: Wildcards for notification list


nbbsn:
I'd love a wildcard feature for the "Notify of Pokémon" filter so you'd get a notification anytime a Pokémon spawns that is not filtered out by the "Hide common Pokémon" filter. 
I think this would especially help players like myself who live in smaller towns that do not see a lot of pokemon spawns.   
Adding every Pokémon to the notification list manually is pretty tedious.

Halvo317:
Yeah. Removing the -i and -o options was a huge mistake in my opinion.

PR #1555, opened 2016-07-23
Add option to notify for all Pokémon, default to off.

DJScias:
## Description

Adds an option to notify for all Pokémon that are found.

- I think the reason why this is off by default speaks for itself.
- Please don't do this in a big city, the notification spam will continue until you close the window.
- Requested through Discord for the purpose of 'smaller towns' and such.

## Motivation and Context

While I don't recommend doing this in a big city, I can understand why people in a remote or small town with a low spawn amount want to be able to be kept notified.

Fixes:
#1544 

## How Has This Been Tested?

Tested it by running it at a remote location, notifications popped up for every single found Pokémon. Tried it at a big city as well and had to close the window to stop the notification spam.

Screenshot
The "All Pokémon" text can be changed if necessary, I had little imagination when I wrote that but it works.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -137,6 +137,7 @@ function initSidebar() {
     $('#pokemon-switch').prop('checked', localStorage.showPokemon === 'true');
     $('#pokestops-switch').prop('checked', localStorage.showPokestops === 'true');
     $('#scanned-switch').prop('checked', localStorage.showScanned === 'true');
+    $('#notifyall-switch').prop('checked', localStorage.notifyAll === 'true');
     $('#sound-switch').prop('checked', localStorage.playSound === 'true');
 
     var searchBox = new google.maps.places.SearchBox(document.getElementById('next-location'));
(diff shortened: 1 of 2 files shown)


AHAAAAAAA:
I like this. Related to: #1415 and #1322

AHAAAAAAA:
Tested, works pretty well! +1

Ashex1:
Since you were getting spammed in big cities can we add some sort of rate limiting for notifications? i.e, no more than _n_ notifications over a 10 second period.

PoorNewb23:
Alternatively, can we make it all minus "hide common"?

andrewda:
Just for styling purposes, please make the "All Pokemon" smaller than "Notify of Pokemon" just so it can be clearly seen as an option, not a header.

AHAAAAAAA:
This also attempts to address all the requests for 'Notify for all except hidden pokemon' #2942.

Issue #1394, opened 2016-07-22
Javascript Errors


macgngsta:
## Expected Behavior
no red in my browser console

## Current Behavior
- missing cross.png (this is referenced in css)
- uncaught reference error (map.js:258)
- uncaught type error (map.js:126)

## Possible Solution
- add cross.png to /static/dist/img/cross.png
- put the google variable in scope for the marker
- define the searchBox in the map.html

## Steps to Reproduce (for bugs)
1. run server as usual 
2. load map

## Your Environment
* Version used: 3ba6b157
* Environment name and version (e.g. Python 2.7): 2.7.12
* Operating System and version (desktop or mobile): desktop

lilchancep:
I think this is what is triggering the map to dissapear when trying to look at it on iOS

BillyWM:
I see a map behind your console so I can see it does load, so some of the map rendering code does run.

Scope isn't the issue. Once the Google maps API script is loaded, google goes into the global scope, available anywhere. What must be happening is a race condition where stray code is accessing this before it's loaded.

What's actually happening here:
1. map.js loads
2. Line 479 invokes updateMap();
3. updateMap() kicks off a JSON load
4. The load callback references setupPokemonMarker
5. setupPokemonMarker references google.maps.Marker

If the google maps API isn't done loading and initializing at this point, there's a ReferenceError. However, the map loads later and tries again. Plus, line 478 earlier had set up a repeat call via SetInterval

Haven't reproduced it but that's my assessment based on eyeballing

For searchBox, make Google maps loaded but the places API didn't.

-------------------

Lesson?

the map.js code is set up wrong. ***nothing should run until DOMContentLoaded fires***. There should be a global DOMContentLoaded listener that kicks off a function that sets everything in motion.

chuyskywalker:
> nothing should run until DOMContentLoaded fires

A good point, especially considering jquery seems to be required in anyway...

chuyskywalker:
https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1464 Should do the trick

PR #1556, opened 2016-07-23
Fix missing image in #1394 by removing unused overlay code

chuyskywalker:
The overlay code is unused, but included as invisible items. As such the CSS is trying to load the image. No such image exists. I couldn't track down the _actual_ correct image, but this is all unused code. So I just pulled it out instead of including a bogus image.


MRokas:
Do not include .min files next time please.

chuyskywalker:
@MRokas any guidance on how the minified stuff is supposed to be handled at this point? I did it cause...well, grunt changed it. But I don't think the codebase is actually referencing any of the built stuff right now. (Well, sorta, the compiled `main.css` gets directly referenced, but not the `app.min.js` final stage...)

MRokas:
don't do `git add .`

Either add->commit->push and only then build, or add only appropriate files.

PR #1561, opened 2016-07-23
Isolate web client from server code

coolaj86:
Minimal steps needed to separate server from client.

I moved the web client configuration from the server `config.ini` to the browser `static/config.js` (note: the languages are still named incorrectly, as they were, but I made them work anyway by loosely matching).

Since the server and the client have nothing to do with each other, it makes more sense to pull them apart so that they can be used with other servers and clients (as others have also made mention)

I know that you're going to close this (a la https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1532), but I'll try to take both, leaving all of the server and client code in the same repository, and then submit the pull request with both sets of changes so that although all of the code will remain in the same repo, the confusing bits of intertwined client/server logic will be teased apart.


AHAAAAAAA:
You are correct, there's no argument here. I'd like to do this, but down the line. I'll keep an eye on your project for now, once we get to the point where the separation makes sense then I will get in touch!

PR #1566, opened 2016-07-23
Cleanup map.js

chuyskywalker:
- Use jquery instead of direct JS in a few places
 - Move vars/functions/onload code into distinct sections
 - Standardized on `function name(params)` style*
 - Ensure that all functions which bind, manipulate, etc are inside a proper domcontentloaded so they can't race condition and throw errors

\* I have no prefernce on this format vs others, it was just the most common, so I went with it

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -1,60 +1,36 @@
-$(function () {
-    if (!Notification) {
-        console.log('could not load notifications');
-        return;
-    }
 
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
Tested, works! Thank you for the cleanup. It's got my +1.

JZ6:
merged

PR #1571, opened 2016-07-23
Add ability to get raw Pokémon filtered by id

fettlaus:
Added a parameter for filtering specific Pokémon.

## Description
Added a paramter to the `/raw_data` endpoint for requesting only selected and active Pokémon by their id. Just add a parameter with comma separated values to the request like this: `http://localhost:5000/raw_data?gyms=false&scanned=false&ids=13,16` and the server will only respond with active pidgeys and weedles.

## Motivation and Context
Using the raw data as an endpoint for further development for example as input to a mobile application, we need some way to filter on the server side. Especially regarding mobile data usage.

## Types of changes

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -55,6 +55,21 @@ class Pokemon(BaseModel):
 
         return pokemons
 
+    @classmethod
+    def get_active_by_id(cls, ids):
+        query = (Pokemon
+                 .select()
+                 .where((Pokemon.pokemon_id << ids) &
+                        (Pokemon.disappear_time > datetime.utcnow()))
+                 .dicts())
+
+        pokemons = []
+        for p in query:
+            p['pokemon_name'] = get_pokemon_name(p['pokemon_id'])
+            pokemons.append(p)
+
+        return pokemons
+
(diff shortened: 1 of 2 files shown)


pogom/models.py
```diff
@@ -55,6 +55,21 @@ def get_active(cls):
 
         return pokemons
 
+    @classmethod
+    def get_active_by_id(cls, ids):
+        query = (Pokemon
+                 .select()
+                 .where((Pokemon.pokemon_id << ids) &
```
MonkeyLeeT: Shifting operator with list? How does this work?

fettlaus: It's a function provided by peewee replacing the  pythonic `in` operator.

Am 23.07.2016 21:46 schrieb "Ted Li" <notifications@github.com>:

MonkeyLeeT: Okay, hopefully no one in this project will wanna use shift operator...

craywong:
Is it just me that this feature is not working?

fettlaus:
What was the URL you called?

PR #1574, opened 2016-07-23
Move configs to examples

invisiblek:
## Types of changes

## Checklist:

* .gitignore doesn't ignore these because they exist
* move them to example files that users can copy to non-example
  files which gitignore will now ignore


pogom/utils.py
```diff
@@ -25,7 +30,14 @@ def parse_unicode(bytestring):
     decoded_string = bytestring.decode(sys.getfilesystemencoding())
     return decoded_string
 
+def verify_config_file_exists(filename):
+    fullpath = os.path.join(os.path.dirname(__file__), filename)
+    if os.path.exists(fullpath) is False:
+        log.info("Could not find " + filename + ", copying default")
+        shutil.copy2(fullpath + '.example', fullpath)
```
MonkeyLeeT: But after copying, isn't it still *.example.json which is not loaded by the rest of program? We should also rename it and take out "example" as a rescue.

invisiblek: nah, it copies config.ini.example to config.ini, before the program attempts to load config.ini

MonkeyLeeT:
Oh just realized that. Rebase and let's make it happen!

PR #1579, opened 2016-07-23
use database url to allow all of peewee supported databases

Chlodochar:
## Description

This changes the use of `-D` to allow database urls to be supplied instead of only SQLite database filenames. It reuses Peewee database urls to allow any type of Peewee-supported database (like MySQL and Postgres). See http://docs.peewee-orm.com/en/latest/peewee/playhouse.html#database-url for more information.

The previous usage of `-D` is still supported. If the parameter does not have an URL (with scheme://) then SQLite is used.

Example usage:
```
python runserver.py ... -D "sqlite:///mydatabase.db"
```
or
```
python runserver.py ... -D "postgresql://postgres:my_password@localhost:5432/my_database"
```
or
```
python runserver.py ... -D "mysql://user:passwd@ip:port/my_db"
```
or without URL will default to SQLite:
```
python runserver.py ... -D "mydatabase.db"
```

## Motivation and Context
Will allow usage of Postgres and MySQL.

## How Has This Been Tested?
Using the parameter to change the URL to another database.

## Types of changes

## Checklist:

I can't change the documentation of https://github.com/AHAAAAAAA/PokemonGo-Map/wiki/Usage
I'd add:

> usage of -D also supports database URLs to support database backends other than SQLite. See http://docs.peewee-orm.com/en/latest/peewee/playhouse.html#database-url for examples.


Pedrock:
This wont work with postgresql. Upsert is not supported, only on sqlite and mysql.

Ashex1:
The reuse of the db flag is a bit confusing, would a person still be able to point it to a sqlite db file? I don't see a test for this in the changes. 

Sidenote, I just noticed there's both a `-d` and `-D` flag...

Also don't understand why you removed the ability to provide these settings in a config file.

Pedrock:
Please check #1618. I have a working implementation for PostgreSQL.

Chlodochar:
@Ashex1 See https://github.com/Chlodochar/PokemonGo-Map/blob/961368ccc35c4ad259b18ffaceacb75454ef698a/pogom/models.py#L29-L31

The change is very minimal because it uses the Peewee database urls.

I made this PR right after the MySQL PR was merged, so I only used `-D` and no `config.ini` (as it didn't use that before this PR). Reverting the other PR was probably a mistake on my part.

jmbeach:
Apparently the app spawns a new connection for each operation and it does it multithreaded so it requires a database which can handle many connections and it is more than the free tier of cleardb on heroku which gives you a max of 10

jmbeach:
I limited the thread count, but the problem is still there which makes me think that the connections aren't being closed properly.

PR #1581, opened 2016-07-23
Thread size=1

AHAAAAAAA:
Until we solved all the DB issues with multi-threading, keep thread at 1. Better to have accurate slower results than fast inaccurate onces.


relrod:
@AHAAAAAAA Has this bug been fixed yet in `develop`?

PR #1587, opened 2016-07-23
Japanese names file has not been merged

takutok:
Japanese names file(pokemon.ja.json) has not been merged.
pokemon.jp.json is incomplete translation.


AHAAAAAAA:
You're correct, later pokemons are in English. Can you change the filename to pokemon.jp.json?

takutok:
Correct japanese language code is ja. 
So change to ja.json from jp.json .

AHAAAAAAA:
You're correct.

PR #1612, opened 2016-07-23
move client config to client

coolaj86:
In relation to https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1532 and https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1561

Here's a small set of changes, without removing code, that allows the client to be developed independently of the server.

The client config variables are moved to config.js, which can are still be templated by the python server, but can just as easily be used with any other server as a static file with no code changes.

I'd like to be able to continue to develop this web client and track its changes (as would others) and this is a good first step.

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -15,18 +15,22 @@ class Pogom(Flask):
     def __init__(self, import_name, **kwargs):
         super(Pogom, self).__init__(import_name, **kwargs)
         self.json_encoder = CustomJSONEncoder
-        self.route("/", methods=['GET'])(self.fullmap)
         self.route("/raw_data", methods=['GET'])(self.raw_data)
         self.route("/loc", methods=['GET'])(self.loc)
         self.route("/next_loc", methods=['POST'])(self.next_loc)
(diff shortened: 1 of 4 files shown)


coolaj86:
whoops, this should have been a PR to the develop branch... seeing if I can fix that...

BillyWM:
Looks like a good start on cleaning up.

One gripe (beside not liking the generic descriptor "helper"): why not document.body? it works even back to IE6.

coolaj86:
Sure, let's use `document.body`. What would you like the `initMapHelper` to be called?

I mean, currently the code can't run in strict mode and has very little separation of concerns. I'm all for simplifying and cleaning up, but I don't even know what the codestyle is supposed to be (and it clearly doesn't follow one as of yet).

If you want I can turn this into proper ES5.1 Crockford-approved code in an hour, but I'm not going to invest a lot of time into making classes and functions and whatnot if I'm not gonna get my PRs pulled in.

So tell me how you want it done and I'll start making the code great and opening places to insert the functionality I need and for others to extend.

PR #1640, opened 2016-07-23
Revert "Use a sprite for the pokemon on the map" https://imgur.com/a/fWanT

JZ6:
https://imgur.com/a/fWanT


Frugghi:
@invisiblek 
```Python
var y = Math.ceil(num / 12) - 1;
```

Jankos1985:
@Frugghi 
Wouldn't that put you two rows down too far?

I think it needs 
`var y = Math.floor(num - 1) / 13;`
since there are 13 rows

Frugghi:
@Jankos1985 
It doesn't matter how many rows you have, for example with 24 rows your math is wrong.

ceil round the number to the next integer so `Math.ceil(num / 12)` with **num** from 1 to 12 is 1, from 13 to 24 is 2, etc.

PR #1129, opened 2016-07-22
Make position marker draggable for location change

POPSuL:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

- Current location marker maked draggable

If this is related to an existing ticket, include a link to it as well.

This PR is replace for https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1111

PR #1650, opened 2016-07-23
make marker position draggable

Chlodochar:
## Description
Allows dragging of the search marker. Instead of typing the place to put the searchmarker in the sidebar, you can just drag it.

## Motivation and Context
It's convenient.

## How Has This Been Tested?
By dragging the marker on my webserver and seeing it search a different area.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -88,18 +88,43 @@ function initMap() {
 
     map.setMapTypeId(localStorage['map_style']);
 
-    marker = new google.maps.Marker({
+    marker = createSearchMarker();
+
+    addMyLocationButton();
+    initSidebar();
+};
+
+function createSearchMarker() {
+    var marker = new google.maps.Marker({
         position: {
             lat: center_lat,
             lng: center_lng
         },
         map: map,
-        animation: google.maps.Animation.DROP
+        animation: google.maps.Animation.DROP,
+        draggable: true
     });
 
-    addMyLocationButton();
-    initSidebar();
-};
+    var oldLocation = null;
+    google.maps.event.addListener(marker, 'dragstart', function() {
+        oldLocation = marker.getPosition();
+    });
+
+    google.maps.event.addListener(marker, 'dragend', function() {
+        var newLocation = marker.getPosition();
+        changeSearchLocation(newLocation.lat(), newLocation.lng())
+            .done(function() {
+                oldLocation = null;
+            })
+            .fail(function() {
+                if (oldLocation) {
+                    marker.setPosition(oldLocation);
+                }
+            });
+    });
+
+    return marker;
+}
 
 function initSidebar() {
     $('#gyms-switch').prop('checked', localStorage.showGyms === 'true');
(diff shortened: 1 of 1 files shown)


Kryakz:
I cant see pokestops or gyms

marauder37:
Really? Seemed to work ok for me

MonkeyLeeT:
#1129 Any improvement?

Chlodochar:
@MonkeyLeeT Most importantly, it's mergable on develop.

Idhanta:
Unable to drag it in Safari

Chlodochar:
@Idhanta any error you came across?

marauder37:
There's no error in Safari, the whole map just moves when you drag the pin

Idhanta:
exactly what marauder said.
Also, the marker moves in Chrome, but no pokemon show up.

Idhanta:
Ran the code in debug mode and the marker was magically movable in Safari. Chrome still no pokemon.

Chlodochar:
@Idhanta Maybe some pokemon was blocking the marker?
As for not seeing pokemon appear, is that actually related to the marker dragging or are they hidden regardless of the marker being dragged? See the sidebar and check whether "change your location" actually works.

Idhanta:
@Chlodochar could be a pokemon blocking. Not sure. 
For chrome, no pokemon load up regardless of the marker being dragged.

Idhanta:
@Chlodochar Little more clarity for the chrome issue, if I run on localhost, it doesn't load the pokemon, while running with 0.0.0.0, it loads them and everything works. No idea why.

Chlodochar:
Are pokemon actually enabled in the sidebar? I've noticed there is a bug related to those filter options (localStorage). I'll check that out and make a PR if it's actually a bug and not something else. I presume localStorage differs between domains.

Also, cache maybe?

Idhanta:
ok, so clearing cache worked. 
Everything looks good for both browsers now.

Chlodochar:
@Idhanta yay, thanks for testing.

Chlodochar:
@marauder37 thanks!

Hoekynl:
Could you implement a button which resets the marker (and scan) to its original location?

JZ6:
+1 reset please

Chlodochar:
Hmm, it seems nicer to have a list of bookmarked locations or something like that? What would be the workflow for resetting the location?

Chlodochar:
At the moment a server restart does reset the location. So it could probably be retrieved from commandline or config.

Hoekynl:
Yeah I think it should be something like
Original location = (get from command execution or config file)
When the button is pressed, set location to original location

primaxius:
Need a way to disable this!

Kryakz:
Why? You don't have to drag it

Hoekynl:
But if im outside, on my phone and im at the default location, walk away and want to scan tjat area. Then I drag the marker. If i walk back I want to be able to reset it

Kryakz:
Or just drag it back...

Hoekynl:
I use the "show scanned locations". If i drag it back its not precisely on the last location and there would be red and green circles over eachother making it hard to see pokemon.

Chlodochar:
@Hoekynl IMO the problem in that case is that the search-cells do not align well. It would be better to split the world into a 'grid' and use the cells of that grid to search. Now you can have off-by-0.0001 errors that cause extra search-cells.

PR #1661, opened 2016-07-23
Add time information in db

Dutzen:
Currently only despawn is listed in the db.
If we can add a spawn time (or even the time the entry was made) we can analyze spawn pattern for specific areas


MonkeyLeeT:
Please PR into develop instead of master.

PR #1675, opened 2016-07-23
Use grunt-babel with babel-polyfill for ES6 transpilation

ravishivt:
## Description
This is a rebase of #1140 and #1119 without the jsHint changes.  It adds grunt-babel to transpile ES6/ES2015 code to ES5.  It also uses babel-polyfill as suggested by @nekitk and @BillyWM for better browser support. 

## Motivation and Context
Without transpilation, ES6 style syntax such as template strings will cause exceptions in browsers that don't support it, such as IE and iOS 8 devices.

## How Has This Been Tested?
Tested to make sure app loads on iOS 8 Chrome which was failing previously.

## Types of changes

## Checklist:


AHAAAAAAA:
Looks good, rebase!

ravishivt:
rebased

templates/map.html
```diff
@@ -148,9 +148,11 @@
 	</div>
 
 	<!-- Scripts -->
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/6.9.1/polyfill.min.js"></script>
 	<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.1.0/jquery.min.js"></script>
 	<script src="https://cdnjs.cloudflare.com/ajax/libs/skel/3.0.1/skel.min.js"></script>
 	<script src="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.3/js/select2.min.js"></script>
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/6.9.1/polyfill.min.js"></script>
```
MonkeyLeeT: Why include twice, and indentation?

templates/map.html
```diff
@@ -161,7 +163,7 @@
 	</script>
 
 	<script async defer src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places,geometry"></script>
-	<script src="static/map.js"></script>
+	<script src="static/dist/js/map.js"></script>
```
MonkeyLeeT: Since we're all editing static/map.js, will this impact that?

ravishivt: Yes and no.  The file this is referencing is static/dist/js/map.js which is what grunt generates as a distributed file so that doesn't change much from development perspective.  I did however move static/map.js to static/js/map.js as it made sense to me to put the source JS files in the js directory.  That change will impact devs in that they need to start editing static/js/map.js.

templates/map.html
ravishivt: Fixed, thanks.

templates/map.html
MonkeyLeeT: I understand your concern and completely agree that. But for now, can you move it back so that most dev PR will still be fine?

You're welcome to send a quick PR to move it when we're down to less than 10 PRs and ready for 2.0 release.

ravishivt: done

MonkeyLeeT: My last question, doesn't this now require that we constantly grunt to get changes from static/map.js to dist/js/map.js, otherwise our test users won't be able to see the new features added to static/map.js?

Any way to get around that and avoid the effort?

BillyWM: Yes, it does. That's by design. It's processing and bundling the files each time.

No, you can't get around it. Using ES6 with Babel, you have to transform your code. It can't run as-is.

For end users, pre-compiled js files should be shipped with the repo. For develop, this grunt step is necessary for developers.

BillyWM: Also one of the grunt subtasks is grunt-watch. This monitors for changes and re-minifies as long as it's running. You can leave it on in the background as you code.

BillyWM: **moving files**: Git can track this. The history will follow the file.

MonkeyLeeT: So how can we make sure that those developers not familiar with Grunt (there are), will know to run grunt to keep changes synced? They may look at map.html and think that dist/js/map.js is the new one and start editing it instead of static/map.js.

BillyWM: In other discussion I had suggested writing a FAQ. Some will make mistakes but this is a super common setup so there will be a segment that just "get it" (like @ravishivt and myself).

Now that the wiki seems to be unlocked I'd be willing to help write a developer-centric getting started page.

MonkeyLeeT: Appreciate the effort.

MonkeyLeeT:
You have my +1. But changes like this will need input from other collaborators.

BillyWM:
I started an instructions page here to ease the transition:

https://github.com/AHAAAAAAA/PokemonGo-Map/wiki/Development

AHAAAAAAA:
Alright, ready for this merge. Rebase.

I don't think it'll cause too much of an issue, running `npm install` and `grunt` is pretty easy.

MonkeyLeeT:
One more rebase and we'll merge.

@BillyWM Could you also edit the document to specify how to kill grunt when development finishes? I know I could do it with kill -9...

ravishivt:
Rebased and added instructions to stop grunt-watch to the wiki.

Issue #1525, opened 2016-07-23
InfoWindow on mouseover should be disabled. At least for touch devices.


rnarian:
Right now InfoWindows open on mouseover. This was introduced with #544. Unfortunately this makes the map almost unusable on mobile. When zooming or dragging the map countless InfoWindows will open/close unexpectedly and the whole screen starts flickering.

![juli-23-2016 15-46-49](https://cloud.githubusercontent.com/assets/1174951/17078182/b809cd00-50ec-11e6-86c8-92b47332ffef.gif)

Since the map will most likely be used on a mobile device while hunting Pokémon—not on a desktop pc—this feature is unnecessary in most use cases.

Also when looking at a regular Google Maps map InfoWindows won't open on mouseover either, so this doesn't even seem right on desktop pcs, even though its kind of working.

frdmn:
:+1:

neobithia:
Disable it on mobile but don't get rid of it.  I use the map on a laptop while hunting.  My laptop is using wifi from my phone.

invisiblek:
https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1679

frdmn:
@invisiblek Awesome!

rnarian:
@invisiblek Thank you! :+1:

trickpattyFH20:
I am very against the idea behind PR #1679 . It completely disables the infoWindow popup for mobile devices.

Seeing the despawn countdown timer is one of the most important features of this application.

The main problem on mobile devices is that the screen jumps (autoPan) when you open an infoWindow so that the entire popup fits on screen.

I created a pull request #1811 to prevent the screen from jumping around when a popup is opened.

I really hope that #1679 does not get pulled, it would be detrimental to the mobile experience to completely disable the popups. @frdmn

invisiblek:
@trickpattyFH20 https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1679 has now been updated. It disables the popups on mouseover **during** touch. Simply tapping the pokemon still brings up the info window (it always has through that patch)

trickpattyFH20:
@invisiblek nice, had me worried there :)

MonkeyLeeT:
#1811 Seems to work better in a simpler way. Can you guys test it? (It's working for me and map is not paning around).

rnarian:
[...]

PR #1679, opened 2016-07-23
Disable mouseover effects on mobile browsers

invisiblek:
## Types of changes

## Checklist:

* insane regex string taken from: http://stackoverflow.com/a/11381730

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -419,17 +419,32 @@ function addListeners(marker) {
         marker.persist = null;
     });
 
-    marker.addListener('mouseover', function() {
+    function mouseover() {
         marker.infoWindow.open(map, marker);
         clearSelection();
         updateLabelDiffTime();
-    });
+    }
 
-    marker.addListener('mouseout', function() {
+    function mouseout() {
         if (!marker.persist) {
             marker.infoWindow.close();
         }
+    }
+
+    marker.addListener('mouseover', function() { mouseover() });
+    marker.addListener('mouseout', function() { mouseout() });
+
+
+    marker.addListener('touchstart', function() {
+        marker.removeListener('mouseover');
+        marker.removeListener('mouseout');
     });
+
+    marker.addListener('touchend', function() {
+        marker.addListener('mouseover', function() { mouseover() });
+        marker.addListener('mouseout', function() { mouseout() });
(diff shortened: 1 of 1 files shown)


BillyWM:
What problem does this solve? Which mobile browsers even fire mouseover events?

invisiblek:
Chrome does for sure, it's annoying when you're panning around

BillyWM:
Which OS and version? I'm not seeing it on Chrome on Android 6.

invisiblek:
Chrome dev on latest cm13 nightly

static/map.js
```diff
@@ -386,6 +386,13 @@ function clearSelection() {
...
+  return check;
+}
+
```
tegob: Please don’t use user-agent sniffing.

Since Modernizr is being used, why not just go with something like:

```js
if (Modernizr.touch) {
  // It’s a touch device.
}
```

You could also check if `window.orientation` exists (i.e. another indication it’s a “mobile” device).

frdmn:
👍

rnarian:
@BillyWM This happens also on mobile Safari, see #1525

BillyWM:
@rnarian: Thanks for that! I was wondering if this could be reproduced for other browsers or if it was just a regression in Chrome dev.

Detecting touch is still the wrong approach. There are touchscreen laptops. A lot of them, now. If you have one, then this disables mouseover. Annoying!

The issue should be addressed head on: we don't want mouseover while panning, so code to that.

1. On touchstart, set a panning = true flag
2. On touchend, panning = false
3. If panning, disable mouseover

rnarian:
In my opinion the mouseover thing should be removed completely, even on desktop pcs. As described in #1525 this is not something I expect when using a map. This is not happening when using Google Maps, so I don't see why this has to happen here.

BillyWM:
Adding a new handler to all markers that swaps other handlers on pan is a bad idea for performance. Again, it should be a single global flag that detects whether panning is in progress.

MonkeyLeeT:
#1811 Also fixed the panning issue, is this PR still needed?

invisiblek:
it would appear this is no longer needed, sweet!

PR #1694, opened 2016-07-24
Fixed retrieving and storing values in localStorage

Chlodochar:
## Description
There were a number of errors with retrieving values from localStorage where it was presumed the value from localStorage was a boolean, whereas localStorage only stores strings (booleans are automatically converted to strings).
For example (https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/static/map.js#L462)
```
if (!(localStorage.showPokemon === 'true')) {
```
Here `localStorage.showPokemon` is retrieved. If it was not set earlier, then null is returned. The default value of `showPokemon` is not used in this case (which should be `'true'`). So for some people, pokemon weren't loaded.

To circumvent this in the future I added `Store` which predefines which options are stored in localStorage, what their default values are and how to parse/stringify those values. I replaced raw usage of localStorage with usage of Store.

The options are predefined here: https://github.com/Chlodochar/PokemonGo-Map/commit/2b5a44872484c275b71b9e061d49c509d862c2d2#diff-a350469a6764edb653c7280bd0cf5adbR72

## Motivation and Context
To circumvent invalid usage of `localStorage` by making sure the type and the default values are predefined.

## How Has This Been Tested?
Run by all controls in webinterface.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -58,6 +58,137 @@ var pokemon_sprites = {
     }
 };
 
+
+//
+// LocalStorage helpers
+//
+
+var StoreTypes = {
+    Boolean: {
+        parse: function(str) {
+            switch(str.toLowerCase()) {
+                case '1':
+                case 'true':
+                case 'yes':
+                    return true;
+                default:
+                    return false;
+            }
+        },
+        stringify: function(b) {
+            return b ? 'true' : 'false';
+        }
+    },
+    JSON: {
+        parse: function(str) {
+            return JSON.parse(str);
+        },
+        stringify: function(json) {
+            return JSON.stringify(json);
+        }
+    },
+    String: {
+        parse: function(str) {
+            return str;
+        },
+        stringify: function(str) {
+            return str;
+        }
+    },
+    Number: {
+        parse: function(str) {
+            return parseInt(str, 10);
+        },
+        stringify: function(number) {
+            return number.toString();
+        }
+    }
+};
+
+var StoreOptions = {
+    map_style: {
+        default: 'roadmap',
+        type: StoreTypes.String
+    },
+    remember_select_exclude: {
+        default: [],
+        type: StoreTypes.JSON
+    },
+    remember_select_notify: {
+        default: [],
+        type: StoreTypes.JSON
+    },
+    showGyms: {
+        default: false,
+        type: StoreTypes.Boolean
+    },
+    showPokemon: {
+        default: true,
+        type: StoreTypes.Boolean
+    },
+    showLuredPokemon: {
+        default: true,
+        type: StoreTypes.Boolean
+    },
+    showPokestops: {
+        default: true,
+        type: StoreTypes.Boolean
+    },
+    showScanned: {
+        default: false,
+        type: StoreTypes.Boolean
+    },
+    playSound: {
+        default: false,
+        type: StoreTypes.Boolean
+    },
+    geoLocate: {
+        default: false,
+        type: StoreTypes.Boolean
+    },
+    playSound: {
+        default: false,
+        type: StoreTypes.Boolean
+    },
+    pokemonIcons: {
+        default: 'highres',
+        type: StoreTypes.String
+    },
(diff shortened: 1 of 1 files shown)


marauder37:
Someone from frontend will need to ok this, it's too big for me

Chlodochar:
@marauder37 who should i ask?

Chlodochar:
@JZ6 Could you take a look?

BillyWM:
Not a "collaborator" but I've written localStorage features in production and this looks solid to me. I think an abstraction layer to bring some more discipline to the code is helpful.

There was that issue for example where bad localStorage code wrote the string "undefined" to a value - https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1162 - and broke things for anyone actively pulling develop, so I think this solves a real problem by enforcing better checks.

MonkeyLeeT:
#1694 #1725 @TomProkop  @Chlodochar  Please take a look and have an agreement.

And rebase.

Chlodochar:
Rebased, I hope I interpreted the changes of #1743 correctly, because that also (incorrectly) made use of `localStorage`.

Chlodochar:
@MonkeyLeeT Who has merge permission that can interpret whether this (or #1725) is worthy of merging?

Since new features come and go I keep needing to rebase. I've done that for a few PRs, but it takes a lot of time (8 non-trivial rebases over 2-3 days for a single PR). Especially a change like this that touches multiple features in the client. Especially when changes like #1816 are coming (reformatting of code).

Chlodochar:
@medghaim Could you have a look?

static/map.js
```diff
@@ -33,6 +33,112 @@ var map_scanned = {}
...
+            throw "Store key was not defined " + key;
+        }
+        return option;
+    },
+    get: function(key) {
+        var option = this.getOption(key);
+        var optionType = option.type || StoreTypes.String;
+        var rawValue = localStorage[key];
+        if (rawValue === null || rawValue === undefined) {
+            return option.default;
```
chuyskywalker: There is no check here to see if the `default` is defined.

static/map.js
```diff
@@ -33,6 +33,112 @@ var map_scanned = {}
...
+        type: StoreTypes.Boolean
+    },
+    playSound: {
+        default: false,
+        type: StoreTypes.Boolean
+    }
+};
+
+var Store = {
+    getOption: function(key) {
```
chuyskywalker: It's weird that this makes things very prescriptive and well defined, but then goes an assumes a `string` datatype when left undefined.

chuyskywalker:
Overall this is a big +1 from me -- a few small feedback items. 

_(Also this is one of those really clear points where you can see the spaghetti-ification of not having something like webpack/require/etc)_

static/map.js
Chlodochar: If there is no default, it'll be undefined. Should we add runtime validation on static object `StoreOptions`?

static/map.js
Chlodochar: Fixed.

Chlodochar:
Fully agree on using something like Webpack. Also would be proponent of some component framework (react, vue, anything).

For now though, the changes are rebased. Ready for merge.

chuyskywalker:
tested by clearing local storage and setting things, unsetting, reloading, etc from the front end :+1: from me

Chlodochar:
In that case I'll merge this one.

Chlodochar:
@chuyskywalker Thanks for giving this a go.

ximex:
wouldn't it be better/easier to store JSON objects?
At saving only `JSON.stringify()` them and onload only `JSON.parse()` them

Chlodochar:
Yes, you can set all types to JSON. However, that way the type is still implicit: you intent to store `0`, but are actually storing `null`. Same thing for retrieving. It's better to explicitly type the values, so that they are always parsed and stringified the same way.
In addition, dates and other 'complex' types are now possible as well ;)

PR #1283, opened 2016-07-22
Mobile styling

parkershepherd:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

The following changes were made

- Clean up styling on mobile endpoint
- Add refresh link for easier reloading
- Set viewport width for mobile devices

previous
new

If this is related to an existing ticket, include a link to it as well.

PR #1696, opened 2016-07-24
Accessibility Page

chuyskywalker:
## Description
Following up on #1283, I worked on the current "low bandwidth" page but didn't find a really nice way to make rows as accessible as a plain table. So I've added an additional page instead.

## Motivation and Context
Creates a way for users with reduced vision to use application to some degree

## How Has This Been Tested?
Load page at `/acc`

## Types of changes


JZ6:
interesting, instead of merging into main release, perhaps you can create a seprate branch for people with vision problems?

MonkeyLeeT:
An separate branch will be ideal.

chuyskywalker:
I've a better idea/pr

chuyskywalker:
#1721 isn't as huge, doesn't induce an extra page, and is easy to maintain.

PR #1699, opened 2016-07-24
Revert "update get_cellid to use s2cell coverage instead of linear path"

JZ6:
This should fix the low pokemon problem

### DIFF
--- a/pogom/pgoapi/utilities.py
+++ b/pogom/pgoapi/utilities.py
@@ -27,7 +27,7 @@ import struct
 import re
 
 from importlib import import_module
-from s2sphere import LatLng, Cell, RegionCoverer, Cap
+from s2sphere import CellId, LatLng
 from google.protobuf.internal import encoder
 from geopy.geocoders import GoogleV3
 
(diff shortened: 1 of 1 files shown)


Chlodochar:
Why is this incorrect?

medghaim:
+1. Let's leave PoGoAPI alone. It's dangerous to build functionality into pogoapi, then build off that. Updating pogoapi would be a nightmare in the future.

MonkeyLeeT:
+1. Better not to touch PoGoAPI.

JZ6:
lets go

andrewda:
Related: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/671 - will prevent more changes like this unless they're done upstream.

Chlodochar:
Can one of you please take a look at #671. I'm trying to keep it alive, because I feel it is necessary to be merged, but I have been forced to rebase it around 8 times (with conflicts each time) because of PRs like this one. It would be really helpful to get some feedback and eventually be merged.

PR #1700, opened 2016-07-24
--only-server mode added to complement --no-server

DeirhX:
Since we already have isolated shared data into database and we have --no-server option just for the searcher, let's complement it with --only-server as well so anyone can easily start a server in one process and N searchers as other processes.

## Description
Very simple addition of new command-line argument and forfeiting instantiation of searcher threads

## Motivation and Context
1) To enable scaling the solution by allowing having multiple workers share the server easily
2) Server in separate process can be more easily set to higher priority than workers 

## How Has This Been Tested?
Ran --no-server and --only-server in separate processes, no issues encountered, database is shared 

## Types of changes

## Checklist:


MonkeyLeeT:
Tested and working across two process in different mode. Nice addition.

andrewda:
:+1: tested and works well on OS X

jangrewe:
Ah, awesome, i was looking forward to that! Thank you.

NenadJAustralia:
Is this for multiple instances?

So one process is the main and the others just contribute to the DB?

jangrewe:
@NenadJAustralia yup

Issue #1573, opened 2016-07-23
Feature request: remove gmaps labels


Skullmonkey:
Add a toggle to remove the street names and other lables on GMaps. I know my city very well and they are a distraction to me.

joveice:
Could be cool if possible

TheReaperr:
Open up static/map.js  in function initMap() theres styles:[...]. There you can set up how your google maps looks like. Find the entry for the labels.

I didn't really like the default look so I looked up a good one on snazzymaps.com ( I'm using https://snazzymaps.com/style/52112/dark-grey-w-contrast)

Skullmonkey:
This is all fine and dandy short time, but this file constantly gets updated and I would have to go in and manually edit the file each time it gets updated. A toggle would be nice so it stays in place.

PR #1704, opened 2016-07-24
Revert "Added 3 custom map styles that remove labels ( #1573 )"

JZ6:
need street labels


MonkeyLeeT:
+1. Guess it could be a standalone toggle if street info is really distracting.

Skullmonkey:
The original maps were still in there for the labels. I just added extra options to the dropdown that had them removed for those that wanted them.

JZ6:
a toggle would be better

Skullmonkey:
I have no clue how to do that.

marauder37:
Would one additional map type be enough for this?

Skullmonkey:
It's better than nothing I suppose. I just thought when making this that if I did 1, I would do the 3 custom ones since I couldn't find a way to also do the street view or satellite view. Also, I was thinking that some people use the other ones that would like to try it out without the labels.

I personally think it looks better and it is easier for me to find Pokemon. I also know my area very well. Other like the street names. To each their own. It at least gave the mass population a choice.

As far as making it a toggle, I don't know how to do that or even where to start. I have very little coding knowledge and got by with what I could do just my dissecting the current code.

marauder37:
It can't be done with a toggle, StyledMapType is immutable

jjbott:
Heres a proof of concept for a switch: https://github.com/jjbott/PokemonGo-Map/commit/2e1ed6f8407807f0663f04727259bc134094c2c4

The switch should probably be in a map options panel somewhere, but it works

PR #1743, opened 2016-07-24
Various lure improvements

hgentry:
## Description

Some of the last few changes have broken some of the lure functionality. This repairs those issues. Namely, a Lure Pokemon toggle was added to provide more granular control over data and this setting is used to ensure that other settings don't break Lure Pokemon.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -548,20 +548,19 @@ function updateMap() {
                 return false;
             }
             var item2 = {pokestop_id: item.pokestop_id, lure_expiration: item.lure_expiration, pokemon_id: item.active_pokemon_id, latitude: item.latitude+ 0.00005, longitude: item.longitude + 0.00005, pokemon_name: idToPokemon[item.active_pokemon_id], disappear_time: item.lure_expiration}
-            if(map_lure_pokemons[item2.pokestop_id] == null  && !!item2.lure_expiration) {
+            if(map_lure_pokemons[item2.pokestop_id] == null  && item2.lure_expiration) {
             	//if (item.marker) item.marker.setMap(null);
                 item2.marker = setupPokemonMarker(item2);
   		map_lure_pokemons[item2.pokestop_id] = item2;
 
   		}
-            if(map_lure_pokemons[item.pokestop_id] != null  && (!!item2.lure_expiration != !!map_lure_pokemons[item.pokestop_id].lure_expiration || item2.active_pokemon_id != map_lure_pokemons[item2.pokestop_id].active_pokemon_id)) {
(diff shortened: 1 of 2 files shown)


MonkeyLeeT:
#1745 ?

hgentry:
I'll make some changes so we don't conflict

MonkeyLeeT:
Or better if you guys decide on one to merge :)

hgentry:
We're doing separate things with these. I'm adding a toggle to fix something that broke, they're fixing something else entirely. I'm fairly confident that with this last change we won't collide, except for possibly a single line. Not sure how forgiving git is.

andrewda:
@hgentry It's not forgiving at all, heh. Which line are you talking about? I can add it to my PR.

hgentry:
@MonkeyLeeT We've got our differences sorted out. If mine goes first, we're good to go.

MonkeyLeeT:
OK, tested in new york, working well. +1.

Chlodochar:
This PR is also uses an incorrect method to set a default value for `localStorage`. Please note that `localStorage.anything` always returns `undefined` or a string. It never returns a boolean. Using `localStorage.showPokemon || true` is therefore incorrect. Any `if (localStorage.showPokemon)` is also incorrect.

See #1694.

BillyWM:
>Using localStorage.showPokemon || true is therefore incorrect. 

In Javascript, the || statement has fallthrough. This is a common practice to set a default value if the first thing has a "falsy" value (0, undefined, null, or the empty string "")

`x = localStorage.showPokemon || true`

where the showPokemon property is undefined this falls through and x gets "true"

--------------------------

Open your dev console (F12) and try this:

`obj = {}
x = obj.showPokemon || "I fell through"`

x gets "I fell through"

Chlodochar:
@BillyWM Thanks for highlighting my point. The problem is indeed hard to spot. If `localStorage.showPokemon` is set to false, it is actually `'false'` (as string). So depending on whether it is set, it'll result in a string `'false'` or a boolean value `false`. The string `'false'` is actually a truthful value, so any `if` or `||` or `&&` that is used with this value will interpret this value as `true`.

Only the `localStorage.showPokespots === 'true'` were somewhat correct, but only if the default value of `localStorage.showPokestops` is false. It is incorrect for `showPokemons` because its default value seems to be `true`. So, due to this error, in some cases `showPokemon` is intepreted as being true and sometimes as being false when the option was not set.

BillyWM:
@Chlodochar: Nah, I was mistakenly only thinking about the case of undefined. I deleted my post for a reason, thanks.

PR #1755, opened 2016-07-24
Implemented logic to change position of the marker and trigger search by double click (map won't zoom)

yang6lang1:
Implemented logic to change position of the marker and trigger search by double click (map won't zoom)

## Description

- Implemented logic to change position of the marker (also re-search) by double click (map won't zoom)
- Implemented the ability to re-center the marker (also re-search) by clicking the myLocationButton
- Refactored code for readability

## Motivation and Context

When there's many un-filtered pokemons on the map, it is hard to drag the marker (especially when accessing using a phone). Double click to change the marker position and triggers a search is more user-friendly. The zoom capability is still available from pinch gesture (or mouse scroll).

## How Has This Been Tested?

Tested on localhost desktop. Also deployed to heroku and tested using phone.

Now that clicking on the myLocationButton will bring marker to the center of the map as well so the myLocationButton itself seems redundant. This is a change of behaviour and I think it's fair to let all other contributors to decide whether this is a more friendly behaviour. :)

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -15,7 +15,7 @@ var notifiedPokemon = [];
 
 var map;
 var locationMarker;
-var marker;
+var searchMarker;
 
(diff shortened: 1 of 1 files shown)


yang6lang1:
I noticed #1808 is providing similar functionalities.

PR #1758, opened 2016-07-24
.gitignore added OSX and Windows System ignores. Updated Python ignores

Darkade:
Pretty much the title.
Added system ignores for:

- OS X
- Windows

Updated Python ignores

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,152 @@
-*.pyc
 *.idea
 *.db
 *.db-journal
 *.sass-cache
 *.orig
-config/config.ini
-config/credentials.json
+config.ini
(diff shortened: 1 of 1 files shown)


MonkeyLeeT:
Do you believe that our current gitignore is good enough?

Darkade:
Goodish? I keep deleting hidden files before commiting. This is not critical, though  :3

MonkeyLeeT:
Well ideally we should just add stuff that is really needed. You're welcome to include new ignores as you find them useful (or to ignore them in your .git/info/exclude which is local to your repo).

Just my two cents :3

PR #1765, opened 2016-07-24
create notifier.py

JawnZ:
Creating a slack notifier

## Description
Reads from the mysql database populated by the app, and sends a slack notification if the pokemon is found. has some logic to not show expired findings.

This could easily be expanded into an email notification, and SMS (though I would recommend using the email version of an SMS number for simplicity)

## Motivation and Context
it would be nice to get notifications other than just on your browser, so I wrote a thing.

It's not integrated into the  rest of the app at all, it should be able to pull the mysql stuff from the config.ini as well as use PyMySQL but I didn't have time for all that.

requirements is pip install slacker (as well as the mysql stuff, but it should be migrated to be in line with what the rest of the application uses).

### DIFF
--- a/notifier.py
+++ b/notifier.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python
+from slacker import Slacker
+import MySQLdb
+
+slack = Slacker('SLACK-API-TOKEN-GOES-HERE')
+
+#what Pokemon are you wanting to be alerted about?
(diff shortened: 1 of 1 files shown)


JZ6:
+1 good stuff

JawnZ:
@JZ6 I clearly herped when I should've derped.

Thanks!

PR #1767, opened 2016-07-24
Add vanilla map (no GUI)

andrewda:
## Description

This is a pretty significant addition, but has no breaking changes to anything existing. This adds `/vanilla` that does not contain a GUI (i.e. it's just a map) to make it easier for people attempting to port this app into their own projects. A great example of this would be [mchristopher/PokemonGo-DesktopMap](https://github.com/mchristopher/PokemonGo-DesktopMap), which is currently using an outdated version of this repository for exactly this reason. This will not affect any other users besides those using/creating ports.

## Motivation and Context

As mentioned above, this will make it much easier for developers to make custom ports of PokemonGo-Map to implement into their own projects. 

## How Has This Been Tested?

This has been tested on Chrome/Firefox on Mac/Linux/Windows. There should be no compatibility issues, as this is almost identical to the regular project, minus the GUI.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -46,15 +46,33 @@ class Pogom(Flask):
     def vanillamap(self):
         args = get_args()
         display = "inline"
+        exclude = []
+        notify = []
+
         if args.fixed_location:
             display = "none"
+
+        if args.exclude_pokemon:
+            exclude = args.exclude_pokemon
+
+        if args.notify_pokemon:
+            notify = args.notify_pokemon
+
+        print(notify)
 
         return render_template('map_vanilla.html',
                                lat=config['ORIGINAL_LATITUDE'],
                                lng=config['ORIGINAL_LONGITUDE'],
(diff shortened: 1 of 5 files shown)


JZ6:
need testing

chuyskywalker:
This duplicates a LOT of code, especially `map.js` which would start making keeping it and the non-vanilla one up-to-date quite difficult. I'd _almost_ rather we set this up in such a fashion that the PokemonGo-Map UI was layered on top of the map, and there was a, perhaps, `get` url param to disable rendering the PGM ui. That way as the pokemon end points update and evolve, the vanilla page won't get forgotten and broken.

Chlodochar:
I don't like this either. Any change that's being done on the gui will need to be considered to be merged here as well. However, to get this feature in properly, the GUI will need some modularity. Otherwise it'll get even messier.

AHAAAAAAA:
I agree with the previous 2 comments, putting this on hold and closing PR.

PR #1771, opened 2016-07-24
Fix exception on /mobile

relrod:
`Pokemon.get_active` now requires args; This fixes a fatal on `/mobile` caused by 5deb540dcac4f5c9600f7797caa30c3c1def3296.


chuyskywalker:
I've tested and know this is all good.

PR #1776, opened 2016-07-24
Autolocation and Login support

coolaj86:
I've added auto-location (using the browser geolocation apis), live location updating, and conditional login support.

The conditional login support is turned on when the config file has `requireLogin: true` (which is necessary for my node server) and otherwise is off.

I didn't try to clean up the existing code as my previous PR is still outstanding, so this is just more free-form functions to the mix.

I'd still really like to lint and clean up the existing code, but I don't want to get some feedback on these (and what style to adopt) first.

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -15,18 +15,22 @@ class Pogom(Flask):
     def __init__(self, import_name, **kwargs):
         super(Pogom, self).__init__(import_name, **kwargs)
         self.json_encoder = CustomJSONEncoder
-        self.route("/", methods=['GET'])(self.fullmap)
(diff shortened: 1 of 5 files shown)


Worthy7:
Login lets what happen exactly?

coolaj86:
My node pokemap server is completely stateless, which means that it can support an unlimited number of users at a time and the client controls the update rate, scanning, and caching - not limited by the server's database capabilities (mentioned as a problem in a few issues).

The config flag I added allows logging in to a server that supports multiple users. For this server there is no change for that aspect.

sp90:
I share this with my friends, and a login would be awesome 👍  💯

Issue #1153, opened 2016-07-22
Use SVG instead of PNG


Whoknowsit:
Currently, your map uses PNG graphics for monsters, which is okay. But on bigger screens, they look blurry and in some cases, you cannot determine what's being displayed until you zoom in.

So i would recommend to use SVG graphics instead of PNGs. I don't think, that they need 1:1 the same as used in Go, you just need to be able to see, what's displayed.

maximal:
Where can we get SVG graphics for all the pokémons?

Whoknowsit:
I'll post them later this day.

jnovack:
http://veekun.com/dex/media/pokemon/dream-world/

jnovack:
https://github.com/jnovack/pokemon-svg

parkershepherd:
@Whoknowsit I ran into trouble trying to generate an svg sprite, but if anyone is able to, it would be easy to add svg as an option. As it is, all the svg icons total about 4.9MB, which isn't great for mobile use though

Whoknowsit:
But this would require that you load all of them at once.

Resize the originals to... Dunno... 24px could be an option. For more speed, an option to use remotely hosted assets could be provided. The remote host is Github and assets are included via Github pages.

I would not recommend the sprite way. 

PS: To avoid SVGS,  you could also use PNGS which are 4x as big as they are displayed later on the map. So if an image is 24x24px, create a 96x96px version and scale it down using CSS. Using media queries allows you to do so on 4k/retina displays. 

If none of these ideas is good, the last alternative would be the sprite. But as already said, i would avoid that.

parkershepherd:
I just went the sprite route because a previous PR was approved switching over to them (and sprites will generally load faster, at least until http/2 is more prevalent)

parkershepherd:
If someone could find a way to resize those svgs to be the same size, I could generate an even larger png, or even a raw svg version, and add it as an option though

Whoknowsit:
I'll see what i can do.

jnovack:
The SVG should be used as the source for the PNGs.

Bulk-export the SVGs to your desired size. That's the point of SVGs is that they look good at every size, that's how they render.

Whoknowsit:
That's what i meant.

PR #1781, opened 2016-07-24
Added retina-capable pokemon icons-sprite image and code

enket:
## Description
1. Added icons-sprite@2x.png generated from http://www.pokemon.com/us/pokedex/ to support high-res display.
2. Added centering parameter to be locate center.

## Motivation and Context
1. In most smartphone, the icons are blurred.
By adding @2x image, icons became more clear.
2. Because of anchor is set center of bottom of icon default, when you zoom-in you confuse the exact location of pokemon.
By adding centering parameter, icons are fixed on map even you zoom-in or out.

## How Has This Been Tested?
On my local computer.
There is no affect to other code.

## Types of changes

## Checklist:


tegob:
The icons in `icons-sprite.png` are different from the ones in `icons-sprite@2x.png`. Having them be the same would be preferable, but in the mean time 👍 to merging this.

Does anyone know where the icons in `icons-sprite.png` come from?

123456abcdef:
Is it possible to achieve the same effect using svg images like mentioned [here](https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1153)? User could also choose their own scale for the icons on the map without decreasing quality.

tegob:
@123456abcdef Ah yes, I didn’t know SVG icons were even available! That would be an even better solution, since then the icons would be the same regardless of screen resolution / DPI.

chuyskywalker:
Ohh, these looks pretty bad on the dark background setup -- the PNG transparency isn't right:

2016-07-24 13_29_09-pokemon go map
2016-07-24 13_28_55-pokemon go map

_These screenshots are from chrome, but safari on ios shows the same problem._

chuyskywalker:
The SVG option works pretty nicely: https://github.com/chuyskywalker/PokemonGo-Map/commit/16c7f0f181c5a368d04a8570516e7f7072a1e892

parkershepherd:
@enket any interest in adding these icons as an option here? https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1886

enket:
SVG is good solution for high-res displays, but its size is too large.
If you use SVG icons from https://github.com/jnovack/pokemon-svg, it become about 4MB large with No.1-151.
I think it is too large for mobile user.
In my opinion, @2x.png is best solution, considering size.

@parkershepherd 
Thank you.
But icons in #1886 seems to comes from https://github.com/jnovack/pokemon-svg and it is better than mine because its shape is more clear and has no transparency probrem.
So icons on High-res display is solved so far.

If there is no objection, I'll close this pull request.

invisiblek:
yea, i would say this could be closed

PR #1793, opened 2016-07-24
Limiting Search_Loop

iamhssingh:
## Description

- [x] I have added a limit on infinite loop. If user specifies a limit via -lim option while passing argument, the search_loop will run that many times only. In any other condition, it will be an infinite one!

- [x] I have added a route of shutdown. Users can shutdown server remotely by visiting **host:port/shutdown**

## Motivation and Context

It helps in many cases, especially in case of server. It consumes less resource.

## How Has This Been Tested?

I have tested changes on my server.

I ran on debian localhost only as of now. Soon I'll test this on my Amazon Cloud Hosting

screenshot from 2016-07-24 15 09 55
screenshot from 2016-07-24 15 08 21

## Types of changes

## Checklist:

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -17,9 +17,19 @@ class Pogom(Flask):
         self.json_encoder = CustomJSONEncoder
         self.route("/", methods=['GET'])(self.fullmap)
         self.route("/raw_data", methods=['GET'])(self.raw_data)
+        self.route("/shutdown", methods=['GET'])(self.shutdown)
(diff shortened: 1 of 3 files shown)


medghaim:
I like the `limit` feature, but not sure I'm a fan of `/shutdown`. A lot of people host this for public use. Maybe others can chime in here? Pretty big design decision.

iamhssingh:
Yeah. That's a pretty serious issue. Let me work on it!

PR #1802, opened 2016-07-24
Add ability to run server through https

mik9:
This pull request includes a

- [ ] Bug fix
- [X] New feature
- [ ] Translation

- Added configuration to allow to run server through https.

As Google Chrome (desktop and mobile) disallow using location data from non-secure websites I added ability to run server as https.

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -93,10 +93,24 @@ def get_args():
     parser.add_argument('-np', '--no-pokemon', help='Disables Pokemon from the map (including parsing them into local db)', action='store_true', default=False)
     parser.add_argument('-ng', '--no-gyms', help='Disables Gyms from the map (including parsing them into local db)', action='store_true', default=False)
     parser.add_argument('-nk', '--no-pokestops', help='Disables PokeStops from the map (including parsing them into local db)', action='store_true', default=False)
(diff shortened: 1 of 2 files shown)


alvinycheung:
I've tested this and can confirm that it works. I used a free service, https://certbot.eff.org/, to get my SSL cert set up and it works great. I can now use location services with my map :D

THANK YOU SO MUCH FOR THIS COMMIT!!!

here is my run command
cd /var/www/PokemonGo-Map/; python runserver.py -a ptc -u [USER] -p [PASS] -l "LOCATION" -st 10 -k [GMAPS_KEY] -H [HOST.URL] -P 443 -hc /etc/letsencrypt/live/[HOST.URL]/cert.pem -hpk /etc/letsencrypt/live/[HOST.URL]/privkey.pem

chuyskywalker:
This PR is pretty important since a lot of the location mobile stuff won't work over non-ssl connections, it seems. I'm going to check this it out and give it a go as well.

chuyskywalker:
@mik9 thanks for this; I'm going to close your PR in favor of the revitalized #1936, which also handles SSL config in the `config.ini` file. I did grab some good ideas you had from here though, thanks!

PR #1808, opened 2016-07-24
Added double-click to change location.

medghaim:
## Description
 - Added a double click to change search location, faster than dragging and dropping, but also works along side it.

## Motivation and Context
- When making a big change in location, zooming in and out and dragging the marker can be annoying. This is a simple solution.

## How Has This Been Tested?
 - Lots of double clicking. Lots. 

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -64,8 +64,8 @@ function initMap() {
         disableDoubleClickZoom: true,
         fullscreenControl: true,
         streetViewControl: false,
-		mapTypeControl: true,
-		mapTypeControlOptions: {
+        mapTypeControl: true,
+        mapTypeControlOptions: {
           style: google.maps.MapTypeControlStyle.DROPDOWN_MENU,
           position: google.maps.ControlPosition.RIGHT_TOP,
           mapTypeIds: [
--- a/static/map.js
+++ b/static/map.js
@@ -398,6 +398,7 @@ function setupScannedMarker(item) {
     var marker = new google.maps.Circle({
         map: map,
         center: circleCenter,
+        clickable: false,
         radius: 100,    // 10 miles in metres
         fillColor: getColorByDate(item.last_modified),
         strokeWeight: 1
--- a/static/map.js
+++ b/static/map.js
@@ -61,6 +61,7 @@ function initMap() {
             lng: center_lng
         },
         zoom: 16,
+        disableDoubleClickZoom: true,
         fullscreenControl: true,
         streetViewControl: false,
 		mapTypeControl: true,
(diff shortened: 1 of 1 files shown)


ngg:
Is it better than https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1755 ?

JamesWP:
tested on mac osx and i can't double click to change location? it just zooms when i double click

medghaim:
Hmm, works on windows and ubuntu for me. You cleared cache/start up in a fresh incognito mode browser?

JamesWP:
my apologies, it does work, however it prevents the default double click to zoom behavior

medghaim:
That is intentional and can be changed, but I personally found it a little weird to both have double click to change loc _and_ zoom.

Xomux:
It does not work. I cleared cache and i opened it in a incognito mode window. It's just zoom-in and zoom-out with the right click :S

medghaim:
Did you

```
git checkout -b medghaim-develop develop
git pull https://github.com/medghaim/PokemonGo-Map.git develop
```

Because regardless, zoom should 100% be disabled.

Chlodochar:
This does disable double click to zoom, which is useful for some people. Changing the position is already possible by dragging the marker or typing a new location in the sidebar. I think double-click to change position should be optional.

medghaim:
I dunno, I personally feel the subset of people who use click to zoom (scroll wheel/pinch!) are dwarfed by those who want double click to change location quickly without having to turn off all filters just so you can grab your marker to drag. 

I also don't agree that something as trivial as this should go on the command line, it's either in or out.

yang6lang1:
I would think double click is still available from mouse scroll and pinch(on phone). Perhaps we can create a toggle so that for certain users that don't like the marker, they can toggle it off and restore the double click zoom behaviour. This is purely depend on the usage of this map. :)

Chlodochar:
@macgngsta Not a command-line option. It has nothing to do with the server, rather a client-side option. I use double-tap to zoom a lot on the phone as the buttons are often too small.

medghaim:
Yeah I get that it has nothing to do with the server, that's why I was so against it being a command line option lol. Still though, don't like the idea of cluttering up the UI with trivial stuff like that either, that's just my 2 cents.

Chlodochar:
Ah, alright. Well, just giving a heads up that there are people using double-click zoom. Whatever the decision will be is fine ;)

joshslark:
Tested this and found that the marker will not move if you double-click over a scanned location bubble.

medghaim:
Good catch!

medghaim:
That fixes things!

tkSimon:
#1912 is a simple alternative, using a link in the header.

PR #1816, opened 2016-07-24
cleanup map.js [DRY]

broglep:
Don't repeat yourself, refactored a lot of copy&pasted code

## Description
- add change listener factory for switches
- remove duplicate switch initialization
- extract API request
- extract processing of API response

## Motivation and Context
It makes the code cleaner and prepares for some other changes allowing to receive data via other then XHR

## How Has This Been Tested?
Run server with new code, scanned our test area, and verified that toggling all switches properly work.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1817, opened 2016-07-24
prevent multiple map-updates

destiny117:
In case a map-update took longer than 5s, multiple updates were started, this is a simple fix so there is always only one map-update running

## Description

a temporary variable window.mapUpdating is set and checked during map-updates

## Motivation and Context

my server is pretty slow, and with growing databases map updates were often taking a while... this fixes it

## Types of changes

## Checklist:


Chlodochar:
:+1: Looks good. Though I'd define the variable right before `updateMap` and not refer to `window` at all.

Using `setTimeout` (after each update) instead of `setInterval` might be another alternative.

destiny117:
Thanks for the feedback

There are all kinds of events calling a map update, so i'd prefer a variable instead of the `setTimeout` method.

It might postpone some UI updates on these events on slower servers for the benefit of not hammering the server with requests, but should not have a noticable impact on very responsive backends.

I could make a new pull request if a script-wide variable is preferred with your coding-guidelines, for now this works for me :)

chuyskywalker:
#1816 got merged in first which has a number of changes that might make this concept here a bit cleaner  (namely that the load_data function, which is what you're trying to have not step over itself) got put into it's own routine. So getting that on "lock down" should be even easier.

destiny117:
i rebased and considered your suggestion, there is now a variable `rawDataIsLoading` which is considered within the ajax object returned by `loadRawData()`

static/map.js
```diff
@@ -508,7 +509,17 @@ function loadRawData() {
...
+		beforeSend: function() {
+		    if (rawDataIsLoading) {
+		        return false;
+		    } else {
+		        rawDataIsLoading = true;
+		    }
+		},
+		complete: function() {
+		    rawDataIsLoading = false;
+		}
```
Chlodochar: The indentation needs to be fixed. This is using tabs, whereas the surrounding code uses spaces.

Chlodochar:
Other than the indentation, this looks good :+1: Worked correctly with the quick test I did.

static/map.js
destiny117: already fixed in newest commit

destiny117:
thanks, indentation is fixed already :)

Chlodochar:
@destiny117 That was fast. :+1: Looks good to me.

static/map.js
```diff
@@ -508,7 +509,17 @@ function loadRawData() {
...
         },
-        dataType: "json"
+        dataType: "json",
+        beforeSend: function() {
+            if (rawDataIsLoading) {
+                return false;
+            } else {
+                rawDataIsLoading = true;
+            }
+        },
```
chuyskywalker: According to [this page](http://stackoverflow.com/questions/10507079/stop-ajax-on-beforesend) you can't just return false here and have the request not be sent.

chuyskywalker: Oops! Read official docs:

```
beforeSend
Type: Function( jqXHR jqXHR, PlainObject settings )
A pre-request callback function that can be used to modify the jqXHR (in jQuery 1.4.x, XMLHTTPRequest) object before it is sent. Use this to set custom headers, etc. The jqXHR and settings objects are passed as arguments. This is an Ajax Event. Returning false in the beforeSend function will cancel the request. 
```

Should work as expected.

chuyskywalker:
Works great. Tested it by turning `rawDataIsLoading` true and false in console and updates stopped.

Could even be a future feature toggle in the options sidebar "Auto Refresh", perhaps ;)

Issue #1735, opened 2016-07-24
MySQL max_connections


jmbeach:
When using a MySQL database with limited connections, connections are left open even when thread count is set to 1, causing a `max_connections` error when the connection limit is exceeded (in my case, 10 connections).

## Expected Behavior
I would expect that if I limit the threads to 1, then only one mysql connection would be opened at a time

## Current Behavior
All connections seem to be left open indefinitely

## Possible Solution
I looked into using the PooledMySQLDatabase from the peewee library instead of the regular mysql class because it has a `max_connections` option. However, even when I set it to one, with only one thread, the app gives me a max_connections error. I am only editing the `models.py` file. I do not really know the whole structure of the app so I'm not sure if connections are also made elsewhere.

## Steps to Reproduce (for bugs)
1. Setup a free heroku mysql cleardb 
2. Edit / create a config.ini file to have the database type "mysql" and give it the credentials
2. `heroku logs` shows many frequent max_connections errors 

## Context
I'm trying to use a free cleardb MySQL database with the app

## Your Environment
Publishing to heroku, but same error locally with same settings

PR #1833, opened 2016-07-24
close db connection after upsert

jmbeach:
## Description
See issue #1735 which I opened because I kept getting an error about max_connections being exceeded on my remote mysql database. I added a call to close the db connection in the model class and this seems to take care of it. I suppose Peewee is good about opening connections but not necessarily closing them.

## Motivation and Context
It allows me to use a heroku cleardb mysql database with my heroku deployed version of the app. I plan on using the database to create a bot to alert of rarer, nearby pokemon.

## How Has This Been Tested?
I tested my change by running the runserver.py as normal. I have set my mysql database connection up in the config.ini. I monitored the open connections on the database's web-interface and saw that now only one connection was ever open instead of 10+

image

## Types of changes
Only the db.close insertion

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -302,6 +302,8 @@ def bulk_upsert(cls, data):
         except OperationalError as e:
             log.warning("%s... Retrying", e)
             continue
+        finally:
+            db.close()
 
         i+=step
 


Chlodochar:
The Peewee documentation suggests to connect and close the connection explicitly for every request: http://docs.peewee-orm.com/en/latest/peewee/database.html#creating-a-database-connection-and-tables

I guess the same goes for every 'search`-iteration.

It seems if you close the connection after every bulk_upsert, then it's opening and closing the connection 4 times (for Pokemon, for Pokestops, for Gyms, for ScanLocations).

jangrewe:
I'd rather see persistent connections (haven't checked the code, but it looks like we don't use it)

http://charlesleifer.com/blog/managing-database-connections-with-peewee/

chuyskywalker:
I agree that persistent connections would be a much better option -- especially since this is a long lived python application, why tear down a perfectly useful connection thread and build it back up over-and-over?

Even barring that, it seems like (http://docs.peewee-orm.com/en/latest/peewee/database.html#flask) would be the proper way to handle this. The search thread has it's own connection, which it should be using near constantly, so I wouldn't expect it to be the source of the connection build up, but instead the application (read: flask) causing it.

MonkeyLeeT:
Would you mind taking suggestions above and make persistent connection changes instead of simple close()?

You're welcome to close this if you need more time to design it :)

jmbeach:
@MonkeyLeeT sure. I'll look into that later today unless you've already taken care of it. 
I might have some trouble figuring out how to wire up the flask events though.
Also, later I realized that my call to close the connection only fixes the problem for the requests made from scanning for pokemon. Each time a user goes to the dashboard, queries to the database are made which open up more connections so either way something more needs to be done.

MonkeyLeeT:
Thanks a lot. Closing this one to prepare for 2.1.0 release. You're welcome to send in a new one tomorrow or later.

PR #1835, opened 2016-07-24
Producer/consumer pattern for search threads instead of task batching (speedup)

DeirhX:
Current implementation uses optional batching of up to --threads requests yet it always waits for the entire batch to finish which equals to waiting for the slowest thread of the batch before launching new batch.  This seriously impacts scalability since the entire node does nothing when waiting for prohibitively slow responses.

## Description
Rewritten the pattern to async producer/consumer queue of configurable SEARCH_QUEUE_DEPTH so all threads can be kept busy even if requests of uneven response speed are encountered. This should speed search times considerably when using a lot of threads (tested with 25 and up to 100).

Also **added retrying** in case server responds with invalid response which can start happening in very high throughput scenarios (previously thread gave up upon encountering log.error('Scan step {:d} failed. Response dictionary key error.'.format(step), creating gaps)

This should bring multiple workers + one server scenario a step closer to reality.

## Motivation and Context
Scalability of current solution is poor as adding more search threads can start slowing the entire process due to current wait_for_join(slowest_thread_of_the_batch)

## How Has This Been Tested?
Tested thoroughly with --threads 25 and seen a dramatic rise in search performance - test for yourself and see.
Note: --threads 100 often causes niantic servers to stop responding properly if on a fast connection - we need to implement adaptive delays for such low-latency scenarios.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/__init__.py
+++ b/pogom/__init__.py
@@ -11,5 +11,6 @@ config = {
     'REQ_SLEEP': 1,
     'REQ_HEAVY_SLEEP': 30,
     'REQ_MAX_FAILED': 5,
-    'PASSWORD': None
+    'PASSWORD': None,
+    'SEARCH_QUEUE_DEPTH': 100
 }
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -6,6 +6,7 @@ import time
 import math
 
 from threading import Thread, Lock
+from queue import Queue
 
 from pgoapi import PGoApi
 from pgoapi.utilities import f2i, get_cellid
@@ -26,6 +27,8 @@ lng_gap_meters = 86.6
 #111111m is approx 1 degree Lat, which is close enough for this
 meters_per_degree = 111111
 lat_gap_degrees = float(lat_gap_meters) / meters_per_degree
+
+search_queue = Queue(config['SEARCH_QUEUE_DEPTH'])
 
 def calculate_lng_degrees(lat):
     return float(lng_gap_meters) / (meters_per_degree * math.cos(math.radians(lat)))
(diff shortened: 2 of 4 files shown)


medghaim:
:O

Testing!

AHAAAAAAA:
Testing, really cool stuff! Good job. I was able to cover a huge area in minutes. Problems I've noticed so far: 
* process refuses to terminate with so many threads.
* db locks, which is expected

AHAAAAAAA:
Been running for 45 minutes now, and tested on 2 machines. Seems pretty solid, +1

medghaim:
This is definitely a huge improvement. +1

AHAAAAAAA:
I wasn't sure how to do this on python, but hopefully this StackOverflow [post](https://stackoverflow.com/questions/1635080/terminate-a-multi-thread-python-program) is helpful.

DeirhX:
Made the threads daemons, hopefully that should help with termination, at least for now using Ctrl+C. 
However in general we might want to start thinking about entirely clean shutdowns, e.g. polling some global condition and making all threads termination-aware (e.g.: post special objects into their work queue once termination is requested)?

AHAAAAAAA:
excellent, +1. I agree, we kind of need to put the whole thing on hold and just do clean up.

medghaim:
Restesting

medghaim:
+1

meeeerge!

k33k00:
I've just updated to this & i'm curoius what exactle is search queue depth? I've randomly thrown 50 threads in there and I have a sea of  "search queue depth is: 99" may be throwing in a few 94 - 98 & the odd response dictionary error

wcastello:
Nice one! I learned something new today.

craywong:
Hows this working for you guys. I kept getting db locks when using 25 threads. Is the error related to the number of threads used?

k33k00:
Are you running sqlite or mysql? I'm under the impression that sqlite will really struggle with that many threads.

adamhwang:
So far so good, but please consider #2128 that addresses a small issue introduced by this change

PR #1836, opened 2016-07-24
easy-setup.sh: updates and cleanup

KhasMek:
## Description

  - Check for debain based OS otherwise quit.
  - Use already bundled get-pip.py over debian's.
  - Use sudo in commands instead of exiting.
  - Be a little smarter with the creation of credentials.
  - Be more verbose for the user.
  - Declare script as shell script
  - Give script executable permissions.
  - Use four space indents to stay inline with the python code.

## Motivation and Context

This now cleanly exits on non debian based systems, is declared as a shell script is more verbose with the user, uses already included scripts for install over semi functional ones before,  is set as executable and all around cleaner.

## How Has This Been Tested?

Tested on fresh clone of project on debain and non debian systems.

## Types of changes

## Checklist:

### DIFF
--- a/Easy Setup/easy-setup.sh
+++ b/Easy Setup/easy-setup.sh
@@ -16,7 +16,7 @@ fi
 echo "Installing pip..."
 sudo python get-pip.py
 echo "Installing required python packages..."
-sudo pip install -r requirements.txt
+pip install -r requirements.txt
 
 echo "Configuring Google Maps API..."
 cp ../config/credentials.json.example ../config/credentials.json
--- a/Easy Setup/easy-setup.sh
+++ b/Easy Setup/easy-setup.sh
@@ -1,18 +1,27 @@
-if [ "$EUID" -ne 0 ]
-  then echo "Please run as root. You can also do: sudo !!"
-  exit
+#!/bin/sh
+
+echo "Installing PokemonGo-Map Requirements."
+echo "This script may ask for your sudo/root password to perform the installation."
+echo ""
+
+if [ "$(grep -Ei 'debian|buntu|mint' /etc/*release)" ]; then
+    echo "Installing python development tools..."
+    sudo apt-get install python python-dev
+else
+    echo "This script only supports debain based Linux distros."
+    echo "Please install manually."
+    exit 1
 fi
 
-apt-get install python python-pip
-pip install -r requirements.txt
-pip install -r requirements.txt --upgrade
-echo -n "Enter your Google API key here:"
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
+1

Easy Setup/easy-setup.sh
```diff
@@ -1,18 +1,27 @@
...
+else
+    echo "This script only supports debain based Linux distros."
+    echo "Please install manually."
+    exit 1
+fi
+
+echo "Installing pip..."
+sudo python get-pip.py
+echo "Installing required python packages..."
+sudo pip install -r requirements.txt
```
Chlodochar: This shouldn't need sudo.

Easy Setup/easy-setup.sh
```diff
@@ -1,18 +1,27 @@
...
+
+echo "Configuring Google Maps API..."
+cp ../config/credentials.json.example ../config/credentials.json
+echo -n "Enter your Google Maps API key here:"
+read key
+sed -i -e "s/\"gmaps_key\"\ :\ \"\"/\"gmaps_key\"\ :\ \""$key"\"/g" ../config/credentials.json
```
Chlodochar: Just overwrite the whole file instead of replacing using sed. Replacing is more error-prone, since credentials.json can change at a later time.

KhasMek: That's precisely why I am replacing the entire line and not overwriting (or in the case of the previous code, amending the file - causing a huge mess if it were run multiple times). If we overwrite the file then this script will need to be modified every time  the file is updated. For example if/when other API keys are required for other functionality. In addition, replacing in this method will not force an overwrite if there is already an api key present, maybe the user added it manually before realizing there was an install script for the rest. Also, in line replacement is only error prone if you do it wrong or give it to vague of rules. This is the neither of those.

Chlodochar: Good points. Ignore my comment.

Easy Setup/easy-setup.sh
chuyskywalker: Yea...isn't using `sudo` during a `pip install` a "you shouldn't do that!" kind of thing? I know `npm` gets real angry about that.

AHAAAAAAA: I can merge after sudo is removed.

Chlodochar:
:+1: Looks good.

C-5:
Updates ever #1836  is the Number?

PR #1837, opened 2016-07-24
Added longpress to change location

adamhwang:
## Description
Click and hold the map 3-seconds to change the next location and map pin.

## Motivation and Context
Easily change the location without needing the search box and doesn't require access to the original pin to drag so you can change the location farther distances.  Doesn't break zooming with double click and is mobile friendly.

## How Has This Been Tested?
Tested on PC and Android in Chrome.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -306,6 +306,25 @@ function createSearchMarker() {
                 }
             });
     });
+    
+    var longpress = false;
+    google.maps.event.addListener(map, 'mousedown', function(e) { 
+        longpress = true;
+        setTimeout(function() {
+            if(longpress === true) {
+                changeSearchLocation(e.latLng.lat(), e.latLng.lng())
+                    .done(function() {
+                        marker.setPosition(e.latLng);
+                    });
+            }
+        }, 3000);
+    });
+    google.maps.event.addListener(map, 'mouseup', function(e) { 
+        longpress = false;
+    });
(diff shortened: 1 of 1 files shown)


andrewda:
Three seconds seems like a very long time. Isn't a longpress usually 1 second?

adamhwang:
Fixed a small issue on touch screen and decreased duration to 1s.

MonkeyLeeT:
I would not recommend this at all. User might be just mousdown for 1 sec and then decide to not drag or  anything. We've already provided many ways to relocate, keep them simple.

adamhwang:
This is the primary method all my friends and I use to move the search location and solves a real gap in the current map.  I understand that there are 3 PRs adding essentially this same functionality (#1912, #1808) in different way, and clearly we do not need three different ways to do this.  However, in my opinion, this is a fairly major usability issue (see pokevision.com).  I would really hope you guys reconsider either this or one of the other methods of doing this.  Personally, I would rather not have zoom than not have this functionality.

MonkeyLeeT:
Apparently our project manager decide to postpone these to next release. Let's wait until then. @AHAAAAAAA

adamhwang:
Long press is fairly unstable on mobile.  I'd still hope to see either of the other two PRs accepted.

vinsilsim:
Possible change for normal click to set new location? without wait 1s?

PR #1839, opened 2016-07-24
cache args to prevent re-requesting password

qadan:
## Description
Adds a simple memoization decorator to cache the argparser args

## Motivation and Context
`pogom/utils.py get_args()` is called three times during startup - [here](https://github.com/AHAAAAAAA/PokemonGo-Map/blob/10086b048711a133770cd0eb1b36569b3a1bed23/runserver.py#L40), [here](https://github.com/AHAAAAAAA/PokemonGo-Map/blob/10086b048711a133770cd0eb1b36569b3a1bed23/pogom/customLog.py#L5), and [here](https://github.com/AHAAAAAAA/PokemonGo-Map/blob/10086b048711a133770cd0eb1b36569b3a1bed23/pogom/models.py#L22). It's not much of a performance hit, but if the `--password` argument is not passed in or set in the config, it's going to be asked for three times [here](https://github.com/AHAAAAAAA/PokemonGo-Map/blob/10086b048711a133770cd0eb1b36569b3a1bed23/pogom/utils.py#L116) before actually getting used.

Using this decorator, the args are cached and future requests for the password - plus constructing the return dict multiple times - don't have to occur.

## How Has This Been Tested?
I added the decorator to the development branch and ran runserver.py to make sure the arguments were still accepted appropriately and that I was only asked for a password once.

## Types of changes

## Checklist:


MonkeyLeeT:
Interesting change, can confirm working as intended. +1 from me.

MonkeyLeeT:
Could you rebase?

qadan:
Merge conflict is resolved; just waitin' on CI

AHAAAAAAA:
No longer an issue!

PR #1844, opened 2016-07-24
Minimize update-size in raw_data

Chlodochar:
## Description
This optimizes the updates send using `raw_data` to only contain items that were changed in-between updates.

### Database

This adds `last_modified` to Pokemon, so that the server knows when a Pokemon was found.
This also restructures the queries, so that lastModified can be used optionally.

### Server

This restructures the `raw_data` function a bit so that queries are done in a more flexible manner. This was not required, but it made the resulting code smaller and (imo) more readable. Another side-effect is that the viewmodels are created in `app.py` instead of `model.py`. It seemed to me `model.py` should only contain the database models.

### Client

The client now tracks when it was last updated and with what parameters (bounds, showPokemon, showGym, etc). If the parameters stay the same, a minimal update can be done. If the parameters do change (by moving the map) a full update is required.

## Motivation and Context

I have a limited data subscription on my phone and sending over the whole set of pokemon, pokestops and gyms every 5 seconds took quite a bit of data.

## How Has This Been Tested?

Ran on my PC. Checked network activity in browser and checked console.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -8,7 +8,7 @@ from flask import Flask, jsonify, render_template, request
 from flask.json import JSONEncoder
 from flask_compress import Compress
 from datetime import datetime
-from dateutil import parser
+import time
 from s2sphere import *
 from pogom.utils import get_args, get_pokemon_name
 
@@ -18,13 +18,6 @@ from .models import Pokemon, Gym, Pokestop, ScannedLocation
 args = get_args()
 log = logging.getLogger(__name__)
 compress = Compress()
-
-from datetime import datetime, tzinfo, timedelta
-class simple_utc(tzinfo):
-    def tzname(self):
-        return "UTC"
-    def utcoffset(self, dt):
-        return timedelta(0)
 
 class Pogom(Flask):
     def __init__(self, import_name, **kwargs):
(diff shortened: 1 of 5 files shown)


Chlodochar:
Just for some information:

Before this PR updates were around 60,000 bytes uncompressed and 10,000 bytes compressed.
With this PR updates are around 350 bytes uncompressed and 100 bytes compressed.

Presuming updates are being done every 5 seconds, that'll be 720 updates per hour. That'll get you 7.2MB every hour before this commit and 72K after this commit. This is only presuming the map isn't moved or options are tweaked.

I tried this on Central Park, NY. In practice the effects will be less than what I've described above.

Chlodochar:
I'd like to get this merged. I've been running this on my server for a while. It seems to be working well. Last change also removed a dependency on dateutil.

Quite a lot of changes and a added column to the pokemon table in the database. It would be nice to know what others think. Atm this PR seems to be dying :(. Is this an unwanted feature? Should I split this PR up into database-related changes and http-related changes?

Any feedback is welcome!

marauder37:
I think it's a good PR, things just get buried now the PR list is back to 3 pages long

Chlodochar:
Please upvote. I can only hope people sort the issue list by +1 instead of last updated.

marauder37:
I have started a new habit of checking the entire PR list for things that don't have merge conflicts, in case they can be merged in speedily

Archalias:
I would really like to see this merged. Currently my mobile data is seeing 10x use compared to pokemon go.
The only thing I use data in my browser for is loading my PokemonGo-Map.
457mb from Internet
52.33 from Pokemon GO

Chlodochar:
Sounds like you're in the same boat as me. The current update method is inefficient.

PR #1864, opened 2016-07-24
Remove pokemon_name from json payload

douglasmiller:
This is part 1 of an attempt to reduce the json payload size.

## Description
Removes the `pokemon_name` key from the json object that is sent from the server to the client. The `pokemon_name` can be obtained in the client by mapping to the `idToPokemon` object that already exists in the client.

## Motivation and Context
This is  an attempt to further reduce the json payload size.

## How Has This Been Tested?
- Manual verification that the mapping data exists in the browser.
- Manual verification that the rendered map marker contains the mapped pokemon name.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -14,7 +14,7 @@ from datetime import datetime, timedelta
 from base64 import b64encode
 
 from . import config
-from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, get_args, send_to_webhook
+from .utils import get_args, send_to_webhook
 from .transform import transform_from_wgs_to_gcj
 from .customLog import printPokemon
 
@@ -93,9 +93,6 @@ class Pokemon(BaseModel):
 
         pokemons = []
         for p in query:
-            p['pokemon_name'] = get_pokemon_name(p['pokemon_id'])
-            p['pokemon_rarity'] = get_pokemon_rarity(p['pokemon_id'])
-            p['pokemon_types'] = get_pokemon_types(p['pokemon_id'])
             if args.china:
                 p['latitude'], p['longitude'] = \
                     transform_from_wgs_to_gcj(p['latitude'], p['longitude'])
@@ -124,9 +121,6 @@ class Pokemon(BaseModel):
 
         pokemons = []
         for p in query:
-            p['pokemon_name'] = get_pokemon_name(p['pokemon_id'])
-            p['pokemon_rarity'] = get_pokemon_rarity(p['pokemon_id'])
-            p['pokemon_types'] = get_pokemon_types(p['pokemon_id'])
             if args.china:
                 p['latitude'], p['longitude'] = \
                     transform_from_wgs_to_gcj(p['latitude'], p['longitude'])
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -250,10 +250,6 @@ def get_pokemon_name(pokemon_id):
 def get_pokemon_rarity(pokemon_id):
     return i8ln(get_pokemon_data(pokemon_id)['rarity'])
 
-def get_pokemon_types(pokemon_id):
-    pokemon_types = get_pokemon_data(pokemon_id)['types']
-    return map(lambda x: {"type": i8ln(x['type']), "color": x['color']}, pokemon_types)
-
 def send_to_webhook(message_type, message):
     args = get_args()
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1,6 +1,5 @@
 # PokemonGo Map![Python 2.7](https://img.shields.io/badge/python-2.7-blue.svg)[![Donate](https://img.shields.io/badge/Donate-PayPal-green.svg)](https://www.paypal.me/PoGoMapDev)
 
-
```
bobvanderlinden: This line is not part of the intended change. Please remove it.

douglasmiller:
I've removed the accidental white space removal from the README. Would you prefer that I rebase the change or leave it as is?

douglasmiller:
I didn't know if there was an aversion to force pushes while in code review. Some places prefer the additional commits.

Chlodochar:
Yes, and I'd agree for normal additions/fixes to the PR. However the README is totally unrelated to the PR. It could cause problems when others are also making the same unrelated changes.

douglasmiller:
Updated indentation issue in map.js.

Chlodochar:
:+1: Looks good to me! Tested and working.

static/map.js
```diff
@@ -327,6 +327,10 @@ function calculateSpritePoints(num) {
     return new google.maps.Point(30 * x, 30 * y);
 }
 
+function getPokemonName(item) {
+    return idToPokemon[item.pokemon_id];
+}
+
```
MonkeyLeeT: Can we just...get rid of this function and use direct access idToPokemon[item.pokemon_id] below?

denniskai1234:
Can you name some numbers how much we save here?

Chlodochar:
@denniskai1234 it isn't just for efficiency though. The names aren't needed by the server, only the client/browser needs them. In addition, the server doesn't need to know whether the browser is using a Japanese culture and therefore needs to render the names in Japanese. That's for the browser (and Javascript) to decide.

static/map.js
douglasmiller: Sure, it's all the same to me.

douglasmiller:
@denniskai1234, when I initially made these changes, I branched off of master. Now that `flask-compress` is added to the develop branch, the data savings will be minimal. There was also no bounds based request for pokemon, so the savings at the time was exceptional. I was already applying gzip compression via the proxy that I use to serve the map, but thought that there would be a non-trivial benefit for those that are not. 

Having said that, the json payload is still smaller even when gzipped. Even if it's only a few hundred bytes, that is not something to balk at when loading 100+ pokemon every 5 seconds. Those of us that do not have unlimited data plans on our phones can benefit from every bit saved.

There is also the separation of concerns that @Chlodochar mentioned. All of the data that is necessary to translate the pokemon id to the name is already present in the client. There is no reason to populate this data in the payload unnecessarily.

Chlodochar:
Btw, this also solves a problem with server vs browser locale, which exists right now. The server has its own locale and loads its own pokemon.x.json and the browser has a different locale and loads a different pokemon.x.json. Which of the two is used isn't clear, thus code and PR will choose whatever the dev finds first: fairly random.

With this PR being merged, devs are forced to use the pokemon names with the browsers locale, which  imo makes the most sense.

douglasmiller:
It looks like there are now 2 copies of `map.js`. Previously, there was only a `static/map.js`, but it looks like the contents of that file were copied to `static/dist/js/map.js`. Do we intend to maintain 2 copies of that file or should the, now unused, `static/map.js` be removed?

Chlodochar:
All files in `static/dist` should be removed from Git, as they are generated by Grunt. Please see #2384 to make those files finally go away.

douglasmiller:
I've also removed the, recently added, `pokemon_rarity` and `pokemon_types` data from the json payload. These data points are generic to all pokemon of a certain id and are not unique to the specific pokemon that is displayed on the map.

douglasmiller:
May I ask why this PR was closed?

danny6167:
https://github.com/AHAAAAAAA/PokemonGo-Map/blob/master/README.md

Chlodochar:
All PRs were closed by @AHAAAAAAA due to C&D.

douglasmiller:
Oh! What a bummer. Thanks for the great project. Working around in here kept me playing the game and expanded my knowledge of python.

Chlodochar:
Same here. Great to have been part of this project. It was a fun chaotic ride.

PR #1869, opened 2016-07-24
Swapped loading of javascript tags map.js and googlemaps

HanSolo80:
Changed loading of app and gmaps javascript

## Description
I reversed the loading of javascript on the bottom of the page.

## Motivation and Context
I often got errors while reloading the page saying that function "initMap" not found in gmaps api callback in script tag.
`<script src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places"></script>`
There were probably race conditions in loading the script tags.
I also removed async and defer loading because it is not really needed in my opinion.

## How Has This Been Tested?
Tested locally and on Heroku.

## Types of changes

## Checklist:

### DIFF
--- a/templates/map.html
+++ b/templates/map.html
@@ -143,7 +143,8 @@
 		var center_lng = {{lng}};
 	</script>
 
-	<script async defer src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places"></script>
+
 	<script src="static/map.js"></script>
+	<script src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places"></script>
 	</body>
 </html>
--- a/templates/map.html
+++ b/templates/map.html
@@ -144,7 +144,7 @@
 	</script>
 
 
-	<script src="static/map.js"></script>
-	<script src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places"></script>
+    <script defer src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places"></script>
+    <script src="static/map.js"></script>
 	</body>
 </html>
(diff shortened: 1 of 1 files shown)


Chlodochar:
Thanks for tackling this issue. It has been bugging me too.

That said, I don't understand why this change solves the issue. Doesn't gmaps need to be loaded before `maps.js`? Does removing `async` and `defer` by itself work already?

HanSolo80:
In my opinion maps.js needs to be loaded first so the function initMap is available when the callback is called after google maps api is loaded.

chuyskywalker:
I'd bet that removing the async option was all that was needed -- the `map.js` functions got wrapped in a jquery `$(function(){});` setup a little while ago, so it will wait for the page to be ready regardless or `<script>` tag source order -- however, the `async` flag probably is what causes the issue.

Can you try this with just removing the `async` option? 

I mean, I guess it doesn't matter, but might be interesting to know.

HanSolo80:
Seems to work like this. Tested locally and on Heroku. Reloaded a couple of times and didn't get js error. Although, since the error only occured spontaneously couldn't make sure if bug really is fixed.

templates/map.html
```diff
@@ -143,7 +143,8 @@
 		var center_lng = {{lng}};
 	</script>
 
-	<script async defer src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places"></script>
-	<script src="static/map.js"></script>
+
+    <script defer src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places"></script>
+    <script src="static/map.js"></script>
```
MonkeyLeeT: Indentation?

HanSolo80:
Indentation fixed

Chlodochar:
:+1: Looks good to me!

HanSolo80:
Lol, https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2043

Chlodochar:
Meh, not sure if I like it, but I guess this one can be closed. Agreed?

PR #1616, opened 2016-07-23
Added option to hide search bar

EdenGottlieb:
## Description

This allows users to hide the search bar by using the -fl (fixed location) parameter. It is designated for use in shared maps where relocation of the marker is not desired.
## Motivation and Context

I have a map of my city I shared with my team members and ever since the search bar was introduced people keep changing it.
## How Has This Been Tested?

Running the app with the parameter hides the search bar (display:none so it doesn't affect the UI) while running it without the parameter keeps the search bar intact.
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

Toggled by adding -fl (fixed location) when calling the function.

PR #1888, opened 2016-07-24
allow fixing map settings

markusz:
Similar to #1616 , this allows for disabling the settings and filters on shared servers

## Description

- Adds an optional additional flag `-fs` that disables the settings in the options menu
- Removes the options button completely if both setting and location are fixed

## Motivation and Context

Same reasoning than merged PR #1616, only for settings:

> This allows users to hide the search bar by using the -fl (fixed location) parameter. It is
> designated for use in shared maps where relocation of the marker is not desired.

and

> I have a map of my city I shared with my team members and ever since the search bar was 
> introduced people keep changing it.

## How Has This Been Tested?

Local server with all combinations

## Types of changes

## Checklist:

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -30,17 +30,20 @@ class Pogom(Flask):
 
     def fullmap(self):
         args = get_args()
-        display = "inline"
+        display_location = "inline"
+        display_settings = "inline"
         if args.fixed_location:
-            display = "none"
-        
+            display_location = "none"
+        if args.fixed_settings:
+            display_settings = "none"
+
         return render_template('map.html',
                                lat=config['ORIGINAL_LATITUDE'],
                                lng=config['ORIGINAL_LONGITUDE'],
                                gmaps_key=config['GMAPS_KEY'],
                                lang=config['LOCALE'],
-                               is_fixed=display
-                               )
+                               is_fixed_location=display_location,
+                               is_fixed_settings=display_settings)
 
     def raw_data(self):
         d = {}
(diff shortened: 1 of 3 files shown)


markusz:
Note: Changes in `templates/map.html` are mostly related to adding a parent `div`, changing the intendation (and only the intendation) of the existing `div` elements

jangrewe:
Why would you want to hide the settings? Yes, i'm also running a shared map and i hide the location search box and block /next_loc, but i think denying users to hide markers they don't want to see isn't really appreciated by many users - or did i completely misunderstand your changes? ;-)

markusz:
@jangrewe the filters unfortunatelly are applied for all users, so if user A filterst out a pokemon, user B will also not see it

jangrewe:
You sure about that? The markers to fetch are set per-request, so if you open your map e.g. Chrome and Firefox, you can have each display different markers. At least i can...

markusz:
Pretty sure. 

I just tried it out with two different windows (not tabs etc.) of chrome and both filtering and displaying the search grid affected the other

jangrewe:
Two tabs of Chrome? Well, there's your problem. Again: the settings are saved to localStorage. 
localStorage is shared between all tabs (on the same page).

Use 2 different browsers (Incognitor Mode MAY also work), and you'll see...

markusz:
Ah I see now. Kinda obvious actually - time to go to bed :D

MonkeyLeeT:
Not a bug to fix as it seems. Closed.

Issue #1877, opened 2016-07-24
Impossible to access from WAN


FlorentDC:
Hey, that's weird.

I'm following this repo for a while, and i've implemented a lot of things from the pull requests, I think that I understand how this things work, but since the v2 I can't access the map from the WAN.

I've tried to change the configs and run with -se, but the map is unreachable, same with -H & -P arguemnts, same with Host and Port hard coded, and same when i change the 'default' value in pogom/utils.py.

I tried to retry from the begining 3 times (deleting everything and retrying), but nothing 

I search in issues and it seems l'ike I'm the only one with this bug hm ?
If someone would have an idea, or somewhere to hard-code it in a file, that would be great !

Excuse my english, I'm French (:

Animosity:
This is a network configuration issue (your unique gateway/router combination, and/or your server firewall configurations are at fault).

fust:
Are you behind a NAT router? If so, make sure you forwarded the specified port to the machine running the PokemonGo-map instance.

FlorentDC:
I thought that it wasn't due to NAT or something else, because accessed it well when I was using the master & dev branches, but if both of you tell it to me, i'll dig this way, thanks guys and sorry and sorry for that false positive

Beeneh:
So i've got the same issue. I try to run the following command (anonymized of course): 

`python runserver.py -u username -p password -k [api key] -l "[location]" -st 10 -H 1.2.3.4 -P 5000`

and it returns: 

`2016-07-24 21:35:34,196 [  runserver] [   INFO] Parsed location is: [location] (lat/lng/alt)
2016-07-24 21:35:34,196 [     search] [   INFO] Map iteration: 0
2016-07-24 21:35:34,196 [     search] [   INFO] Attempting login to Pokemon Go.
Traceback (most recent call last):
  File "runserver.py", line 83, in <module>
2016-07-24 21:35:34,196 [   auth_ptc] [   INFO] PTC login for: [username]
    app.run(threaded=True, debug=args.debug, host=args.host, port=args.port)
  File "D:\Python27\lib\site-packages\flask\app.py", line 843, in run
    run_simple(host, port, self, **options)
  File "D:\Python27\lib\site-packages\werkzeug\serving.py", line 694, in run_simple
    inner()
  File "D:\Python27\lib\site-packages\werkzeug\serving.py", line 656, in inner
    fd=fd)
[...]

PR #1890, opened 2016-07-24
Nicer error if invalid IP/port combination is given (See #1877)

fust:
## Description
Added check before binding to any addresses to see if they are valid in the current context

## Motivation and Context
Solves users entering invalid IP addresses to bind to getting exceptions printed in their console.
A lot of users don't have the technical know-how to understand the exceptions, better error handling prevents a lot of unneeded issues.
See #1877

## How Has This Been Tested?
Using some invalid IP addresses

## Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [ x] My code follows the code style of this project.

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -5,6 +5,8 @@ import os
 import sys
 import logging
 import time
+import socket
+import sys
 
 from threading import Thread
 from flask_cors import CORS, cross_origin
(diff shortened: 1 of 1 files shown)


MonkeyLeeT:
Shouldn't Flask's default error message be enough? Why do the needless?

fust:
It should, and to developers it probably is, however as seen in the referenced ticket it's not clear to users without the technical know-how. If we can display nice error messages describing exactly what's wrong and that helps to cut down on issues I think it's something we should want.

MonkeyLeeT:
Again it seems weird to do binding twice only for the sake of better notification to average user. Average user will soon be able to use front-end directly (supposedly), so can we just trust Flask's default message for now?

fust:
Agreed it's a bit weird, couldn't whip up a better solution though.
If the average user can set all this data through the frontend so we can do even better error reporting (or even better, -prevention) that would be the preferred solution and Flask's default message is enough for now.
Closing this PR.

Issue #1897, opened 2016-07-24
PGoApi is not thread-safe but it being used as such (missing pokemon on map)


DeirhX:
Increase to concurrency from https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1835 now reveals that PGoApi is not thread-safe but is used concurrently to generate and dispatch requests. This currently leads into following behavior:
Threads 1 to N enter send_map_requests() simultaneously
Threads 1 to N call api.set_position simultaneously and overwrite the one shared value
Threads 1 to N call api.get_map_objects simultaneously and each of them adds its request. api now contains N such requests.
Threads 1 to N call api.call simultaneously which means the N requests get dispatched N times [O(N*N) !] all the while sharing the same position.
Chaos ensues, some cells are skipped since call() uses the position set by set_position which was overwritten by the last thread.

## Expected Behavior
Concurrent requests work the same way as if were sent serially

## Current Behavior
PGoApi keeps internal state between constructing and sending requests, opening a window to any concurrently running thread to overwrite it. This leads to missing **pokemon on map** since some cell requests are skipped due to api's internal position being overwritten by another thread before the RPC call is dispatched.

## Possible Solution
PGoApi needs to be thread-safe or immutable after login. Sharing login state between all requests definitely makes sense, sharing position does not.

DeirhX:
I suggest either separating login state and request construction into two classes (login state being immutable so it doesn't have to be locked when checking if already logged in) or removing all the other internal state from PGoApi (position, req_method_list) so multiple threads can each construct their own requests independently without affecting its shared state.

AHAAAAAAA:
@tejado thoughts?

DeirhX:
As easiest solution for now I would opt for splitting PGoApiRequest from PGoApi so each thread can construct its request object independently. I can code it and send a Pull Request for review.

AHAAAAAAA:
Good idea, I'd love to test that out. It would be great if you can do a PR against the main pgoapi repo too!

DeirhX:
Sure thing. Should be ready shortly, I hope.

Chlodochar:
Sounds good. It probably doesn't need changes to pgoapi itself right? Otherwise, please consider merging #671.

DeirhX:
[...]

PR #1910, opened 2016-07-25
Refactored PGoApi to make it immutable, PGoApiRequest introduced (solve missing pokemon)

DeirhX:
This is a solution to concurrency problems with PGoApi as mentioned in https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1897. PGoApi was split into immutable (login) and mutable parts (position, request collection) so it can be used safely between threads.
Mutable parts were transferred into PGoApiRequest class which now represent the state of an RPC request and each thread now gets a new instance of it to prevent data races as are happening now.

## Description
Spinned of PGoApiRequest from PGoApi. PGoApi is now immutable and thread-safe, representing  established login to server. PGoApiRequest is now instanced from PGoApi for each new RPC request you want to make.

Feel free to refactor further and adjust coding style if needed ;) 

## Motivation and Context
Described in depth in https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1897. Data-races are now causing scan results to be incomplete and pokemon missing from scan results.

## How Has This Been Tested?
Ran a stress test and scans are now complete and faster than before.

## Types of changes

## Checklist:

- [x] My code follows the code style of this project.   // I hope ;)

### DIFF
--- a/pogom/pgoapi/pgoapi.py
+++ b/pogom/pgoapi/pgoapi.py
@@ -38,142 +38,175 @@ import protos.RpcEnum_pb2 as RpcEnum
 
 logger = logging.getLogger(__name__)
 
-class PGoApi:
 
+
+class PGoApi:
     API_ENTRY = 'https://pgorelease.nianticlabs.com/plfe/rpc'
 
     def __init__(self):
-    
+
         self.log = logging.getLogger(__name__)
 
         self._auth_provider = None
-        self._api_endpoint = None
-        
+        self._api_endpoint = self.API_ENTRY
+
         self._position_lat = 0
         self._position_lng = 0
         self._position_alt = 0
 
-        self._req_method_list = []
-        
-    def call(self):
-        if not self._req_method_list:
-            return False
-        
-        if self._auth_provider is None or not self._auth_provider.is_login():
-            self.log.info('Not logged in')
-            return False
-        
-        player_position = self.get_position()
-        
-        request = RpcApi(self._auth_provider)
-        
-        if self._api_endpoint:
-            api_endpoint = self._api_endpoint
-        else:
-            api_endpoint = self.API_ENTRY
-        
-        self.log.info('Execution of RPC')
-        response = None
-        try:
-            response = request.request(api_endpoint, self._req_method_list, player_position)
-        except ServerBusyOrOfflineException as e:
-            self.log.info('Server seems to be busy or offline - try again!')
-        
-        # cleanup after call execution
-        self.log.info('Cleanup of request!')
-        self._req_method_list = []
-        
-        return response
-    
-    #def get_player(self):
-    
-    def list_curr_methods(self):
-        for i in self._req_method_list:
-            print("{} ({})".format(RpcEnum.RequestMethod.Name(i),i))
-    
-    def set_logger(self, logger):
-        self._ = logger or logging.getLogger(__name__)
+    def get_auth_provider(self):
+        return self._auth_provider
+
+    def get_endpoint(self):
+        return self._api_endpoint
 
     def get_position(self):
         return (self._position_lat, self._position_lng, self._position_alt)
 
-    def set_position(self, lat, lng, alt):   
-        self.log.debug('Set Position - Lat: %s Long: %s Alt: %s', lat, lng, alt)
-        
-        self._position_lat = f2i(lat)
-        self._position_lng = f2i(lng)
-        self._position_alt = f2i(alt)
+    def set_logger(self, logger):
+        self._ = logger or logging.getLogger(__name__)
 
-    def __getattr__(self, func):
-        def function(**kwargs):
-        
-            if not self._req_method_list:
-                self.log.info('Create new request...')
-        
-            name = func.upper()
-            if kwargs:
-                self._req_method_list.append( { RpcEnum.RequestMethod.Value(name): kwargs } )
-                self.log.info("Adding '%s' to RPC request including arguments", name)
-                self.log.debug("Arguments of '%s': \n\r%s", name, kwargs)
-            else:
-                self._req_method_list.append( RpcEnum.RequestMethod.Value(name) )
-                self.log.info("Adding '%s' to RPC request", name)
-   
-            return self
-   
-        if func.upper() in RpcEnum.RequestMethod.keys():
-            return function
-        else:
-            raise AttributeError
-            
-        
-    def login(self, provider, username, password):
-    
+    def create_request(self):
+        request = PGoApiRequest(self, self._position_lat, self._position_lng, self._position_alt)
+        return request
+
+    def login(self, provider, username, password, position_lat, position_lng, position_alt):
+
+        self._position_lat = f2i(position_lat)
+        self._position_lng = f2i(position_lng)
+        self._position_alt = f2i(position_alt)
+
         if not isinstance(username, basestring) or not isinstance(password, basestring):
             raise AuthException("Username/password not correctly specified")
-        
+
         if provider == 'ptc':
             self._auth_provider = AuthPtc()
         elif provider == 'google':
             self._auth_provider = AuthGoogle()
         else:
             raise AuthException("Invalid authentication provider - only ptc/google available.")
-            
+
         self.log.debug('Auth provider: %s', provider)
-        
+
         if not self._auth_provider.login(username, password):
-            self.log.info('Login process failed') 
+            self.log.info('Login process failed')
             return False
-        
+
         self.log.info('Starting RPC login sequence (app simulation)')
-        
+
         # making a standard call, like it is also done by the client
-        self.get_player()
-        self.get_hatched_eggs()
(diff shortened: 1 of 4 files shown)


AHAAAAAAA:
To replace #1913 hotfix once #671 is merged in following API update.

...
Say that 10 times in a row.

Chlodochar:
Best would be to do the PR on pgoapi. After #671 is merged, it's trivial to use the new pgoapi.

pogom/pgoapi/pgoapi.py
```diff
@@ -38,142 +38,175 @@
...
         self._position_alt = 0
 
+    def get_auth_provider(self):
+        return self._auth_provider
+
+    def get_endpoint(self):
+        return self._api_endpoint
+
+    def get_position(self):
+        return (self._position_lat, self._position_lng, self._position_alt)
```
Chlodochar: Is `get_position` and position in general still needed in PGoApi? `set_position` is being called on the request before every call. That seems good. It seems passing the position from PGoApi to request isn't needed.

CT9AJ:
Is there anyway I could apply this hotfix?

Chlodochar:
From what I remember:
```
git fetch https://github.com/DeirhX/PokemonGo-Map dev-feature-lockfree && git merge FETCH_HEAD
```

CT9AJ:
I am running the Master version, I assume that would not apply for me right?

Chlodochar:
@CT9AJ You'll probably be better off waiting for this all to drop into master.

CT9AJ:
I actually already just switched to the Dev build lol.. Where do I put that line you just gave me? Do I do a Git Bash Here from the Dev folder?

Chlodochar:
Yes, but please leave these questions somewhere else. See discord. The pull requests are specific to feature additions that are usually (initially) unstable.

CT9AJ:
Ok thanks.

Josephvb10:
Hmm, seems like this could be introducing a problem. 

Tried to use this fix, but after a few hours, the infamous bug that causes Pokémon to stop appearing return. The solution is to restart the script manually.  @DeirhX

pogom/pgoapi/pgoapi.py
DeirhX: Yeah, I was considering removing at least the getter. As for the position value itself I was unsure if the original login position wouldn't be needed for future requests but it isn't right now.

DeirhX:
@Josephvb10 Interesting, I wouldn't expect it to be introducing new issues. If Pokemon stop appearing altogether that would be probably due to server issues as currently the PTC login is again unstable. Or do you mean you are getting incomplete results and Pokemon missing from some of the scanned cells?

Josephvb10:
@DeirhX Sorry, I edited the post. Seems like the TPC login server has been down a lot for the last hour, so that could be the problem. I will keep this version running and report back if I found another issue.

DeirhX:
Myself I have kept running it overnight and apart from scanners having problems with PTC servers, results seem stable for me still.

pogom/pgoapi/pgoapi.py
DeirhX: Anyway I agree that the position can go to keep it simple. We can reintroduce it should future needs call for it.

everix1992:
I pulled this down last night so I could test it as well. It worked flawlessly for an hour or two and then I ran into the same errors that @Josephvb10 did. Restarting the script seemed to fix the problem. I assumed it had something to do with the number of API requests that were being sent from a properly multithreaded server, but this is just wild speculation.

Josephvb10:
Yes. Happened to me again. Looks like it wasn't related to the servers being down after all. @DeirhX

Chlodochar:
It's quite hard to wrap my head around the code, but having `api` be global in `search.py` seems dangerous. https://github.com/DeirhX/PokemonGo-Map/blob/cf64dd3d1f1007926b09681cc4a8f2447a7de9bb/pogom/search.py#L153
Is this actually needed? I'm interested whether a login per thread (and completely separating state) will solve the issue. Or are multiple login sessions for the same account not allowed?

DeirhX:
Since PGoApi is now immutable, sharing it as global is harmless as there is no way you can change its state (as long you synchronize all assignments to it). I don't think we need a separate login per thread so that's why it is shared - but feel free to suggest a better way to share it. What we do need is are separate PGoApiRequests per thread so constructing the RPC call in one thread doesn't overwrite values set from another thread which is what was happening now (since it is being shared and has internal state).

@everix1992 , @Josephvb10 this refactor doesn't attempt to address issues with failing the requests and subsequent relogin to PTC server. That's a different issue (however I have a potential solution ready that depends on this PR being merged in - but it's hard to say if it's 100% yet). This PR addresses the issue of pokemon being missing from some cells you successfully scanned because multiple threads overwrote the position data of each other's request and thus multiple requests used the same position (yet still landed in different cells since that's a different variable). It also addresses performance in cases where multiple threads concurrently wrote into PGoApi request and duplicated many of their requests.

DeirhX:
To explain the reasoning behind founding PGoApiRequest class better:
Currently, all threads share the same PGoApi.position and PGoApi.req_method_list. This is inherently bad because requests are made by first updating the position variable then populating req_method_list and then calling call() to actually start dispatching the request. If another thread happens to also want to call at the same moment and starts preparing the request, it overwrites the values of the first thread because **there is only one global PGoApi object**.  Position and req_method_list must not be shared between threads! They represent unique state of each thread's requests and each thread wants to make a different request!

DeirhX:
On the other hand login information is never modified after login is successfully established. Therefore this can be safely shared between threads without it requiring to be copied around. And since Niantic uses HTTPS as the transport layer, there is no connection, no open socket that would make a problem when accessed from multiple threads. So I think we can safely retain one login state per process - which is what we already do.

DeirhX:
On the other hand login information is never modified after login is successfully established. Therefore this can be safely shared between threads without it requiring to be copied around. And since Niantic uses HTTPS as the transport layer, there is no connection, no open socket that would make a problem when accessed from multiple threads. So I think we can safely retain one login state per process - which is what we already do.  It's the mutable internal state of constructing additional requests that is a problem.

pogom/search.py
```diff
@@ -149,21 +150,20 @@ def search(args, i):
...
         if remaining_time > 60:
             log.info("Skipping Pokemon Go login process since already logged in for another {:.2f} seconds".format(remaining_time))
         else:
-            login(args, position)
-    else:
-        login(args, position)
+            active_api = None
 
-    lock = Lock()
+    if not active_api:
```
Chlodochar: It seems safer to have one thread at once do `api = login(args, position)` and have others wait for that operation. At the moment multiple threads can login and set `api` at the same time. I'm not sure that always results in the right token. Why not add a global lock `apiLock` that is used like:

```
apiLock = Lock()
...
global apiLock
...
if not active_api:
    apiLock.acquire()
    try:
        # Check whether another thread already refreshed api while we were waiting for the lock.
        if api._auth_provider._ticket_expire / 1000 - time.time() < 0:
            api = login(args, position)
        active_api = api
    finally:
        apiLock.release()
```

Let me know if it's unnecessary.

DeirhX: You are completely right, I forgot the lock in this revision. All other threads should indeed wait whenever a login operation is already in progress, at least that was my intention as well (and current unpushed implementation). 

Thanks for the heads up. I'll commit the fix once I get home.

DeirhX:
@Chlodochar Pushed new commit that resolves the missing lock and also removes redundant lock on search pass since it doesn't use any shared data anymore (except database commit which should take care of its own locking dependent on engine). Please check whether you agree with it being redundant. It should speed up the scan a little bit more.

Also moved the login code into the search worker, closer to actual request happening. This will give the search worker opportunity to request a relogin if it feels the connection has gone stale (wrong responses or no responses at all). Previously there was not any possibility to wait for the relogin from within inside the search. This should help with issues @everix1992 and @Josephvb10 were experiencing since relogin (done by restarting the process) seems to help in these cases. It is, however, a larger change so feel free to cherry pick only the solution to the previous problem.

everix1992:
Running for an hour now and everything seems fine. I'll keep you updated on whether I have any issues.

everix1992:
I'm not sure if I just got used to using your changes, but the search seems to run much slower on the main develop branch as well, so the sooner we can get this pulled in the better. Let me know if there is anything I can do to help with that! I may have a harder time testing it now though as it's become a pain switching between this and develop now (due to config and webhook changes).

Chlodochar:
This looks good. We need to get this into pgoapi as well. (@tejado)

tejado:
I have it already in my local codebase. Will push it today.

tejado:
FYI: https://github.com/tejado/pgoapi/commit/1b355130ad6d0a64a3267eb9dc3186cbbcd6aa68

AHAAAAAAA:
Work on this is moved to #3106

PR #1911, opened 2016-07-25
Speed up JSON encoding

TomProkop:
## Description
Switch from the Flask's jsonify to remove a bottleneck.

## Motivation and Context
While processing a large amount of data server was really slow. I've narrowed down the issue.

## How Has This Been Tested?
Windows/Chrome/MySQL
With jsonify I ran ~20 threads and soon after 1500 rows serialization took about 20 seconds to process.
Simplejson managed to do it within 3 seconds.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -3,8 +3,9 @@
 
 import calendar
 import logging
+import simplejson as json
 
-from flask import Flask, jsonify, render_template, request
+from flask import Flask, render_template, request
 from flask.json import JSONEncoder
 from flask_compress import Compress
 from datetime import datetime
@@ -64,14 +65,14 @@ class Pogom(Flask):
         if request.args.get('scanned', 'true') == 'true':
             d['scanned'] = ScannedLocation.get_recent(swLat, swLng, neLat, neLng)
 
-        return jsonify(d)
+        return json.dumps(d, cls=self.json_encoder)
 
     def loc(self):
         d = {}
         d['lat']=config['ORIGINAL_LATITUDE']
         d['lng']=config['ORIGINAL_LONGITUDE']
 
-        return jsonify(d)
+        return json.dumps(d, cls=self.json_encoder)
 
     def next_loc(self):
         args = get_args()
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -65,14 +65,14 @@ class Pogom(Flask):
         if request.args.get('scanned', 'true') == 'true':
             d['scanned'] = ScannedLocation.get_recent(swLat, swLng, neLat, neLng)
 
-        return json.dumps(d, cls=self.json_encoder)
+        return json.dumps(d, cls=self.json_encoder), 200, {'Content-Type': 'application/json'}
 
     def loc(self):
         d = {}
         d['lat']=config['ORIGINAL_LATITUDE']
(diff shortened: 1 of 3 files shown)


pogom/app.py
```diff
@@ -64,14 +65,14 @@ def raw_data(self):
...
-        return jsonify(d)
+        return json.dumps(d, cls=self.json_encoder)
 
     def loc(self):
         d = {}
         d['lat']=config['ORIGINAL_LATITUDE']
         d['lng']=config['ORIGINAL_LONGITUDE']
 
-        return jsonify(d)
+        return json.dumps(d, cls=self.json_encoder)
```
chuyskywalker: http://stackoverflow.com/questions/7907596/json-dumps-vs-flask-jsonify Looks like jsonify was deigned for direct responses and will set some headers/etc properly. Do we lose that here?

http://flask.pocoo.org/docs/0.11/api/#flask.json.jsonify

If the speed up is really from using `cls=self.json_encoder` perhaps that's all that was needed to be added?

TomProkop:
simplejson because: [http://stackoverflow.com/questions/712791/what-are-the-differences-between-json-and-simplejson-python-modules/17823905#17823905](url)

pogom/app.py
TomProkop: `cls=self.json_encoder` is there because of this [http://stackoverflow.com/questions/10721409/why-does-json-serialization-of-datetime-objects-in-python-not-work-out-of-the-bo](url)

Regarding headers: indeed it's not there. However in this case we can omit that or just add it manually.

TomProkop: I've done it.

TomProkop:
And I just found this [http://stackoverflow.com/questions/37931927/why-is-flasks-jsonify-method-slow](url)

chuyskywalker:
It's interesting that this speeds anything up since the `jsonify` code itself states:
```
1. Single argument: Passed straight through to :func:`dumps``
```
But that's a comment, and comments can be wrong or dumb :)

With the header/etc, though, who cares as long as it's faster!

TomProkop:
Some kind of weird heuristic :)
Anyway Jsonify uses json.dumps and does the indentation stuff..

From the reading simplejson.dumps should be slightly faster than json.dumps.

This simple trick and switching from SQLite helped me with covering a large area.

AHAAAAAAA:
According to the same post you linked, apparently [ujson](https://pypi.python.org/pypi/ujson) is 3x faster than simplejson. Thoughts?

colinchilds:
Don't you need to add this to requirements.json in the root folder as well?

DerScreever:
http://artem.krylysov.com/blog/2015/09/29/benchmark-python-json-libraries/

AHAAAAAAA:
Looks good, but I'm still wondering why we're not using ujson instead of simplejson if the benchmarks show it clearly outperforming it?

TomProkop:
I am not able to test it atm. They've banned my account and pokemon.com returns 503 while trying to create a new one.

MonkeyLeeT:
Let's wait until you can test out ujson.

AHAAAAAAA:
Any updates on this?

MonkeyLeeT:
Please just send in a new one if ujson can be that much faster. For now we're fine with default one.

PR #1912, opened 2016-07-25
Add scan-here button

tkSimon:
## Description
Adds a "scan here" link to the header. When clicked/touched, will set the current map center as new search location.

## Motivation and Context
When outside hunting Pokemon you often want to check a few nearby places before deciding which way to go. This is way more convenient when you can just move the map one or two kilometers/miles and tap "scan here" instead of having to type in the name of some location.

## How Has This Been Tested?
Tested using docker.

## Types of changes

## Checklist:
- [X] My code follows the code style of this project. (I think?)

### DIFF
--- a/static/sass/layout/_main.scss
+++ b/static/sass/layout/_main.scss
@@ -31,3 +31,26 @@ html, body {
 	bottom: 0px;
   width: 100%;
 }
+
+.fab {
+  background: darken(_palette(accent2, bg), 12);
+  color: _palette(accent2, fg);
+  width:30px;
+  height:30px;
+  border-radius:100%;
+  border:none;
+  outline:none;
+  font-size:19px;
+  box-shadow: 0 3px 6px rgba(0,0,0,0.16), 0 3px 6px rgba(0,0,0,0.23);
+  transition:.3s;
+  -webkit-tap-highlight-color: rgba(0,0,0,0);
+  text-align: center;
+  @include icon('\f05b');
+}
+
+#scan-here {
+  position: fixed;
+  left: 50%;
+  margin-left: -15px;
+  bottom: 1.5em;
+}
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -297,6 +297,11 @@ function initMap() {
     redrawPokemon(map_data.pokemons);
     redrawPokemon(map_data.lure_pokemons);
   });
+
+  $('#scan-here').on('click', function() {
+    var loc = map.getCenter();
+    changeLocation(loc.lat(), loc.lng());
+  });
 }
 
 function createSearchMarker() {
--- a/templates/map.html
+++ b/templates/map.html
@@ -172,6 +172,7 @@
 		</div>
 	  </nav>
       <div id="map"></div>
+      <div class="fab" id="scan-here" style="display:{{is_fixed}}"></div>
     </div>
     <!-- Scripts -->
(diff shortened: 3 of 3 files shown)


static/map.js
```diff
@@ -119,6 +119,14 @@ function initMap() {
...
         updateMap();
     });
+
+    $('#scan-here').on('click touch', function(){
+	var loc = map.getCenter();
+	$.post("next_loc?lat=" + loc.lat() + "&lon=" + loc.lng(), {}).done(function (data) {
+            $("#next-location").val("");
+            marker.setPosition(loc);
+        });
+    });
```
andrewda: Please use the [`changeLocation(lat, lng)` function](https://github.com/tkSimon/PokemonGo-Map/blob/develop/static/map.js#L792-L798);

andrewda:
Needs a rebase, then +1

tkSimon:
rebased

tkSimon:
This is how it looks at the moment:
Screenshot

fholzer:
resolved merge conflict

tkSimon:
@andrewda 
Is there a discussion going on anywhere I didn't notice about which PR concerning the absence of such a feature should be pulled?

If not... if this topic is up for discussion by the community, maybe we should create an issue discussing pros and cons of each PRs. Obviously I'm a big fan of my solution, but at this point I just want to see any of these PRs pulled.

Also, if the reasoning why many users will want such a feature is not clear, I'd be happy to explain and give an example.

tkSimon:
rebased

tkSimon:
Updated "Motivation and Context."

andrewda:
@tkSimon just need one more collab to check off on this, then I can merge :)

tkSimon:
Sure thing. Btw, the screenshot is from back when the header bar layout was still broken (text not vertically centered)

fakemustache12:
Seems to interfere with the stats button now in the nav bar, at least on an iPhone 6 size device, this scan button overflows outside the nav bar.

Wondering if it might be better to switch to icons only for this and stats.

tkSimon:
That was bound to happen with people adding things to the header. I reworked the header now to show only icons and the page name on xsmall screens.

AHAAAAAAA:
Placement of 'scan here' feels clumsy, especially on a big screen. It also doesn't seem to be well-placed in Mobile resolutions. I'm unsure of how to proceed with this one, I'd opt for a floating button (e.g. like the one [here](http://materializecss.com/buttons.html)) that's not in the header and not intrusive.

tkSimon:
I guess you want that without adding materialize-css as a dependency?

tkSimon:
http://imgur.com/a/lxzsp
like so?

tkSimon:
updated screenshot: http://imgur.com/a/VYadA

tkSimon:
I'm not really happy with the icon, as it's similar to google map's integrated "Your location" button...
any suggestions?

AHAAAAAAA:
I would put that button bottom-center. I see what you mean about the button looking like google's, I'll give it a look.

tkSimon:
Just came to my mind that i still need to add something so this button is hidden when the server is started with `-fl`...

static/sass/layout/_main.scss
```diff
@@ -31,3 +31,26 @@ html, body {
 	bottom: 0px;
   width: 100%;
 }
+
+.fab {
```
tkSimon: not sure if this should go into main.scss....

tkSimon:
FAB hidden when using `-fl` now.

PR #1913, opened 2016-07-25
Hotfix to clone PGoAPI object when about to make multi-threaded calls.

DeirhX:
This is a temporary hotfix to https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1897 causing missing pokemon from cells. It should be replaced by https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1910 as soon as possible.

## Description
Implemented cloning of entire PGoApi object when about to be used in concurrent environment. Lazy hack but at least the map scans won't be incomplete anymore.

## Motivation and Context
See https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1897

## How Has This Been Tested?
Not very much, please test. I've run out of time today.

## Types of changes

## Checklist:

!TEMPORARY!  This should be removed and replaced by appropriate refactor.

### DIFF
--- a/pogom/pgoapi/pgoapi.py
+++ b/pogom/pgoapi/pgoapi.py
@@ -54,6 +54,17 @@ class PGoApi:
         self._position_alt = 0
 
         self._req_method_list = []
+
+    def copy(self):
+        other = PGoApi()
+        other.log = self.log
+        other._auth_provider = self._auth_provider
+        other._api_endpoint = self._api_endpoint
+        other._position_lat = self._position_lat
+        other._position_lng = self._position_lng
+        other._position_alt = self._position_alt
+        other._req_method_list = list(self._req_method_list)
+        return other
         
     def call(self):
         if not self._req_method_list:
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -36,12 +36,13 @@ def calculate_lng_degrees(lat):
 
 def send_map_request(api, position):
     try:
-        api.set_position(*position)
-        api.get_map_objects(latitude=f2i(position[0]),
+        api_copy = api.copy()
+        api_copy.set_position(*position)
+        api_copy.get_map_objects(latitude=f2i(position[0]),
                             longitude=f2i(position[1]),
                             since_timestamp_ms=TIMESTAMP,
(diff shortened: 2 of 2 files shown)


AHAAAAAAA:
Tested, +1

wesleytsai:
Been using this for a while now while catching pokemon. Works great as a hotfix. +1

pogom/pgoapi/pgoapi.py
```diff
@@ -54,6 +54,17 @@ def __init__(self):
...
+    def copy(self):
+        other = PGoApi()
+        other.log = self.log
+        other._auth_provider = self._auth_provider
+        other._api_endpoint = self._api_endpoint
+        other._position_lat = self._position_lat
+        other._position_lng = self._position_lng
+        other._position_alt = self._position_alt
+        other._req_method_list = list(self._req_method_list)
+        return other
```
chuyskywalker: Does python not have some kind of...`clone`?

chuyskywalker:
I pulled and tested this branch with `-t 1` and `-t 5` and saw not difference in number of found pokemon in the area. Which I suspect means this is working.

That said, I also downgraded to current `develop` and ran `-t 5` without any missing pokemon. I can, at the least say, it's not any worse.

andrewda:
Tested and looks like I got the same results as @chuyskywalker. Might just be that I have few Pokémon in my area to begin with, so this won't have much affect on my area.

pogom/pgoapi/pgoapi.py
DeirhX: Tell me if it does. I'm from C++ land, feel free to pythonize.

Chlodochar: It seems so: https://docs.python.org/2/library/copy.html

I'm not that familiar with Python, so not sure if this fits the purpose.

Chlodochar:
No problems other than `OperationalError: database is locked` with high concurrency. I think that was already there.

pogom/pgoapi/pgoapi.py
Chlodochar: Using copy might also keep pogom/pgoapi untouched.

PR #1916, opened 2016-07-25
SSL Support Returns

chuyskywalker:
Get your SSL on!

2016-07-24 17_56_14-pokemon go map

I also got rid of the old `credentials.json` file and added a polite warning that it should not be used since I was all up in the config code areas.


MonkeyLeeT:
Working solid for me. Https make me happy.

andrewda:
Working on Chrome+Firefox/Mac OSX.

AHAAAAAAA:
AttributeError: 'Namespace' object has no attribute 'ssl_key'.

Breaks regular build that doesn't use ssl

PR #1936, opened 2016-07-25
SSL Attempt 2

chuyskywalker:
Add SSL (again)

## How Has This Been Tested?
 - Command-line options only
 - cli options with `config.ini` present and filled in, no `--se`
 - cli options with `config.ini` present and filled in, with `--se`
 - various states of `credentials.json` (missing, there and correct, there and incorrect)

## Types of changes

- [x] Deprecation (feature being removed, but still functional)

## Checklist:


relrod:
Is there a convincing reason that this should be supported in the upstream as opposed to suggesting that people offload stuff like this to a reverse proxy (apache, nginx, etc.)?

I'm not a core developer for the project, but I feel like the app server should be as stupidly simple as possible (less things to get wrong). I feel like things like this and #1986, etc. should be handled elsewhere.

chuyskywalker:
@relrod: Normally, I'd agree with you, but I think that with this project and its userbase, it makes more sense to keep it as self-contained as possible. You can see from the bug tracker and discourse chat how much effort people are putting in to just get the python server working; asking them to then further complicate the setup by putting a reverse proxy with ssl/auth in front of that wouldn't be wise.

jnovack:
>  as self-contained as possible

You know what I'm going to say. ;)

`docker run --rm -it --link pogomap wernight/ngrok ngrok http pogomap:5000 -log stdout`

Dredalious:
It looks like introducing a lot of complexity, while only needed for users wanting to run this publicly (on a server). These users should already know how to configure ssl termination on their preferred proxy (nginx/apache/haproxy/...).

Just adding a default key+cert for everyone (it looks like that is what happens now) gives a false sense of security and won't work for people wanting to make it available over the big bad internet. Just adding the hazzle of requesting own cer, configure it, etc. Personally, I'll keep my HAProxy with SSL termination, I don't like my keys being all over the place anyway.

chuyskywalker:
> Just adding a default key+cert for everyone (it looks like that is what happens now) gives a false sense of security 

Feature does not do that at all.

Again, y'all are clearly very capable, technical users. A very large majority of those who want to use this are not. That said, @jnovack 's ngrok thing really does seem to solve the problem very nicely, and I'm a big fan of "install docker, run one command, done" deals. There are some other fixes wrapped up in this PR I will extract, and turn that suggestion above into a wiki/documentation instead.

Dredalious:
On some closer inspection you're right, my bad, I saw some reference to ssl.crt/.key but your not shipping them. I'm agreeing that @jnovack solution solves the problems (in case that wasn't clear).

GL with the fixes!

jnovack:
If I have to say `docker` one more time around here, I swear...

:)

relrod:
/me never hopped on the Docker bandwagon.

Issue #1493, opened 2016-07-23
Mobile device - Notifications make newly appeared pokemon un-selectable.


Sp4rkR4t:
## Expected Behavior

You should be able to still click the pokemon on the map to get directions

## Current Behavior

If notifications are enabled the pokemon the notification tells you about is un-clickable.

## Steps to Reproduce (for bugs)
1. Enable notifications for a pokemon type (we used eevee)
2. When an eevee appears try clicking on it
3. nothing will happen

## Context

While the notification system is a nice idea it's current implementation causes this issue and is a little redundant, clicking the notification itself simply opens another copy of the map a better use of that space would be to open maps for directions to said pokemon.

## Your Environment

* Android 6.01 & 7 running chrome stable and dev versions

Teleyks:
I cant seem to access ipv4 address from my android, any idea why?

MRokas:
@Teleyks - you should join discord #help chatroom for issues like this.

andrewda:
@MRokas #1605 is the new version of that PR to work on the newest version of the `develop` branch.

Sp4rkR4t:
Checked the latest dev branch and this issue still persists, if you want to be notified of a pokemon, the notifications do not appear and you cannot select any of the selected pokemon types as they spawn on the map, you also get some graphical artifacts displaying around the pokemon icon for the ones you cannot select.

![screenshot_20160724-185005](https://cloud.githubusercontent.com/assets/3926906/17085448/302cd5c8-51d0-11e6-941d-2321e036c685.png)

PR #1831, opened 2016-07-24
Add bounce animation to markers for notified pokemans

jjbott:
## Description
Added a bounce animation to Pokemon markers that we've created notifications for.

## Motivation and Context
It can be hard to find the Pokemons that you were just notified about. This makes it really simple

## How Has This Been Tested?

Watched the map after getting notifications

## Screenshots (if appropriate):
bouncing pokemans

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1941, opened 2016-07-25
Fix clicking "notified" Pokemon and bounce animation on mobile

jjbott:
## Description

Chrome on mobile wants us to use ServiceWorkerRegistration.showNotification() for notifications, and will throw an exception if we try to create a Notification object. This exception is isn't caught, which causes our click handler to fail.

Fixed to catch the exception (and log it in case anyone is watching), and also try to avoid sending notifications for browsers that don't support them at all.

Does *not* fix notifications on mobile. That's going to take more work.

Some tab->space fixes snuck in as well by accident, but it looks like that's the style we want, so I left them.

## Motivation and Context

Fixes https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1493 . Also fixes bounce animation on mobile (from comment in https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1831)

## How Has This Been Tested?

Tested on Chrome in Windows and Android. Added/remove Pokemon from notification list, verified that (on Windows) notifications worked and all Pokemon were still clickable.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -954,25 +954,38 @@ function getPointDistance(pointA, pointB) {
 }
 
 function sendNotification(title, text, icon, lat, lng) {
-  if (!("Notification" in window)) {
-    return false; // Notifications are not present in browser
+  try
+  {
+    if (!("Notification" in window)) {
+      return false; // Notifications are not present in browser
+    }
+
+    if (Notification.permission !== "granted") {
+      Notification.requestPermission();
+    } else {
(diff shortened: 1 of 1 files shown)


fmavdb:
It is worth noting that to implement ServiceWorkerRegistration.showNotification(), a Service Worker needs to be registered. To register a Service Worker, the host needs to use the HTTPS protocol.

PR #1961, opened 2016-07-25
Added "Notify of all Pokémon"

sumihiro:
Added an option to notify when found all of Pokemon (excepts "Hide common Pokémon").

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -104,7 +104,7 @@ function initMap() {
               'style_pgo_nl']
         },
     });
-	
+
 	var style_NoLabels = new google.maps.StyledMapType(noLabelsStyle, {name: "No Labels"});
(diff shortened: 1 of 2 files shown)


sp90:
@sumihiro maybe even add support for browser notifications

https://developer.mozilla.org/en/docs/Web/API/notification

sumihiro:
@sp90 browser notification is already supported.

jjbott:
Just a thought... How about the switch changes the "notify" list from a whitelist to a blacklist? Then you get your functionality with an empty notify list, plus users that care about all but a few can more easily exclude those few from notifications

sumihiro:
I want to get notification of Pokemon which I don't have it.
I add my Pokemon to "Hide common Pokemon", I will get notification of other Pokemon.

michaellindahl:
Related to #1555

PR #1872, opened 2016-07-24
Reduce the number of logged lines

parkershepherd:
## Motivation and Context
4 lines are logged for every single search, requiring lots of scrolling if you're looking through logs, and filling up log files if being run on a server. I combined the 3 model messages into one line, and set the previous 4 log statements as `debug` so they're still available to anyone who needs them.

## How Has This Been Tested?
Ran locally, only affects logging

## Screenshots (if appropriate):
screenshot at jul 24 12-58-07

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1966, opened 2016-07-25
Use created variables for logging

bobvandevijver:
Simple use the created variables for the output, small performance optimization.


Yossi:
This is a trivial fix for a dumb oversight back in #1872 

The code works either way, but it makes more sense this way.

PR #1982, opened 2016-07-25
Ability to display only Lure Pokestops (AHAAAAAAA#1954)

georgo:
## Description
Added switcher to allow display only lured Pokestops

There is one thing I'm not sure about. It's loading of pokestops in raw_data, please verify it.

## How Has This Been Tested?
Tested on Chrome and iOS 9.x (iPhone 6)

screenshot_20160725_122119

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -782,7 +782,6 @@ function processPokestops(i, item) {
         item2 = map_data.pokestops[item.pokestop_id];
         if (!!item.lure_expiration != !!item2.lure_expiration || item.active_pokemon_id != item2.active_pokemon_id) {
             item2.marker.setMap(null);
-            item.marker.setMap(null);
             item.marker = setupPokestopMarker(item);
             map_data.pokestops[item.pokestop_id] = item;
         }
--- a/static/map.js
+++ b/static/map.js
@@ -134,7 +134,7 @@ var StoreOptions = {
         default: true,
         type: StoreTypes.Boolean
     },
-    showLuredPokestops: {
+    showLuredPokestopsOnly: {
         default: 0,
         type: StoreTypes.Number
     },
@@ -318,7 +318,7 @@ function initSidebar() {
     $('#gyms-switch').prop('checked', Store.get('showGyms'));
     $('#pokemon-switch').prop('checked', Store.get('showPokemon'));
     $('#pokestops-switch').prop('checked', Store.get('showPokestops'));
-    $('#lured-pokestops-switch').val(Store.get('showLuredPokestops'));
+    $('#lured-pokestops-only-switch').val(Store.get('showLuredPokestopsOnly'));
     $('#geoloc-switch').prop('checked', Store.get('geoLocate'));
     $('#scanned-switch').prop('checked', Store.get('showScanned'));
     $('#sound-switch').prop('checked', Store.get('playSound'));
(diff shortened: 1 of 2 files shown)


sp90:
Looks really great

invisiblek:
I'm not sure I'm on board with the UX here. Does disabling lured pokestops, but leaving regular pokestops show only non-lured pokestops? Should it?

Would spawning a second option upon turning on pokestops to hide non-lured pokestops make more sense?

(Just trying to drum up some conversation on this, it feels off to me)

Sp4rkR4t:
Personally I feel changing the Pokestops options to s drop down with Off, Only Lured & All as the option would be the better way to go with this.

AHAAAAAAA:
I believe a checkbox under Show Pokestops titled 'Only Lured' would work best. It should be an extension of Show Pokestops, not its own toggle, having 2 toggles is confusing.

Sp4rkR4t:
Checkbox would be fine, I just thought since there are already drop downs in the filter menu it would make sense to keep to that.

W1ndst0rm:
I think I agree with @Sp4rkR4t on this. Turning it into a drop down keeps the interface from getting overwhelmed and cluttered with toggles and checkboxes. Just my 2 cents.

georgo:
Let me change it for ya.

georgo:
Code is now ready to review. Switcher is now dropdown with 2 options, special function to display only lured pokestops is now removed.

screenshot_20160726_084914

static/map.js
```diff
@@ -748,16 +753,26 @@ function processPokestops(i, item) {
...
     }
+    if (Store.get('showLuredPokestops') && !item.lure_expiration) {
+        if (map_data.pokestops[item.pokestop_id] && map_data.pokestops[item.pokestop_id].marker) {
+            map_data.pokestops[item.pokestop_id].marker.setMap(null);
+        }
+        return true;
+    }
     if (map_data.pokestops[item.pokestop_id] == null) { // add marker to map and item to dict
         // add marker to map and item to dict
-        if (item.marker) item.marker.setMap(null);
```
invisiblek: unecessary change for this patch

static/map.js
```diff
@@ -134,6 +134,10 @@ var StoreOptions = {
         default: true,
         type: StoreTypes.Boolean
     },
+    showLuredPokestops: {
```
invisiblek: should this be showOnlyLuredPokestops ?

static/map.js
```diff
@@ -314,6 +318,7 @@ function initSidebar() {
     $('#gyms-switch').prop('checked', Store.get('showGyms'));
     $('#pokemon-switch').prop('checked', Store.get('showPokemon'));
     $('#pokestops-switch').prop('checked', Store.get('showPokestops'));
+    $('#lured-pokestops-switch').val(Store.get('showLuredPokestops'));
```
invisiblek: lured-pokestops-only-switch ?

static/map.js
```diff
@@ -748,16 +753,26 @@ function processPokestops(i, item) {
...
+            item.marker.setMap(null);
+        }
         item.marker = setupPokestopMarker(item);
         map_data.pokestops[item.pokestop_id] = item;
     }
     else {
         item2 = map_data.pokestops[item.pokestop_id];
         if (!!item.lure_expiration != !!item2.lure_expiration || item.active_pokemon_id != item2.active_pokemon_id) {
-            item.marker.setMap(null);
+            if (item.marker) {
```
invisiblek: unecessary change for this patch

georgo: Removed patch.

static/map.js
georgo: Good idea. Applied.

static/map.js
georgo: To preserve convention, I kept showLuredPokestopsOnly.

static/map.js
georgo: Patch removed.

georgo:
Rebased and applied required changes.

invisiblek:
We probably want to hide the drop down if the toggle is off, otherwise this is looking good to me.

georgo:
That shouldn't be problem. Let me do that tomorrow morning ;)

CarlTaylor1989:
+1 tested and works, make the change @invisiblek suggested and good to merge.

AHAAAAAAA:
+1 too, as soon as we get those changes!

georgo:
Display of switcher added. I tried to keep animation, hope you will like it ;)

MonkeyLeeT:
Wait wut animation? @georgo

PR #1986, opened 2016-07-25
Added basic flask HTTP auth

jakubjosef:
## Description
Added basic flask HTTP auth, for securing pokemap. 
Users config in app.py, maybe move user settings to runtime parameters can be done.

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -4,13 +4,19 @@
 import calendar
 from flask import Flask, jsonify, render_template, request
 from flask.json import JSONEncoder
+from flask_httpauth import HTTPBasicAuth
 from datetime import datetime
 from s2sphere import *
 
 from . import config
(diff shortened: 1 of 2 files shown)


relrod:
I'm not a core developer, but: **-1** on this because:

- This should *definitely* be optional, if included at all
- I'm skeptical that this should be included at all. If you need things like this, it's probably time to consider throwing a reverse proxy (apache or nginx) in front of the app server
- The config definitely shouldn't be in `app.py` - configuration is already a bit of a mess (in my opinion...There are two configuration files, plus they can all be overridden by specifying commandline arguments), there's no need to make it even more so.

jakubjosef:
Thank you for your response, i completely agree, but i very busy now, so this is more like PoC for people want to secure your map easily. I've considered a lot about this PR, this is 5 minutes implementation and may be better stay in my fork only.

Issue #1815, opened 2016-07-24
Override config


Whoknowsit:
When starting with `-se`, the config.ini is used. However, i would like to have the ability to override all settings when starting, because currently, i can't seem to start multiple instances using different locations.

`nohup python runserver.py -l "Central Park, New York City, New York, USA" -se`

This fails, if `Location` in `Search_Settings` is empty. But if set, it doesn't allow me to override it.

abighairyspider:
i'd like to see this as well, currently the one way to run multiple instances with mysql is to have a folder of each instance.

confuzzledduck:
The full set of args is passed to parse_config() so it should be possible to check each element is None when reading in the config file. I'll have a fiddle and see if I can throw together a patch.

confuzzledduck:
Quick question: this would require removing some of the defaults from get_args() (as they will always automatically override the config.ini) or checking for None or the defaults when loading from config.ini. Which is preferable?

confuzzledduck:
First pass here: https://github.com/confuzzledduck/PokemonGo-Map/commit/809b0736ad53c3747db0115590f7c87e4f7d3f55

I have gone down the remove defaults route, but happy to change it if people prefer checking for defaults. I think the remove defaults route will require a change to the documentation as some of those settings are IP and port, but I will check this in the morning before submitting the pull request.

Whoknowsit:
Seems to work :)

chuyskywalker:
This one drives me nuts as well. CLI options should *definitely* take precedence over config file options.

confuzzledduck:
Re-implemented default for server and port if not specified on command line or in config file. Means there's no reason to change documentation or make it harder to get the server up and running: https://github.com/confuzzledduck/PokemonGo-Map/commit/b88a2ea14dea904b328450c09aa848714dad27c5

Whoknowsit:
```
# python runserver.py -l "Cologne, Germany" -L de -se
sqlite
2016-07-25 14:19:24,240 [     models] [   INFO] Connecting to local SQLLite database.
/usr/local/lib/python2.7/dist-packages/cffi/model.py:526: UserWarning: 'point_conversion_form_t' has no values explicitly defined; next version will refuse to guess which integer type it is meant to be (unsigned/signed, int/long)
  % self._get_c_name())
2016-07-25 14:19:26,909 [  runserver] [  ERROR] Could not get a position by name, aborting.
```

[...]

PR #1990, opened 2016-07-25
Ability to override config.ini settings with command line switches

confuzzledduck:
## Description
Allows user to specify that the config file should be used (-se) while allowing them to override specific settings therein on the command line at runtime.

## Motivation and Context
Makes spinning up multiple instances with just one difference (ie. a different location) easier by overriding just one element.  See https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1815

## How Has This Been Tested?
Tested with various permutations of command line and config file settings on CentOS 7 implementation. Confirmed working by another (see ticket).

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -112,24 +112,32 @@ def get_args():
 
     args = parse_db_config(args)
 
+    # Read settings from config.ini
     if (args.settings):
         args = parse_config(args)
-    else:
-        if args.only_server:
-            if args.location is None:
-                parser.print_usage()
-                print sys.argv[0] + ': error: arguments -l/--location is required'
-                sys.exit(1);
-        else:
-            if (args.username is None or args.location is None or args.step_limit is None):
-                parser.print_usage()
-                print sys.argv[0] + ': error: arguments -u/--username, -l/--location, -st/--step-limit are required'
-                sys.exit(1);
 
-            if config["PASSWORD"] is None and args.password is None:
-                config["PASSWORD"] = args.password = getpass.getpass()
-            elif args.password is None:
-                args.password = config["PASSWORD"]
+    # Apply defaults if not in config or on cmd line
+    if args.host is None:
+        args.host = '127.0.0.1'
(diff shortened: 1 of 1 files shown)


pogom/utils.py
```diff
@@ -37,19 +37,33 @@ def parse_config(args):
...
+    if args.no_gyms is None:
+        args.no_gyms = Config.getboolean('Search_Settings', 'Disable_Gyms')
+    if args.host is None:
+        args.host = Config.get('Misc', 'Host')
+    if args.port is None:
+        args.port = Config.get('Misc', 'Port')
+
+    if args.gmaps_key is None:
+        if Config.get('Misc', 'Google_Maps_API_Key'):
+            args.gmaps_key = Config.get('Misc', 'Google_Maps_API_Key')
```
efiku: I'm not 100% sure about this.
Why?
Because right now script uses  two sources to get data. 
Does We really need app with args and config.ini? 
What if it will be only  the one parameter called -c / --config config_file.ini
Example: 
```bash
sudo -H pip install --upgrade -r req....txt
cp config/config.ini.dist config/configMap_1.ini
vi config/configMap_1.ini  # customize  config for your map 
python ./run.py -c configMap_1.ini 
python ./run.py -c configMap_2.ini 
python ./run.py -c configMap_3.ini 

```
Probably BC's:
 * All :smile: 

Better change is to remove from code file *credentials.json.example*

confuzzledduck:
It's somewhat helpful to have one config file with username, password, etc. and then be able to spin up different locations from the command line. This helps greatly in case you want dynamically-created maps (from user requests, etc.).

I do agree that we should be able to specify a config file name on the command line. Even more so I also agree that there shouldn't be _yet another_ way to specify the gmaps_key (config/credentials.json).

confuzzledduck:
This appears to be superseded by https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2021.

Issue #1889, opened 2016-07-24
Deadlock when upsert


jangrewe:
Running the latest commit from #1821 , one of my workers threw this:
```
24.7.2016 22:18:192016-07-24 20:18:19,530 [     search] [   INFO] Search queue depth is: 100
24.7.2016 22:18:192016-07-24 20:18:19,875 [     models] [   INFO] Upserting 103 pokestops
24.7.2016 22:18:192016-07-24 20:18:19,889 [     models] [   INFO] Upserting 6 gyms
24.7.2016 22:18:202016-07-24 20:18:20,892 [     search] [   INFO] Search queue depth is: 99
24.7.2016 22:18:212016-07-24 20:18:21,308 [     models] [   INFO] Upserting 65 pokestops
24.7.2016 22:18:21Exception in thread search_thread 0:
24.7.2016 22:18:21Traceback (most recent call last):
24.7.2016 22:18:21  File "/usr/local/lib/python2.7/threading.py", line 801, in __bootstrap_inner
24.7.2016 22:18:21    self.run()
24.7.2016 22:18:21  File "/usr/local/lib/python2.7/threading.py", line 754, in run
24.7.2016 22:18:21    self.__target(*self.__args, **self.__kwargs)
24.7.2016 22:18:21  File "/opt/pogom/search.py", line 124, in search_thread
24.7.2016 22:18:21    parse_map(response_dict, i, step, step_location)
24.7.2016 22:18:21  File "/opt/pogom/models.py", line 278, in parse_map
24.7.2016 22:18:21    bulk_upsert(Pokestop, pokestops)
24.7.2016 22:18:21  File "/opt/pogom/models.py", line 301, in bulk_upsert
24.7.2016 22:18:21    InsertQuery(cls, rows=data.values()[i:min(i+step, num_rows)]).upsert().execute()
24.7.2016 22:18:21  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3244, in execute
24.7.2016 22:18:21    cursor = self._execute()
24.7.2016 22:18:21  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 2656, in _execute
24.7.2016 22:18:21    return self.database.execute_sql(sql, params, self.require_commit)
24.7.2016 22:18:21  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3492, in execute_sql
24.7.2016 22:18:21    self.commit()
[...]

PR #2016, opened 2016-07-25
Database deadlocks fixed

Pedrock:
This is the fix I suggested for issue #1889 which seems to work.

## Description
Changed numbers of simultaneous upserts from 120 to 1.

## How Has This Been Tested?
I have run this for many hours without any problem. Before, when running with several threads, I would get deadlocks after just some minutes, which would stop the program from working correctly.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -308,7 +308,7 @@ def parse_map(map_dict, iteration_num, step, step_location):
 def bulk_upsert(cls, data):
     num_rows = len(data.values())
     i = 0
-    step = 120
+    step = 1
 
     while i < num_rows:
         log.debug("Inserting items {} to {}".format(i, min(i+step, num_rows)))


Dredalious:
What's the impact on performance, especially when running multiple workers? As I see it you now insert row per row (instead of 120 at a time), I'm guessing that on not-so-crowded places this is a good quickfix, but I'm a but unsure how this behaves on crowded places (else, why didn't we do it row per row from the start?)

Pedrock:
@Dredalious The performance should be worse, but what is the point of it being faster if after a while it crashes?
Also, it can always be changed back if someone finds a better solution.

Dredalious:
Allright, but now the guys in charge know that accepting this fix comes at price on performance level.
For me, this fix is OK (I don't use it on super crowded places and I can beef up the server to just work faster if it annoys me, but not everyone is so lucky)

NenadJAustralia:
@Dredalious 

Do you mean with a better performing server I could avoid the deadlocking?

I'm only on a 1 core VPS, would going up help?

Dredalious:
@NenadJAustralia It wasn't what I meant. This PR should fix deadlocking, but will make things a tad bit slower (especially in crowded areas).

If you're seeing deadlocking, I would also suggest to try switching the DB to MySQL as I've heard (https://www.reddit.com/r/pokemongodev/comments/4u6lwe/pokemap_v20_like_the_original_but_waaaaay_better/d5nolzr) sqlite doesn't handle big DB's and lots of queries well, resulting in all kind of shenanigans (deadlocks, 1000+IOPS, ...)

Pedrock:
@Dredalious This problem is occurring in MySQL.
@NenadJAustralia Try my fix. See if performance is a problem.

Dredalious:
:o I have the reverse, occurring in sqlite and not in mysql. (seems to be in line with the experiences on Discord). Ohwell... let's hope your PR fixes everything then and that it gets merged fast.

Pedrock:
@Dredalious It might be happening in sqlite too. Are you really getting deadlocks or just database lock errors?

Dredalious:
@Pedrock I've switched to MySQL and no problem since, with sqlite I restarted all workers each half hour because they hanged, so I'm gonna guess deadlocks.

Pedrock:
@Dredalious How many threads per worker are you using?

Dredalious:
4 threads on each worker, about 10 workers (-st 15)

Pedrock:
I am using only one worker with 25 threads. With 4 threads, the probability of it happening must be pretty low.

invisiblek:
I'm not sure this fixes the actual problem. The way I understand this is you're inserting 1 row at a time as opposed to up to 120, so its less likely that you're going to hit a deadlock, but by no means guaranteed.

It seems to me this fixes the sympton, not the cause.

Pedrock:
@invisiblek I think the problem occurs when two overlapping areas are being updated simultaneously. If one transactions updates row A and B and another updates B and A, a deadlock occurs. If only one row is updated at a time, it can never happen.

invisiblek:
@Pedrock yea, that's what's happening

There's an off change that this could fix it: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2616

I don't think we should move to single requests to the database, that just seems silly to me.

Pedrock:
@invisiblek It's already retrying. In order to fix this without single requests, I believe it's necessary to write a procedure or trigger... That makes using peewee and supporting sqlite more difficult.

Znuff:
If you're getting deadlocks with MySQL, I suggest you go switch to InnoDB as a table engine, instead of MyISAM.

MyISAM does table locking, while InnoDB does row-locking.

invisiblek:
https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2687

PR #2020, opened 2016-07-25
Add MySQL DB details to command line parameters.

clbuttic:
## Description
New command line parameters:

```
  -dh DB_HOST, --db-host DB_HOST
                        MySQL DB Host or IP. Overrides value in config.ini
  -do DB_PORT, --db-port DB_PORT
                        MySQL DB port (usually 3306). Overrides value in
                        config.ini
  -dn DB_NAME, --db-name DB_NAME
                        MySQL DB name. Overrides value in config.ini
  -du DB_USER, --db-user DB_USER
                        MySQL DB User Name. Overrides value in config.ini
  -dp DB_PASS, --db-pass DB_PASS
                        MySQL DB Password. Overrides value in config.ini

```
These override whatever values are set in config.ini

## Motivation and Context

Request in discord channel, and I found it to be useful

## How Has This Been Tested?

Tested with the following tests:

1.  Ran default clean install, which uses SQLite.
2. Ran default clean install, with MySQL details in config.ini.
3. Ran default clean install, with incorrect MySQL details in config.ini and correct details in command line. Verified that the command line parameters overrode those set in config.ini.
4. Ran default clean install, with config.ini set to use SQLite, and specifying MySQL details in above parameters. MySQL server used instead of SQLite.

## Types of changes

## Checklist:

Parameters might be useful to mention in docs.


leTokki:
How about adding a port option?

clbuttic:
@leTokki Is it not inherited with host? host:port as the full, if no host specified, then 3306 is assumed?

leTokki:
Not sure, but for some reason host:port never connected for me, i had to add port='' for it to finally connect

clbuttic:
I'm happy to add it. What would you suggest as the short form? -do ? -dp is taken for --db-password

leTokki:
-do seems good, thanks

clbuttic:
I had to change a few other things to add a port specifier: e461e1e3ef114359aa066f41979655f39f430dcb

Talamar1:
Someone needs to figure out how to handle sockets instead of ports, more secure and faster

clbuttic:
@Talamar1 It's doable, but require connection string as an alternative. Peewee does accomodate this. Will have a look.

clbuttic:
@Talamar1 I believe Peewee uses the unix socket when localhost is the given hostname. Still, a connection string might be useful. I'll leave this for another PR.

Plop:
This is what I want ! Please merge ! :D

invisiblek:
superseded by https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2021

ycraaron:
Cooooooooooool!!!

badroum:
I have seen that there is no argument / config option for the DB Port when you do want to use the non default-one. I searched through pull requests and I landed on [e461e1e](https://github.com/AHAAAAAAA/PokemonGo-Map/commit/e461e1e3ef114359aa066f41979655f39f430dcb?diff=split) where a request had already been sent. Could you tell me if you plan on putting back this option ?  had already been sent and treated

PR #2021, opened 2016-07-25
Config cleanup

invisiblek:
## Description

Kill credentials.json with fire
Match up our command line options with config.ini options
Switch to ConfigArgParser to make the task of overriding
  config.ini options via command line parameters easier
This provides a much easier long-term method of maintaining
  config/argument options

## Motivation and Context

Because I'm tired of this mess

## Types of changes

## Checklist:


invisiblek:
ok, this should be ready for review now (what a mess)

AndyBean:
Is there any wiki or readme updates necessary to accompany this change in config location?

Ashex1:
This greatly improved upon the previous implementation. I had hoped #1579 would get merged in so we wouldn't have to deal with the extra flags but that seems necessary for now. 

When I've got some free time I'll see about automating the parsing of arguments/config so we don't have to add a line for each one.

pogom/utils.py
```diff
@@ -33,90 +30,56 @@ def verify_config_file_exists(filename):
...
-    parser.add_argument('-nk', '--no-pokestops', help='Disables PokeStops from the map (including parsing them into local db)', action='store_true', default=False)
+    parser.add_argument('-t', '--num_threads', help='Number of search threads', type=int, default=DEFAULT_THREADS)
+    parser.add_argument('-np', '--no_pokemon', help='Disables Pokemon from the map (including parsing them into local db)', action='store_true', default=False)
+    parser.add_argument('-ng', '--no_gyms', help='Disables Gyms from the map (including parsing them into local db)', action='store_true', default=False)
+    parser.add_argument('-nk', '--no_pokestops', help='Disables PokeStops from the map (including parsing them into local db)', action='store_true', default=False)
+    parser.add_argument('--db_type', help='Type of database to be used (default: sqlite)', default='sqlite')
+    parser.add_argument('--db_name', help='Name of the database to be used')
+    parser.add_argument('--db_user', help='Username for the database')
+    parser.add_argument('--db_pass', help='Password for the database')
+    parser.add_argument('--db_host', help='IP or hostname for the database')
```
andrewda: dashes are a CLI convention. underscores are typically not used in CLI arguments

andrewda:
:-1: until dashes are added back instead of underscores in the CLI

invisiblek:
i went with underscores so they match config.ini options

Ashex:
We should definitely stay with dashes, if you're trying to keep it consistent for programmatic parsing then just add `dest` to `add_argument` with the desired name.

invisiblek:
Then we have to go with dashes in config.ini as well. (I'm fine with that)

This is how ConfigArgParser works, they need to match.

DerScreever:
maybe add the thread flag.
but nice work. testing this later and merge this after test.
:+1:

JZ6:
+1 good work

clbuttic:
Connection string, to allow any supported pewee DB:

```
diff --git a/pogom/models.py b/pogom/models.py
index a85d8f5..2f14cb8 100755
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -6,6 +6,7 @@ import os
 from peewee import Model, MySQLDatabase, SqliteDatabase, InsertQuery, IntegerField,\
                    CharField, DoubleField, BooleanField, DateTimeField,\
                    OperationalError
+from playhouse.db_url import connect
 from datetime import datetime
 from datetime import timedelta
 from base64 import b64encode
@@ -27,6 +28,10 @@ def init_database():
     if db is not None:
         return db
 
+    if args.db_connection_string:
+        db = connect(args.db_connection_string)
+        return db
+
     print args.db_type
     if args.db_type == 'mysql':
         db = MySQLDatabase(
diff --git a/pogom/utils.py b/pogom/utils.py
index 643ea9b..bd58956 100755
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -109,6 +109,8 @@ def get_args():
     parser.add_argument('-dn', '--db-name', help='MySQL DB name. Overrides value in config.ini', default=False)
     parser.add_argument('-du', '--db-user', help='MySQL DB user name. Overrides value in config.ini', default=False)
     parser.add_argument('-dp', '--db-pass', help='MySQL DB password. Overrides value in config.ini', default=False)
+    parser.add_argument('-ds', '--db-connection-string', help='Peewee DB Connection String (http://docs.peewee-orm.com/en/latest/peewee/playhouse.html#db-url). '
+                        'Overrides all other DB settings.', default=False)
     
     parser.set_defaults(DEBUG=False)
     args = parser.parse_args()

```

I haven't tested it, other than the override for any of the DB settings.

Chlodochar:
Yes, this is very much appreciated. Unifying config file and commandline makes a lot of sense. Tested, old configuration resulted in crash, with the following line:
```
runserver.py: error: Unexpected line 5 in config/config.ini: Username:
```
I think it'll  trigger people to look at the example file again. After removal, all is working well. The code looks good as well.

I do would like to bring up an earlier wish: use database urls(!) instead of separate parameters.

```
sqlite:///myfile.db
mysql://user:password@server/database
postgres://user:password@server/database
```
It makes sense. It is a standard. Heroku uses this as well when you use the MySQL or Postgres addon. Maybe not in scope of this PR.

invisiblek:
yea, adding that sort of option should be handled in a separate commit imo

invisiblek:
/me braces for impact at https://github.com/AHAAAAAAA/PokemonGo-Map/issues

Chlodochar:
Indeed ;)

chuyskywalker:
Neat, does this work with the `-c` option? Can you we finally have multiple instances from one install?

invisiblek:
`    parser.add_argument('-c', '--china', help='Coordinates transformer for China', action='store_true')
`

willgrnr:
Seem to be getting an unknown argument error for `-se`

invisiblek:
@willgrnr its no longer an option. config.ini is parsed no matter what now, with command line options overriding them

willgrnr:
@invisiblek ah ok. I'm assuming uncomment the lines I want to use in config.ini? Either way (commented or uncommented) runserver.py is requiring `-k` and throwing an error.

hemsleyk:
Finally easy external SQL with docker. Thanks! Going to let this shake out for a while before implementing though.

Rikilsta:
TS NOT WORKING WHY??? IT WAS WORKING BEFORE 
runserver.py: errors : arguement -k--gmaps-key is required
I ENTERED MY GMAPS API KEY AND ITS STILL GIVING ME THIS ERROR WHY??? I PUT THE QUOTES LIKE THE GUIDE SAYS AND I EVEN PUT A CREDENTIALS.JSON TO SEE IF IT WORKS.
#GMAPS KEY "xxxxxxxxxxxxxxxxx"

invisiblek:
WHY ARE WE YELLING

Chlodochar:
This'll be a fun thread. *Grabs popcorn*

efiku:
Awesome work @invisiblek !

Ashex:
@efiku I'm pretty sure that issue is unrelated to this PR (the changes to the setup scripts are for the config file). You can submit a PR to resolve that or file an issue so someone can do so :)

efiku:
@Ashex I will submit but actually I don't have my laptop.  :(

Issue #2031, opened 2016-07-25
One Ping Only


jforr:
## Expected Behavior
Add a command line argument to only complete a single scan and then halt scanning when the step limit has been reached.

## Current Behavior
Currently after completing a scan the service will restart scanning from the start.

## Steps to Reproduce (for bugs)
1. Run a scan, wait for it to complete and restart on its own.
2.
3.
4.

## Context
I am trying to populate all stops and gyms at a state level.  I don't want or need up to the minute spawn data, gym control and lured stop data at that large of an area.  This would mainly be for reference and assisting players in finding new hubs to find pokemon.

Sp4rkR4t:
Have to agree that this would be a worthwhile addition to the app.

PR #2032, opened 2016-07-25
Added parameter for single scan

EdenGottlieb:
Added -ss parameter that enables scanning the map once.
## Motivation and Context

#2031 

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -181,7 +181,7 @@ def search(args, i):
 def search_loop(args):
     i = 0
     try:
-        while True:
+        while (not args.single_scan) or i == 0 :
             log.info("Map iteration: {}".format(i))
             search(args, i)
             log.info("Scanning complete.")
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -86,6 +86,7 @@ def get_args():
     parser.add_argument('-ns', '--no-server', help='No-Server Mode. Starts the searcher but not the Webserver.', action='store_true', default=False, dest='no_server')
     parser.add_argument('-os', '--only-server', help='Server-Only Mode. Starts only the Webserver without the searcher.', action='store_true', default=False, dest='only_server')
     parser.add_argument('-fl', '--fixed-location', help='Hides the search bar for use in shared maps.', action='store_true', default=False, dest='fixed_location')
+    parser.add_argument('-ss', '--single-scan', help='Scans the map once then shuts down searcher.', action='store_true', default=False, dest='single_scan')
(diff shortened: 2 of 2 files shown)


JZ6:
but why?

EdenGottlieb:
Because #2031 asked for it. It is mentioned in the "Motivation and Context" section

DeviantEng:
I can see a benefit to this.  My current setup is scanning 5 rural-ish towns in eastern Kentucky.  Very spread out, and if I set the location on the city name with a large number of steps, the scan will take 5-8 minutes to finish which is no good.  Instead I am using `-st 7` with coordinates, so I have to take the time to setup coordinates exactly where I want them to overlap nicely.  

Using `-ss`, I could run a loop and kick off a new scan of each city every 60 seconds.  Would keep things fresh, even though each scan is taking 5-10 minutes to finish.  That would be a 3 fold improvement on my current ~3 minute scan times with a 7 step.  

If you're interested in seeing how I have mine currently set, it's publicly available at https;//ekypokemap.com

ewold:
I would also see the benefit of this. For example i can give the coordinates of my cell phone so the application tracks only the surrounding of my location where i'm currently at.

DeviantEng:
Any movement with getting this merged?

jforr:
I've been testing this and the only issue is that the shell will remain open after the search has been completed.  This is probably expected behavior if a server is also being run off this instance, however if you're using the -ns flag the shell should exit upon completion of the search.

DeviantEng:
I plan on testing this tomorrow, as well.  I'm getting to the point with my current setup (about 35 scans running on different coordinates, 7 steps) isn't able to keep up.  Scans are taking quite a while to finish, and I'm looking to add more locations as it is.  Box is CentOS 7 with 8 CPU cores and 8GB RAM, and the CPU is definitely the limiter.  

Anyway, moving to city names for my locations and a larger radius, should work out nicely with a single scan.  I can have a script for each city that would kick off a scan, wait 60 seconds and kick off another.  Continuous loop at that point.  As long as each scan is finishing, nothing should ever be out of date more than 60 seconds.  

Also considering a multi-server setup to distribute the load (I have lots of resources sitting unused with Cloud at Cost, so I'm using them for a regional map for many users).

jforr:
I've done some logging and testing on my own and the thread is getting locked at the line of code below.

`i, step_location, step, lock = q.get()`

AHAAAAAAA:
Difficult to rework this into current project, but I opened an issue #3294

PR #2034, opened 2016-07-25
removed app.min.js and app.min.css as they are gitignored

Chlodochar:
## Description
app.min.* are generated by Grunt. Checking these into git is annoying especially with multiple PRs: everything is always conflicting. The files are already in gitignore, but weren't removed from git yet.

## Types of changes

## Checklist:


andrewda:
:+1:

Chlodochar:
Do note that this will require people to run `npm install` as well as `grunt`. It is annoying, but conflicting  dist files is more annoying. The build steps do need to be documented in the wiki.

andrewda:
Honestly, I don't even know why we have grunt. The benefit is minimal in a non-production app such as this. If we are going to have it (for whatever reason), though, you're right that the users should generate the mins themselves.

jjiko:
agree with @andrewda. Why are we using a compiler for such minimal benefit?

Chlodochar:
@andrewda I agree. That wasn't really my focus though. I'm just trying to make doing PRs more pleasant and less rebasing after every merge ;).

That said, it will probably be handy to have something like grunt once we need components in javascript. Seeing how map.js grows and where each feature is scattered over the file, I'm guessing it might be useful to split it up in components fairly soon.

brycedorn:
+1 @jjiko, at the very least should be using [flask-assets](https://flask-assets.readthedocs.io/en/latest/) rather than needing a `watch` task running in the background at all times.

I agree it's out of the scope of this PR though - I may make it in a bit.

bsphil:
The change makes plenty of sense but without meaningful error reporting and an explanation of how to properly generate the files, they shouldn't just be removed.

Chlodochar:
It was added to the wiki. I'm not sure what else to do in this case?

hemsleyk:
Wouldn't this render the Dockerfile inert  .... ?

I'm fine with removing minified stuff but please add the relevant steps to the Dockerfile if you're going to make a PR.

Chlodochar:
If the steps are added to the dockerfile, will this be ok to merge?

hemsleyk:
My objection is eliminated as long as the docker image can still be successfully built but I can't speak for the repo owner and those w/ merge rights.

Chlodochar:
See #2384

PR #2040, opened 2016-07-25
Doesn't hammer the server when map download fails.

aedm:
## Description
Waits a little when the map download request fails.

## Motivation and Context
Right now this client hammers Pokemon Go servers, which isn't very nice of us.

## How Has This Been Tested?
Run the client when the a PGo server is down.
Without this patch, watch as 'Map Download failed. Trying again.' messages flood your screen sometimes.
Tested on Linux, Ubuntu.

After patch: http://i.imgur.com/4XQCWjC.png
Before is the same, only there's a hundreds of messages per second.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -132,6 +132,7 @@ def search_thread(args):
                         response_dict = {}
             else:
                 log.info('Map Download failed. Trying again.')
+                time.sleep(config['REQ_SLEEP'])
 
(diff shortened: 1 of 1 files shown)


JZ6:
+1

MonkeyLeeT:
lgtm. Less message printed.

PR #2043, opened 2016-07-25
Fix order of loading map.js and the map api

invisiblek:
## Types of changes

## Checklist:

* Let's load map.js first. The callback for the api requires initMap
  to be defined or things break.
* Furthermore, defer, but don't async, the api call so we're sure
  everything is ready to go before it fires.

### DIFF
--- a/templates/map.html
+++ b/templates/map.html
@@ -173,8 +173,8 @@
 		var center_lng = {{lng}};
 	</script>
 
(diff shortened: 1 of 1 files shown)


DerScreever:
tested
:+1:

Chlodochar:
What? #1869

invisiblek:
heh, never noticed that commit. sorry 😅

Issue #2003, opened 2016-07-25
Cannot read property 'setMap' of undefined


Racle:
## Expected Behavior

Page should loading data from server

## Current Behavior

Script hangs and I have to reload web page. Currently not seeing this happening on mobile because I can't see console on Android chrome.

## Possible Solution

for lines in map.js (near line 633)

```
        item2 = map_data.pokestops[item.pokestop_id];
        if (!!item.lure_expiration != !!item2.lure_expiration || item.active_pokemon_id != item2.active_pokemon_id) {
          item.marker.setMap(null);
            item.marker = setupPokestopMarker(item);
            map_data.pokestops[item.pokestop_id] = item;
        }
```

change ` item.marker.setMap(null);` to `if (item.marker) item.marker.setMap(null);`

Quick testing shows that it doesn't hang script. Don't know if there is other side effects to this.

## Steps to Reproduce (for bugs)

1. Open map
2. Change location (not sure if required)
3.
4.

## Your Environment

* Version used: Latest develop version server
* Environment name and version (e.g. Python 2.7): Python 2.7
* Operating System and version (desktop or mobile): Chrome on Windows 10.

Racle:
My solution might not work. Seems that it leaves pokemons to map or set their remaining time to very high (not correct value)

parkershepherd:
I've been seeing this as well: 
![image](https://cloud.githubusercontent.com/assets/3112960/17105489/aa1df938-5244-11e6-9381-2dc882b4607a.png)

It also seems to halt the auto-refresh when it occurs for me

danielhampton:
It's todo with lured pokemon, just change

`if (!!item.lure_expiration != !!item2.lure_expiration || item.active_pokemon_id != item2.active_pokemon_id) {
            item.marker.setMap(null);
            item.marker = setupPokestopMarker(item);
            map_data.pokestops[item.pokestop_id] = item;
        }`

to

`if (!!item.lure_expiration != !!item2.lure_expiration || item.active_pokemon_id != item2.active_pokemon_id) {
            **if (item.marker) item.marker.setMap(null);**
            item.marker = setupPokestopMarker(item);
            map_data.pokestops[item.pokestop_id] = item;
        }`

It will stop the error being thrown and will not halt the map

parkershepherd:
I've got that in a PR: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2053

PR #2053, opened 2016-07-25
Fix "cannot read property setMap of undefined"

parkershepherd:
## Description
Updated variable names, added comments, and added a null check to a bit of code that has been throwing errors for me

## Motivation and Context
I'm still not 100% sure what causes this to happen, as it is pretty infrequent, but when it does get thrown, it stops `updateMap` from being called as the `rawDataIsLoading` variable never gets reset to false. I tried to rename the variables as I understood them, and added a null check where the error was being thrown.
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2003

## How Has This Been Tested?
Chrome/OSX, Firefox/OSX

Should just be a pretty harmless bugfix.

## Types of changes

## Checklist:


BillyWM:
This fixes the symptom - accessing a property on a non-existent object - but hides the root cause. As you said, 

>I'm still not 100% sure what causes this to happen

parkershepherd:
Agreed. I figured "some pokestops behave funny" might be better behavior than "app silently crashes in the console and stops updating" though. I could add in a console.warn perhaps?

BillyWM:
I just think there's not a huge hurry to sweep this under the rug. It hasn't always been happening, to my recollection, so there was a bad change(s) somewhere that started this.

Can you think of anything that makes this more likely to happen? I'd like to spend some time in the debugger trying to catch this.

parkershepherd:
I can pretty reliably trigger it on my machine by simply zooming out and waiting for a minute or two. (I had a scan size of 9+ I think). I would vote for pushing this (maybe with a console warning) so as to avoid the worse bug this is causing. Code wise, it just looks like the null check was missed. I haven't seen anything weird looking since using this fix myself

parkershepherd:
I did just notice this: 
image though @BillyWM

BillyWM:
Okay, can you add a console.warn so this isn't forgotten?

Re: the image, sorry I'm not sure what I'm looking at.

parkershepherd:
@BillyWM It was a double pokestop. I added the warning. I know there were some existing bugs with pokestops, but the PR that fixed them got reverted.

This should be good to go for now though

BillyWM:
Marker management is a bit more convoluted than I realize, so nevermind; under the current system, sprinkling those null checks around is necessary. It could use a refactor but you're not doing anything wrong, so my bad.

parkershepherd:
Bug appears to be fixed with a previous commit

Issue #1806, opened 2016-07-24
Code.


efiku:
Sorry guys but I want to tell something. I don't know if this is right place.
I hope everyone will understand. :stuck_out_tongue: 

___

This project is great  (forks and stars in such a short time) but in my opinion there is some **mess** in code and I want to point some "bugs": 

- Unused code and  two different variable naming conventions  [LINK](https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/pogom/customLog.py#L18L25) 
- Variables don't say what they are.  [LINK](https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/pogom/transform.py#L3L4)  [LINK2](https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/pogom/app.py#L46) 
- I suggest consider   DBAL + ORM  option for python.  #1618  [LINK](https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/pogom/models.py#L31L40) 
-  [LINK](https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/pogom/models.py#L26) Global variables == problem in feature?

Please take care about code health. :) 

### Also I suggest

-  Create wiki docs about API served by backend. *This will help new contributors with front skills*
-  Backend documentation 
-  Create CONTRIBUTING.md file with rules about code style / Pull request / Issues etc.
-  Just a nice REST app to decouple all UI from backend. (Separating application layers ) ( +1  for  HTML/CSS/JS masters ) ( +1 for server health ) (+1 for code quality)  (format: json) 

Of course this is only my suggestions, you do not have to take any attention.
PokemonGo-Map works fine on Linux Debian  ;)

ping @AHAAAAAAA

sykhro:
This is why I wanted everyone to do core review on a weekly basis and accept only clean PRs

MRokas:
> a = 6378245.0
This is length around earth at equator. It should be pretty common knowledge tbh.

> ee = 0.00669342162296594323
cant remember this one atm, but is also widely known in map/gis

Pedrock:
I think you should make people squash/merge their commits where appropriate.

houkensjtu:
Is there any user forum or communication platform to discuss the tech detail about this project?

MRokas:
@houkensjtu - Discord chat.

houkensjtu:
[...]

PR #2058, opened 2016-07-25
Use explicit styling for Javascript

Chlodochar:
This PR partly follows #2748 (Added Javascript styling rules).

## Description
At the moment the JS files were a bit of a mess:

* Mixed tabs and spaces (on same line! My eyes!)
* Mixed naming conventions (camel-case, usage of _, etc)
* Different usage of spaces before and after keywords
* Undefined variables (... and implicitly define them in `window`)

To circumvent this, a style guide needs to be chosen or created. To get things going, I chose [StandardJS](http://standardjs.com/). I don't care which style-guide we'll use, but it has to be explicit, creating a style-guide ourselves is hard and a lot of work. StandardJS seems to have enough tools available to check and format code. So that's why I chose it.

In addition I added eslint configuration, so that javascript can be checked. Lastly I tried to automatically format map.js and app.js. I've made a script to do this, so rebases are still doable. There are still a number of open issues that eslint reports, but that can be fixed after the general style is applied (manual work, not something I'll like to rebase).

See the resulting map.js here:
https://github.com/Chlodochar/PokemonGo-Map/blob/pr-jsstyle/static/map.js

## Motivation and Context
Code style was a mess. To get a unified code-base, some style-guide needs to be used.

## How Has This Been Tested?
Go through some of the features in the webinterface.

## Types of changes

## Checklist:

- [ ? ] My code follows the code style of this project. (which one?!)

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -748,7 +748,7 @@ var Store = {
   getOption: function (key) {
     var option = StoreOptions[key]
     if (!option) {
-      throw 'Store key was not defined ' + key
+      throw new Error('Store key was not defined ' + key)
     }
     return option
   },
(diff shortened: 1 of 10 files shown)


Chlodochar:
Related to #1806. This will also allow usage of Travis or SideCI.

CarlTaylor1989:
I'm getting issues with this. Can you have a look?

screen shot 2016-07-26 at 23 11 00

Chlodochar:
@CarlTaylor1989 Could you give this another try?

Chlodochar:
I've now only ran only safe changes on the code. At the moment it will not rename variables. That does leave some errors left, but those can be changed manually one the general styling of the code is ok.

invisiblek:
✖ 230 problems (230 errors, 0 warnings)

Are you still working on this? Otherwise I'll jump on it and finish it up

Chlodochar:
Ah thanks. I'll work on the leftover problems.

static/js/map.js
```diff
@@ -1,233 +1,804 @@
...
+  'stylers': [{
+    'visibility': 'off'
+  }]
+}, {
+  'featureType': 'all',
+  'elementType': 'labels.icon',
+  'stylers': [{
+    'visibility': 'off'
+  }]
+}]
```
andrewda: Might want these to go in separate files or stay one-liners. I don't think the styles are a great use of about 600 lines.

Chlodochar: agreed. however, we also do not want lines that are 1000+ characters long.

Chlodochar:
No linting errors left. All features that I've found seem to be working like they should. There is still a problem in `/stats` relating to jQuery Accordion not being included, but that was incorrect before this PR.

PR #2017, opened 2016-07-25
Bugfix: Skip notification when merely redrawing existing pokemon

parkershepherd:
Notifications were being triggered on every pokemon redraw, so I added an option to skip notifications if only a redraw was happening

## Description
Just added the `skipNotification` option for `setupPokemonMarker` to prevent notifications from being triggered on zoom alone

## Motivation and Context
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1997

## How Has This Been Tested?
Chrome/OSX, Firefox/OSX

Only `redrawPokemon` is affected (triggered on zoom)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1979, opened 2016-07-25
Added bounce animation end on click

MikeDuister:
## Description
3 lines added that makes sure the animation ends when the marker is clicked

## Motivation and Context
Prevents continuous loop of bouncing pokémon ( #1955 )

## How Has This Been Tested?
By running the setup several times on Linux and Windows.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #1955, opened 2016-07-25
Stop notified icon bounce after given time


FabianSperrle:
Since #1831 map icons of notified Pokémon bounce, which is nice. Nonetheless I believe that the bouncing should stop eventually, before the 15 minutes run out.

## Expected Behavior

- I get notified of a new Pokémon and click the notification. The map shows the bouncing Pokémon for another 60 seconds, then the bouncing stops.
- I don't react when I get the notification, the bouncing stops after 120 seconds.

## Current Behavior

Bouncing does not stop until the Pokémon is removed from the map.

## Steps to Reproduce (for bugs)

1. Load map, add a Pokémon to be notified about
2. Wait for it to appear
3. Icon never stops bouncing

## Context

You can't always run to all Pokémon, the remaining bounce animation is distracting and not helpful. 

## Your Environment

* Version used: Latest dev #d7aa801
* Environment name and version (e.g. Python 2.7): Python 2.7
* Operating System and version (desktop or mobile): Heroku dyno, affects all end user devices

MikeDuister:
Made a PR in which the bounce animation stops after 3000ms (= 4 bounces)

PR #2076, opened 2016-07-25
Bounce ends when marker is clicked

MikeDuister:
## Description
Bounce animation stops when marker is clicked. 

## Motivation and Context
Prevents continuous loop of bouncing pokémon ( #1955 )

**rework of #1979 based on improvements in #2017**

## How Has This Been Tested?
By running the setup several times on Linux and Windows.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -497,13 +497,18 @@ function getGoogleSprite(index, sprite, display_height) {
     };
 }
 
-function setupPokemonMarker(item, skipNotification) {
+function setupPokemonMarker(item, skipNotification, isBounceDisabled) {
 
     // Scale icon size up with the map exponentially
     var icon_size = 2 + (map.getZoom()-3) * (map.getZoom()-3) * .2 + Store.get('iconSizeModifier');
     var pokemon_index = item.pokemon_id - 1;
     var sprite = pokemon_sprites[Store.get('pokemonIcons')] || pokemon_sprites['highres']
     var icon = getGoogleSprite(pokemon_index, sprite, icon_size);
+	
+	var animationDisabled = false;
+	if(isBounceDisabled == true){
+		animationDisabled = true;
+	}
 
     var marker = new google.maps.Marker({
         position: {
@@ -514,7 +519,13 @@ function setupPokemonMarker(item, skipNotification) {
         optimized: false,
         map: map,
         icon: icon,
+		animationDisabled: animationDisabled,
     });
+	
+	marker.addListener('click', function() {
+		this.setAnimation(null);
+		this.animationDisabled = true;
+	});
 
     marker.infoWindow = new google.maps.InfoWindow({
         content: pokemonLabel(item.pokemon_name, item.disappear_time, item.pokemon_id, item.latitude, item.longitude, item.encounter_id),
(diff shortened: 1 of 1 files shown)


MikeDuister:
# request
Late night variable naming might not have been the best idea, please report if variables need to be renamed.

clbuttic:
Why not time them out? 

```
            setTimeout(function() {
                marker.setAnimation(null);
            }, 4000);
```

louispires:
@clbuttic Please see comments here: #1979

Basically timing it out will be useless for some people as there is no certain way of knowing that the user has noticed the bouncing animation.

static/map.js
```diff
@@ -497,13 +497,18 @@ function getGoogleSprite(index, sprite, display_height) {
...
-function setupPokemonMarker(item, skipNotification) {
+function setupPokemonMarker(item, skipNotification, isBounceDisabled) {
 
     // Scale icon size up with the map exponentially
     var icon_size = 2 + (map.getZoom()-3) * (map.getZoom()-3) * .2 + Store.get('iconSizeModifier');
     var pokemon_index = item.pokemon_id - 1;
     var sprite = pokemon_sprites[Store.get('pokemonIcons')] || pokemon_sprites['highres']
     var icon = getGoogleSprite(pokemon_index, sprite, icon_size);
+	
+	var animationDisabled = false;
```
invisiblek: tabs vs spaces

clbuttic:
@louispires I see. Is it not going to be too annoying? Wouldn't it be better to show some other indicator that also incorporates time left? Either fading the opacity, or a spinner.

MikeDuister:
@clbuttic Notifications about pokémon are meant for rare pokémon that the person is willing to travel for. With the default radius this means that about 5 pokémon will be bouncing at max.

But even if the animation needs to be changed, it will be in another separate PR. This issue stand aside from this fix.

static/map.js
```diff
@@ -514,7 +519,13 @@ function setupPokemonMarker(item, skipNotification) {
         optimized: false,
         map: map,
         icon: icon,
+		animationDisabled: animationDisabled,
```
invisiblek: animationDisabled: isBounceDisabled

static/map.js
invisiblek: this variable is useless, just use the parameter passed in (see comment below)

static/map.js
```diff
@@ -860,8 +872,12 @@ function updateMap() {
 function redrawPokemon(pokemon_list) {
     var skipNotification = true;
     $.each(pokemon_list, function(key, value) {
+		var isBounceDisabled = false;
+		if(this.marker.animationDisabled != false){
+			isBounceDisabled = true;
+		};
         var item =  pokemon_list[key];
-        var new_marker = setupPokemonMarker(item, skipNotification);
+        var new_marker = setupPokemonMarker(item, skipNotification, isBounceDisabled);
```
invisiblek: can you just send this.marker.animationDisabled? no need to define more variables for all this stuff imo

MikeDuister:
@invisiblek Ff the default isn't defined by this variable it will return undefined at first run, isn't this an issue?

AHAAAAAAA:
Tested, works +1

Idhanta:
The bounce doesn't stop after I click on the marker. Using safari.

PR #2079, opened 2016-07-25
added python linting using flake8 to travis

Chlodochar:
## Description
This adds flake8 to Travis, which will check for stylistic problems in Python files. I've initially haven't set any options, so it'll output the defaults. We can decide which errors should be disabled once Travis shows them.

## Motivation and Context
We need a styleguide for Python. Flake8 seems to do this well.

## How Has This Been Tested?
Running flake8 locally.

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ env:
   - FLASK=0.11.1
 install:
   - pip install Flask==$FLASK
+  - pip install flake8
   - pip install -r requirements.txt
+before_script:
+  - flake8 .
 script:
   - nosetests


Chlodochar:
Alternatively we can use https://landscape.io, a service like Travis, but specifically for code quality checking. Works well for PRs as well. Here is an example output: https://landscape.io/github/Chlodochar/PokemonGo-Map/3

andrewda:
:+1: for landscape.io

Chlodochar:
@andrewda Yes, it seems more user friendly by giving sane defaults for the check-rules and shows the errors inline.

Chlodochar:
@AHAAAAAAA What do you think? You can add it to your account.

Chlodochar:
Oh, one note though. Landscape only supports Python. Not Javascript.

AHAAAAAAA:
I prefer we keep it all testing in Travis for now. What does everyone else think? Also man did you see those errors? Over 1000 lines of style errors.

Chlodochar:
@AHAAAAAAA The downside is that you do not see what code is actually incorrect. It's all cobbled up as text output. It does give flexibility with what linters we want to use though.
I do have to say, using different services isn't that bad. Githubs PRs support that nicely (shows a list of each service and whether it failed the PR or not). What do you have against using multiple?

AHAAAAAAA:
@Chlodochar valid points. I preferred Travis just to centralize linting/testing/build checks that needed to pass into 1 service, but you make good arguments. +1 for Landscape, I'll set it up

AHAAAAAAA:
Integrated landscape.io, I'll close this repo.

Chlodochar:
Looking good for the branches. It seems the checks for PRs are getting stuck. I've opened an issue on Landscape here: https://github.com/landscapeio/landscape-issues/issues/240

Chlodochar:
@AHAAAAAAA Also once things are working nicely. It might be nice to put a badge up in the readme. That way people might be more motivated to fix those linting issues :D

Issue #2037, opened 2016-07-25
Geolocation option should be unavailable with --fixed-location


silverwind:
## Expected Behavior
Users should not be able to change the location when the server was started with `--fixed-location`.

## Steps to Reproduce
1. Start server with `--fixed-location` or `-fl`
2. Notice that the Geolocation checkbox is available on the web interface

trol4ever:
I'm too lazy to make a pull request and submit the change at the moment. Just apply the same styling to the Geolocation button as the 'Change your location' textfield has.

In a more step by step explanation:
1. Go to templates\map.html
2. Right click it and edit it with a text editor.
3. Go to line 100.
4. You should see `<div class="form-control switch-container">`
5. Change that line to `<div class="form-control switch-container" style="display:{{is_fixed}}">`
6. Save and restart the webserver.

silverwind:
Looks like this is fixed in https://github.com/AHAAAAAAA/PokemonGo-Map/commit/d838ee910439b285786b8aa7e2078c1a1a5de60c.

PR #2081, opened 2016-07-25
Hide Geolocation button and switch when -fl is on

EdenGottlieb:
Geolocation button and switch are now hidden when -fl is on.
## Description

Added display:{{is_fixed}} to the geolocation switch and geolocation button

## Motivation and Context

#2037 

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -926,6 +926,7 @@ function myLocationButton(map, marker) {
     var locationContainer = document.createElement('div');
 
     var locationButton = document.createElement('button');
+    locationButton.style.display = '{{is_fixed}}';
     locationButton.style.backgroundColor = '#fff';
     locationButton.style.border = 'none';
     locationButton.style.outline = 'none';
--- a/templates/map.html
+++ b/templates/map.html
@@ -87,7 +87,7 @@
 				</div>
 			</div>
 
-			<div class="form-control switch-container">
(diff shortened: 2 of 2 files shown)


trol4ever:
You are overcomplicating it. Just leave it as an individual inline style, instead of making it an internal stylesheet that needs to be declared in 2 different places.
"Simply using {{is_fixed}} doesn't work anymore because the geolocation button is generated later on." This is not true.

EdenGottlieb:
@trol4ever You're right... This was my initial attempt. Commit updated.
I couldn't get it working because the browser was caching my file and giving me headaches. 

Currently the best way I found to check JS is simply copying the entire map.js file into a <script> in HTML, nothing else quite works.

trol4ever:
`locationButton.style.display = '{{is_fixed}}';`
What's this for?

Also I think it's pointless. Javascript will never read python vars without an escape sequence.
So if you wanna hide that small "Your location" button based on is_fixed, give an id to either 'next-location' or the geolocation div element prior to the style.
Something like `<div class="form-control switch-container" id="dispType" style="display:{{is_fixed}}" >`
After that, in map.js get the style with something like 
` var dispType = document.getElementById("dispType").style
location.button.style = dispType;` 

Needs testing tho, as I just pulled this outta my head.

PR #2094, opened 2016-07-25
added grunt to travis

Chlodochar:
## Description
Grunt always needs to be run to check and build JS and SASS files to get the webinterface going. Travis needs to do this as well. That is what this PR does.

## Motivation and Context
Make sure Travis goes through the entire setup that other people need to do as well. It might find errors during this setup that can be caught early.

## How Has This Been Tested?
It hasn't, we'll see what Travis does.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -75,6 +75,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-contrib-concat');
   grunt.loadNpmTasks('grunt-html-validation');
 
-  grunt.registerTask('default', ['jshint', 'sass', 'cssmin', 'uglify']);
+  grunt.registerTask('build', ['jshint', 'sass', 'cssmin', 'uglify']);
+  grunt.registerTask('default', ['build', 'watch']);
 
 };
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -75,6 +75,6 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-contrib-concat');
   grunt.loadNpmTasks('grunt-html-validation');
 
-  grunt.registerTask('default', ['jshint', 'sass', 'cssmin', 'uglify', 'watch']);
+  grunt.registerTask('default', ['jshint', 'sass', 'cssmin', 'uglify']);
```
MonkeyLeeT: Hmm, why remove watch?

Chlodochar: @MonkeyLeeT Travis will probably wait indefinitely to build it. Watch should not be default.

bmds:
It would make sense to split it out like this
```js
  grunt.registerTask('build', ['jshint', 'sass', 'cssmin', 'uglify']);
  grunt.registerTask('default', ['build', 'watch']);
```

The Gruntfile is primarily for humans so make the default work the best for humans.
It is a minor thing to then get Travis to run `grunt build`

Chlodochar:
That seems like the best solution. Added.

AHAAAAAAA:
rebasez

Gruntfile.js
invisiblek: but watch is great for developing locally...

it would make using grunt even more painful to remove it =P

Chlodochar:
Ty!

PR #2097, opened 2016-07-25
Fixes ignored --scan-delay switch

aedm:
## Description
1. Fixes --scan-delay argument being ignored.
2. Makes --scan-delay a float value, so values like 0.5 are now accepted.

## Motivation and Context
I wanted to use --scan-delay to make scans faster.

## How Has This Been Tested?
1. Tried "-sd 5" and it was completely ignored before, and worked well with the patch.
2. Tried "-sd 0.5" and it made scanning intervals shorter.

Taken at "-sd 0.5", scan intervals are shorter than the defaul one second. http://i.imgur.com/5sRXfu8.png

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -97,7 +97,7 @@ def login(args, position):
 
     while not api.login(args.auth_service, args.username, args.password):
         log.info('Failed to login to Pokemon Go. Trying again.')
-        time.sleep(config['REQ_SLEEP'])
+        time.sleep(args.scan_delay)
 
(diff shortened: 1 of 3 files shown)


invisiblek:
using 0.5 doesn't quite get it down to 0.5 seconds, it is less than 1 though (i imagine its due to it waiting for the previous response before kicking off the new one)

the last patch most definitely fixes the scan-delay parameter

+1

PR #2099, opened 2016-07-25
map.js: updateMap timer should not be set independently on map instantiation.

DeirhX:
updateMap timer in current code base is set on page load, regardless on how long can asynchronous map instantiation through initMap() actually take. If it takes longer than the currently hardcoded 5000ms on timer tick, it's called with map object still being null and crashes on first statement that tries to access map.

## Description
Moved the timer to be set synchronously after map instantiation inside initMap().

## Motivation and Context
updateMap() was crashing on slow connections / machines on first page load.

## How Has This Been Tested?
Set the timer to 1ms instead of 5000ms and see the problem for yourself. Then move to initMap().
Tested on Chrome.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -276,6 +276,8 @@ function initMap() {
         redrawPokemon(map_data.pokemons);
         redrawPokemon(map_data.lure_pokemons);
     });
+
+    window.setInterval(updateMap, 5000);
 };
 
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
No longer necessary.

PR #2102, opened 2016-07-25
Fix permissions

sfjuocekr:
Fixed some things that were merged and then removed later.


sfjuocekr:
Feck, fixed in next commit.

PR #1899, opened 2016-07-24
Add Rarity and Type info to data set and display rarity

tim95030:
## Description
This feature adds a bunch of data collected from [Silph Road](https://thesilphroad.com/) that adds rarity, pokemon type, and spawn rate. This format also makes it easy to add new fields to this data. Refactored i8ln to use a function to load translations from json files. This should be easily replaced with any i8ln library if so desired, but for now it is a good first step I feel. Added types to pokemon labels as well.

## Motivation and Context
This will add the ability to display the rarity for now. But also allows for future features to be easily added. Filter out all common drops with one click, show only pokemon of a certain type, etc.

## How Has This Been Tested?
I have currently only tested this on MacOS in Chrome, but the only real change is the data files and the display on the map so it seems unlikely to break other things. Extra testing for other locales would be appreciated and adding translations for the new fields is also needed before merging most likely. (Though currently it just displays an empty string for unknown field types like rarity)

## Screenshots (if appropriate):
Normal Pokemon
screen shot 2016-07-27 at 11 19 20 pm

Lured Pokemon
screen shot 2016-07-27 at 11 38 41 pm

Translated - with valid translation
screen shot 2016-07-27 at 11 19 59 pm

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2109, opened 2016-07-25
Show Pokemon types in map popups

UzEE:
## Description
Added the ability to show Pokemon types in the map popups.

## Motivation and Context
It would be very helpful to have the ability to show Pokemon types in their popups, to make it easier to hunt down and track certain types of Pokemon. I've added type labels to Pokemon marker popups, and will also send in another pull request tomorrow that adds the ability to set notifications for certain types.

For example, it would be possible to get an alert notification whenever a Dragon or Fairy type Pokemon spawns in the search area.

## How Has This Been Tested?
This PR only adds a couple of lines of code to the Pokemon marker label code. All the styles are inline and I've tested the code in Chrome (Windows, iOS), Edge (Windows), and Safari (iOS).

image

## Types of changes

## Checklist:


tim95030:
Can we merge this into my pull request? As I have all the type data loaded into the json files.

See #1899

static/map.js
```diff
@@ -352,8 +369,9 @@ function pokemonLabel(name, disappear_time, id, latitude, longitude, encounter_i
             <small>
                 <a href='http://www.pokemon.com/us/pokedex/${id}' target='_blank' title='View in Pokedex'>#${id}</a>
             </small>
+            <span> - </span>
+            <small>` + typeStr + `</small>
         </div>
-        <div>
```
redprince: intended? seems like a typo :D

oceanor: intended? seems like a typo :D

UzEE: Good catch. Must have happened when squashing the commits.

UzEE:
@tim95030 Ah I didn't see your PR before. I wanted to avoid storing types in the JSON files since it wasn't as efficient as packing them into a numeric array, which is smaller in size. Also because I don't know python, I wanted to avoid fiddling around with the server side code.

If everyone feels this would be a better fit in #1899 then, we can tweak the UI code to use the JSON data.

chaze117:
Nice feature, by the way I wrote a rarity script based on your code. Thanks for sharing.

tim95030:
I have integrated this to #1899

UzEE:
Great. I guess this should be closed now then?

Crazy-Drew:
This would be a great feature I think, I like it.

PR #2124, opened 2016-07-26
Only allow one info window at a time.

joshluongo:
## Description
I've made it so you can only open one info window at a time. This also fixes and issue where you can open multiple info windows for the same marker.

I've also added a convince function `clearInfoWindows()` to close all open info windows.

## Motivation and Context
On a phone the screen get's very crowded with multiple info windows open. If the user keeps tapping the marker multiple times then it can be frustrating to close all of the windows.

## How Has This Been Tested?
This is a minor change but I've tested it on iOS and Safari on Mac OS X.

_I can't show the change in a screenshot_

## Types of changes

## Checklist:


brycedorn:
thanks for addressing this! was bugging me 👼 

rather than iterating through everything in `map_data` though, how about adding `info_windows` as a hash in `map_data`? this would be a lot more efficient, and would make checking if a pokemon has an `info_window` already open much easier

static/map.js
```diff
@@ -619,8 +619,37 @@ function clearSelection() {
...
+
+    $.each(map_data.pokestops, function(key, value) {
+        // Close all windows
+        value.marker.infoWindow.close();
+    });
+};
+
 function addListeners(marker) {
     marker.addListener('click', function() {
+    	clearInfoWindows();
```
brycedorn: also indentation slightly off here^ 🔍

joshluongo:
@brycedorn The indentation is because I edited it via the GitHub editor 🙄. I think placing it in a hash map is a good solution. I'll update my branch with that change (and I'll fix the indentation 😉) when I get a chance.

joshluongo:
I'm going to rewrite my changes and submit a new PR.

PR #1948, opened 2016-07-25
show pokemon icons in select2 elements

niklaswa:
## Description
show pokemon icons in the select2 elements (excluded and notify pokemon)

## Motivation and Context
only visual enhancement

## How Has This Been Tested?
ran the server myself

## Screenshots (if appropriate):
Pokémon icons in select element

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2126, opened 2016-07-26
Only show pokémons available in the game

EdenGottlieb:
Hides pokemon that are not available in the game from notify and exclude lists.
## Description

Currently, the notify and exclude lists show all existing pokemons from all gens.
This PR shows only pokemon currently available in the game, plus the legendary ones and ditto that will be added later (all pokemons up to #151). This declutters both the exclude and notify boxes and makes searching for pokemons easier. This is done by limiting the key to 151 while transferring JSON data to the search boxes, in a manner that does not damage the original json lists for future use.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -1052,6 +1052,9 @@ $(function () {
         var pokeList = []
 
         $.each(data, function(key, value) {
+            if (key > 151) {
+            	return false;	
+            }
             pokeList.push( { id: key, text: value + ' - #' + key } );
             idToPokemon[key] = value;
         });


brycedorn:
good consideration, but rather than continuing to iterate through all of `data` (even after the 151st is reached) you can do `if (key > 151){ return false; }` 

also indentation's a little off.

EdenGottlieb:
@brycedorn Switched to your solution, seriously more efficient. I actually tried it yesterday while working on this but for some reason it didn't seem to work, I was simply having mad JS problems...

Also fixed indentation, My computer and GitHub don't seem to interpret spaces and dashes in the same manner

static/map.js
```diff
@@ -1052,6 +1052,8 @@ $(function () {
         var pokeList = []
 
         $.each(data, function(key, value) {
+            if (key<=151)
+            	return false;
```
brycedorn: super minor, but should add brackets around this `if` block to follow convention of the rest of this file

brycedorn:
one last comment then looks 👌!

EdenGottlieb:
@brycedorn Done

przerwap:
If we're to follow convention here, there should be a space before and after the `<=`

EdenGottlieb:
@przerwap Done

static/map.js
```diff
@@ -1052,6 +1052,9 @@ $(function () {
         var pokeList = []
 
         $.each(data, function(key, value) {
+            if (key <= 151) {
```
AHAAAAAAA: return false if it IS first gen? conditional looks wrong.

niklaswa:
Hey, it was already included in #1948 
look here: https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/static/map.js#L1107

static/map.js
EdenGottlieb: Damn. Fixed.

PR #2127, opened 2016-07-26
Added and fixed stuff that was previously added and later removed.

sfjuocekr:
Removed pgoapi.
Fixed lured Pokemon.
Fixed some permissions.


sfjuocekr:
Not sure what travis is complaining about, is it missing libprotoc?

Chlodochar:
Please see #671. This PR unnecessarily does the same thing.

Issue #1935, opened 2016-07-25
Step delay not working


Talamar1:
When setting the wait between loops, it doesn't actually wait:
16-07-25 00:19:10,179 [     models] [   INFO] Upserting 1 gyms
2016-07-25 00:19:10,776 [     search] [   INFO] Search queue depth is: 99
2016-07-25 00:19:10,776 [     search] [   INFO] Scanning complete.
2016-07-25 00:19:10,777 [     search] [   INFO] Waiting 10 seconds before beginning new scan.
2016-07-25 00:19:11,183 [     search] [   INFO] Search queue depth is: 99
2016-07-25 00:19:11,198 [     models] [   INFO] Upserting 2 pokemon
2016-07-25 00

As you can see, it is set to 10 seconds, but it didn't wait.

tkSimon:
This has been introduced by #1835 , i think. Before scan_delay was the sleep between scans, now it is the sleep between initiating the scan. If one full search iteration takes longer than scan_delay there won't be any forward progress.

PR #1451, opened 2016-07-23
Threading

chuyskywalker:
Improve overall thread handling
- Prevent flask from re-running the app causing duplicate search threads 
  - _(Technically this disables hot-reloading of code, if I understand correctly, but that doesn't seem a high cost. Obliterating niantic servers does.)_
- Prevent scan system from recursing to infinity
- Increase the search loop default delay (with the search threading, give Niantic's poor servers breathing room)
- Handle the mock data/search thread duality better
- Make `-ns` mode interruptable

Misc fixes found while working:
- Fix EOL on config.ini
- Ignore PEP8 harder
- Allow enough space for "connectionpool" debug output

## Motivation and Context
Discussion here: https://github.com/AHAAAAAAA/PokemonGo-Map/commit/3bafbfdfd5f219b87a8c127cea433d1a83860cc5

## How Has This Been Tested?
Run under `-ns` and `-m` modes combinations.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation. _**N/A**_
- [x] I have updated the documentation accordingly. _**N/A**_

PR #2128, opened 2016-07-26
Waits for the search queue to complete before next map iteration

adamhwang:
## Description
Joins the search queue back to the producer thread.  Producer thread waits until queue is finished before queuing up the next iteration. 

## Motivation and Context

Addresses bug #1935 - fixes the delay.  Also fixes instantly blasting off 100 / step-count issue (see screen shots).

The drawback however is that this will be slightly more inefficient towards the beginning and tail ends of the map iteration as threads start to finish up and the queue empties out.  I think that should be acceptable though.  With a small search space, multi-threading the same map locations isn't very useful.  With a large search space, the difference will be negligible.

## How Has This Been Tested?

Normal usage (Chrome+Win8.1) didn't instantly crash.

Original with "-st 1" - check timestamps... 0-100 in 1 second:
image

With PR:
image

## Types of changes

## Checklist:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,3 +7,4 @@ Easy\ Setup
 Gruntfile.js
 package.json
 pogom.db
+core
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ config/config.ini
 static/dist/
 node_modules/
 venv/
+core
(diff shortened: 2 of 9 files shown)


adamhwang:
Fixed in #1451.  Closed

Issue #1951, opened 2016-07-25
Multiple notifications for same Pokemon


AHAAAAAAA:
Using the 'Notify of Pokemon', I get sent a notification everytime I:

* refresh
* click on notification
* blink
* apologize
* random

Expected behaviour:
One notification per pokemon

ericpadilla:
I confirmed this is true.

nallsnowblind:
Confirm
It happens when the map reloads
Same is when you zoom in and out again. The pokemon will then be redrawn on the map and notification is sent again.

p4pergoat:
Yeah, this is true. It also spams you with notifications when the pokemon you wanted to get notified for despawns (gets removed from the map).

rjwestman:
it also happens everytime the scanner finds the pokemon - so 3 times for each pokemon (since it uses a triangulation to find its exact location, right?
So basically the notification trigger needs to be moved to where the pokemon is added to the map. Would also be nice to give it a glow of some kind to find it more easily (when scanning a large area)

AHAAAAAAA:
@rjwestman It doesn't use triangulation.

DerScreever:
should be fixed with #2017

parkershepherd:
@DerScreever the fix I made should only apply when you zoom in/out but the pokemon remains in the view the whole time. If you zoom out to a new area or pan to a new area, you will still get notified about pokemon that you could have already been notified about before (although maybe that's the best behavior?)

parkershepherd:
those notifications are a side effect of `clearOutOfBoundsMarkers`, which does speed up the map hugely if you have a large scan area

invisiblek:
https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2206

invisiblek:
ah shit, that didn't work as expected

invisiblek:
not the prettiest thing in the world: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2214

PiaFraus:
I think it's fixed in #2131

invisiblek:
ended up being fixed with: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2282

PR #2131, opened 2016-07-26
Improved and fixed notifications

PiaFraus:
Improved and fixed notifications

## Description
* Notifications has nothing to do with creating markers. So I moved it out of there to a special processor.
* Notifications don't spam anymore - if more then 3 notification will be spawned in the same update - one general notification will popup.
* Added wildcard(*=all) pokemon to notification selector.
* If something is selected along with the wildcard - it will be excluded from the wildcard. So developers can think of wildcards as inverters of what is selected.
* Pokemon bouncing will happen on click and (by request in comments) when page is just in the background.
* Pokemon bouncing will stop after 2 seconds of notification click.

## Motivation and Context
Notifications is one of the most important feature for me and my friends/colleagues.
Previously we had to select hundreds of them one by one. Now there is a wildcard.
Previously if you refresh a page or just too many pokemons spawned - you will have way too many notifications. Now they are joined together.

## How Has This Been Tested?
Tested locally.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -203,6 +203,16 @@ function excludePokemon(id) {
     ).trigger('change')
 }
 
+function isPokemonExcluded(id) {
+    var wildcard = excludedPokemon.indexOf(-1) >= 0;
+    return excludedPokemon.indexOf(id) >= 0 ? !wildcard : wildcard;
+}
+
+function isPokemonNotified(id) {
+    var wildcard = notifiedPokemon.indexOf(-1) >= 0;
+    return notifiedPokemon.indexOf(id) >= 0 ? !wildcard : wildcard;
+}
+
 function notifyAboutPokemon(id) {
     $selectNotify.val(
         $selectNotify.val().concat(id)
@@ -503,7 +513,7 @@ function getGoogleSprite(index, sprite, display_height) {
     };
 }
 
-function setupPokemonMarker(item, skipNotification, isBounceDisabled) {
+function setupPokemonMarker(item, isBounceDisabled) {
 
     // Scale icon size up with the map exponentially
     var icon_size = 2 + (map.getZoom()-3) * (map.getZoom()-3) * .2 + Store.get('iconSizeModifier');
@@ -537,18 +547,6 @@ function setupPokemonMarker(item, skipNotification, isBounceDisabled) {
         content: pokemonLabel(item.pokemon_name, item.disappear_time, item.pokemon_id, item.latitude, item.longitude, item.encounter_id),
         disableAutoPan: true
     });
-
-    if (notifiedPokemon.indexOf(item.pokemon_id) > -1) {
-        if (!skipNotification) {
-            if (Store.get('playSound')) {
-              audio.play();
-            }
-            sendNotification('A wild ' + item.pokemon_name + ' appeared!', 'Click to load map', 'static/icons/' + item.pokemon_id + '.png', item.latitude, item.longitude);
-        }
-		if (marker.animationDisabled != true){
-			marker.setAnimation(google.maps.Animation.BOUNCE);	
-		}
-    }
 
     addListeners(marker);
     return marker;
@@ -668,7 +666,7 @@ function clearStaleMarkers() {
     $.each(map_data.pokemons, function(key, value) {
 
         if (map_data.pokemons[key]['disappear_time'] < new Date().getTime() ||
-                excludedPokemon.indexOf(map_data.pokemons[key]['pokemon_id']) >= 0) {
+                isPokemonExcluded(map_data.pokemons[key]['pokemon_id'])) {
             map_data.pokemons[key].marker.setMap(null);
             delete map_data.pokemons[key];
         }
@@ -677,7 +675,7 @@ function clearStaleMarkers() {
     $.each(map_data.lure_pokemons, function(key, value) {
 
         if (map_data.lure_pokemons[key]['lure_expiration'] < new Date().getTime() ||
-                excludedPokemon.indexOf(map_data.lure_pokemons[key]['pokemon_id']) >= 0) {
+                isPokemonExcluded(map_data.lure_pokemons[key]['pokemon_id'])) {
             map_data.lure_pokemons[key].marker.setMap(null);
             delete map_data.lure_pokemons[key];
         }
@@ -763,7 +761,7 @@ function processPokemons(i, item) {
         return false; // in case the checkbox was unchecked in the meantime.
     }
     if (!(item.encounter_id in map_data.pokemons) &&
-        excludedPokemon.indexOf(item.pokemon_id) < 0) {
+        !isPokemonExcluded(item.pokemon_id)) {
         // add marker to map and item to dict
         if (item.marker) item.marker.setMap(null);
         if (!item.hidden) {
@@ -814,11 +812,11 @@ function processLuredPokemon(i, item) {
         disappear_time: item.lure_expiration
     };
 
-    if (item2.pokemon_id == null) {
+    if (item2.pokemon_id == null || isPokemonExcluded(item2.pokemon_id)) {
         return;
     }
 
-    if (map_data.lure_pokemons[item2.pokestop_id] == null && item2.lure_expiration) {
+    if (map_data.lure_pokemons[item2.pokestop_id] == null && item2.lure_expiration > new Date().getTime()) {
         //if (item.marker) item.marker.setMap(null);
         if (!item2.hidden) {
             item2.marker = setupPokemonMarker(item2);
@@ -826,7 +824,7 @@ function processLuredPokemon(i, item) {
         }
 
     }
-    if (map_data.lure_pokemons[item.pokestop_id] != null && item2.lure_expiration && item2.active_pokemon_id != map_data.lure_pokemons[item2.pokestop_id].active_pokemon_id) {
+    if (map_data.lure_pokemons[item.pokestop_id] != null && item2.lure_expiration > new Date().getTime() && item2.active_pokemon_id != map_data.lure_pokemons[item2.pokestop_id].active_pokemon_id) {
         //if (item.marker) item.marker.setMap(null);
         map_data.lure_pokemons[item2.pokestop_id].marker.setMap(null);
         if (!item2.hidden) {
@@ -880,10 +878,70 @@ function processScanned(i, item) {
 
 }
 
+function processNewPokemons(pokemons, oldPokemonKeys) {
+
+    newPokemonItems = []
+
+    $.each(pokemons, function (key, item) {
+        if(oldPokemonKeys.indexOf(key) == -1 && isPokemonNotified(item.pokemon_id)){
+            newPokemonItems.push(item);
+        }
+    });
+
+    if(newPokemonItems.length==0) return;
+
+    if (Store.get('playSound')) {
+      audio.play();
+    }
+
+    $.each(newPokemonItems, function (key, item) {
+        if (item.marker.animationDisabled != true) {
+            item.marker.setAnimation(google.maps.Animation.BOUNCE);
+        }
+    });
+
+    if(newPokemonItems.length < 4) {
+        $.each(newPokemonItems, function (key, item) {
+            var notification = sendNotification('A wild ' + item.pokemon_name + ' appeared!', 'Click to load map', 'static/icons/' + item.pokemon_id + '.png');
+            if(notification){
+                notification.onclick = function () {
+                    window.focus();
+                    notification.close();
+                    centerMap(item.latitude, item.longitude, false);
+                    if (item.marker.animationDisabled != true) {
+                        item.marker.setAnimation(google.maps.Animation.BOUNCE);
+                    }
+                    setTimeout(function() {
+                        redrawPokemon(pokemons);
+                    }, 2000);
+                };
+            }
+        });
+    } else {
+        var notification = sendNotification('Several pokemons appeared!', 'Click to load map', 'static/appicons/114x114.png');
+        if(notification){
+            notification.onclick = function () {
+                window.focus();
+                notification.close();
+                $.each(newPokemonItems, function (key, item) {
+                    if (item.marker.animationDisabled != true) {
+                        item.marker.setAnimation(google.maps.Animation.BOUNCE);
+                    }
+                });
+                setTimeout(function() {
+                    redrawPokemon(pokemons);
+                }, 2000);
+            };
+        }
+    }
+
+}
 
 function updateMap() {
 
     loadRawData().done(function (result) {
+        oldPokemonKeys = Object.keys(map_data.pokemons)
+        oldLurePokemonKeys = Object.keys(map_data.lure_pokemons)
         $.each(result.pokemons, processPokemons);
         $.each(result.pokestops, processPokestops);
         $.each(result.pokestops, processLuredPokemon);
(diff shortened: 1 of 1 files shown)


Toeler:
> Pokemon bouncing will happen on click and not when page is just in the background.

I rely on this when the map is sitting on my other monitor. Chrome doesn't pop up the toast when you are running a full screen application, so the bounce + ping is what I use.

PiaFraus:
@Toeler I can added a default bounce as well. But for how long?

Toeler:
For the duration that the marker is alive, preferably.

A feature like this is always going to be tough to please everyone.

static/map.js
```diff
@@ -520,20 +530,9 @@ function setupPokemonMarker(item, skipNotification) {
...
 
-    if (notifiedPokemon.indexOf(item.pokemon_id) > -1) {
-        if (!skipNotification) {
-            if (Store.get('playSound')) {
-              audio.play();
-            }
-            sendNotification('A wild ' + item.pokemon_name + ' appeared!', 'Click to load map', 'static/icons/' + item.pokemon_id + '.png', item.latitude, item.longitude);
-        }
-        // Icons still get a bounce, even on redraw
-        marker.setAnimation(google.maps.Animation.BOUNCE);
```
AHAAAAAAA: Please include the Bounce animation in your fix.

static/map.js
```diff
@@ -1049,7 +1100,7 @@ $(function () {
 
     // Load pokemon names and populate lists
     $.getJSON("static/locales/pokemon." + language + ".json").done(function(data) {
-        var pokeList = []
+        var pokeList = [ { id: -1, text: '*'}]
```
AHAAAAAAA: Can you explain why this is necessary?

PiaFraus: * When pokemons are excluded from showing on the map - they already excluded from loading. Most of users I know excluded the "I don't care about pokemons" and had to manually select all other pokemons in the notification selector. This allowed them to just select "all but excluded".
* It also allows to have "Include only selected pokemons", since everything that is selected along with the wildcard will be excluded.

static/map.js
PiaFraus: @AHAAAAAAA Included. It will still stop bouncing after notification click, so I assume it will have best from both worlds, since people need a way to stop them bouncing.

PiaFraus:
@Toeler I've returned the previous behaviour. But bouncing will still be stopped if user clicks notfication. So I hope it will please the most.

hhu94:
It seems this fix doesn't contain #2017, every time I zoom in/out I get notifications.

PiaFraus:
@hhu94 I agree. I misunderstood with another issue.

hhu94:
A problem I've found is that disabling the Wild Pokemon switch on the sidebar doesn't hide lured pokemon, only normal pokemon.

PiaFraus:
@hhu94 Well I haven't done anything about this switch. But I can actually add the switch check to "isPokemonExcluded" so it would also work as a wildcard.

andrewda:
> Notifications don't spam anymore - if more then 2 notification will be spawned in the same update - one general notification will popup.

Please change this to more than 3.

PiaFraus:
@andrewda Updated to 3.
 That's quite a number of commits. Should I rebase?

static/map.js
```diff
@@ -235,7 +245,7 @@ function initMap() {
               'style_pgo_nl']
         },
     });
-
```
invisiblek: while i agree whitespace should be removed, it should be done so in a separate commit

static/map.js
```diff
@@ -531,21 +541,9 @@ function setupPokemonMarker(item, skipNotification, isBounceDisabled) {
...
-            sendNotification('A wild ' + item.pokemon_name + ' appeared!', 'Click to load map', 'static/icons/' + item.pokemon_id + '.png', item.latitude, item.longitude);
-        }
-		if (marker.animationDisabled != true){
-			marker.setAnimation(google.maps.Animation.BOUNCE);	
-		}
-    }
-
     addListeners(marker);
     return marker;
-};
```
invisiblek: unrelated to this commit (along with all the other semicolons you removed below)

static/map.js
PiaFraus: Ok, I'll remove it.
Used to the rules at my work, where if we see style errors, which don't change the behaviour - we just add them with next change instead of creating separate change for them.

static/map.js
```diff
@@ -852,21 +850,84 @@ function processScanned(i, item) {
...
         $.each(result.pokestops, processLuredPokemon);
         $.each(result.gyms, processGyms);
         $.each(result.scanned, processScanned);
-        clearOutOfBoundsMarkers(map_data.pokemons);
-        clearOutOfBoundsMarkers(map_data.lure_pokemons);
-        clearOutOfBoundsMarkers(map_data.gyms);
-        clearOutOfBoundsMarkers(map_data.pokestops);
-        clearOutOfBoundsMarkers(map_data.scanned);
+        // If someone needs this feature - please implement a version that doesn't break notifications
+        // clearOutOfBoundsMarkers(map_data.pokemons);
```
invisiblek: i'd rather we leave this in personally, it helps quite a bit with larger maps

PiaFraus: It forces markers to be deleted and this introduced a bug, that when you move a map - all notifications are spawned again. What problem this supposed to fix? Memory issues?
Personally if I am on map and move left-right - I would prefer to immidately see existed markers, since they already loaded instead of spawning a new one and notifying me about it. I need notification for something I couldn't see. Something that is currently out of the bounds.

invisiblek: well, when we fetch new data, we're already leaving out anything not on our map, so when we pan, we see what was there before, sure, but  we wont see anything new until it refreshes again

i guess what i'm saying is if we're going this direction (not pruning out of bounds markers) we should just pull all the data too (if the reason for doing the former is for quicker markers when panning)

invisiblek: regardless, it feels to me this wildcard thing is a new feature that i don't quite understand, and i'm not sure we want to introduce it for 2.1.0 (which https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1951 needs to be fixed for)

hhu94:
I really like the wildcard feature, it allows me to exclude the pokemon I'm not interested and receive notifications for all the other pokemon that aren't in the ignore list.

Also, this PR is precisely meant to fix #1951. As far as I have tested it, it does indeed fix it.

static/map.js
BillyWM: Let's not break one feature to fix another

> It forces markers to be deleted and this introduced a bug, that when you move a map - all notifications are spawned again. 

That's a more systemic problem with the code. Markers and notifications shouldn't be tied together to begin with. setupPokemonMarker shouldn't have this responsibility. It isn't setupPokemonMarkerAndSendNotifications. It should do one thing....set up markers.

There should probably be one function that just decides whether to send a notification. processPokemons is calling setupPokemonMarker. It should also call this new function e.g. sendNotifications, and that function can run logic to decide whether a notification is warranted (is this pokemon in the viewable area? Has a notification already been sent for this spawn? Is it part of the exclusion list. Etc)

>What problem this supposed to fix? Memory issues?

Memory and performance. If you think a function like that is useless, you shouldn't just yank it out of the code. Removing something unrelated that has been in there a while should be an isolated PR of its own, open for discussion.

hhu94: I really like the wildcard feature, it allows me to exclude the pokemon I'm not interested and receive notifications for all the other pokemon that aren't in the ignore list.

Also, this PR is precisely meant to fix #1951. As far as I have tested it, it does indeed fix it.

PiaFraus: I agree. Personally I'd just get all the pokemon data for my needs.
But I am fine with this middle-ground solution: 
Good sides:
-  it doesn't load everything for those who have multiple scanners on big areas
-  it also doesn't delete them on pan for people like me and doesn't force second notification.

Bad sides:
- people who constantly pan a lot for areas with new pokemons might have slow expirience.
- people who forgot to reset their view will not be notified about out-of-bound pokemons

PiaFraus: @BillyWM, 

> Let's not break one feature to fix another

Agree. This lines were added and broke the notification feature, so I am removing the code, that broke a feature.

PiaFraus: @invisiblek 
Actually wildcard is extremely simple. It just adds * to the list of pokemons. I think it's common notation, that star means "include everything".

BillyWM: >Memory and performance.
>Only for those who constantly pans. If it's out of bounds - it's not loaded in the first place.

Panning is kind of an important feature of the map! If the new notification code has to hamper panning, it's broken, and it should not be merged until it's fixed.

Please either restore the clearOutOfBoundsMarkers calls or rework the function so that it doesn't interfere with notifications.

What is it doing that interferes anyhow?

>And clearOutOfBoundsMarkers is not called clearOutOfBoundsMarkersAndAlsoAllThePokemons,

Pokemon, stops, and gyms on the map are markers. I don't get what your point is.

This is called the Single Responsibility Principle. I didn't make this idea up myself.

PiaFraus: @BillyWM

> Panning is kind of an important feature of the map! If the new notification code has to hamper panning, it's broken, and it should not be merged until it's fixed.

Notifications of pokemons is an important feature of an app which is about telling people about pokemons.
If this panning code broke the already implemented notification system, which I just fix - it was broken and should be removed as a fix which my change is about.

> Please either restore the clearOutOfBoundsMarkers calls or rework the function so that it doesn't interfere with notifications.

@invisiblek is fixing this method I think. I might wait for him to finish and see later. 
As right now with this change - everything works perfectly fine. All my friends and colleagues are much happier with my fix. I am not in rush.

> Pokemon, stops, and gyms on the map are markers. 

No. No they are not. They are map_data. - Data from the server.
For simplicity instead of creating markers in separated container, code ammends this data and ADDS to data also a marker property. Addition to a data. But map_data.pokemons - is not a list of markers. So
clearOutOfBoundsMarkers actually violates the separation of concerns principle and deletes not just markers, but all the received data.

BillyWM: >clearOutOfBoundsMarkers actually violates the separation of concerns principle and deletes not just markers, but all the received data.

Yes, that's wrong, but it doesn't *need* to be doing that. And nor do notifications need to be ripping out performance features.

invisiblek: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2214

The best I could come up with (working on the current constraints) was that.

BillyWM: > It might a little bit slow down for people who constantly look at new pokemons.

So, with this notification code, adding markers to the map triggers notifications.

So....no markers can be removed, because they would re-trigger notifications.

So if someone scans a wide or dense area (San Francisco, NYC, etc) they just have to live with tons of markers degrading performance, even if they zoom in only to their neighborhood or college campus?

BillyWM: I see this as analogous to game engine rendering. If there's a ton of entities, you don't blindly run all their draw logic all the time. You only have a draw list of those on screen.

If markers are on the map, Google Maps spends time managing them. Tons of markers = bad performance, regardless of whether they're currently rendered.

Managing the viewable area is a necessary step for performance.

PiaFraus: @BillyWM,
> So....no markers can be removed until they're expired, because they would re-trigger notifications
No. No data should be removed if it's used by other features. It was broken by this function. I removed the broken code. If someone wants to properly implement the perfomance boost by not destroying important features - good! But this is about fixing previously broken code.

> I  see this as analogous to game engine rendering. If there's a ton of entities, you don't blindly run all their draw logic all the time. You only have a draw list of those on screen.
Brilliant example. If you don't see where enemy is - your code shouldn't delete all info about this enemy. It shouldn't draw it - yes. But enemy can interact with you. Shoot at you even when it's not at your sight. If you submit a code, that deletes all enemies as soon as you don't see them - you remove their change and ask them to properly implement what they wanted.

PiaFraus: I might spend this night trying to properly implemented what they wanted to do. I'll see if I have time.

jjbott: At least for the clearOutOfBoundsMarkers fix, I think we're looking for something simple like : https://github.com/jjbott/PokemonGo-Map/commit/f841bc613109323200b68f7371d4b4253d23aa8a . Don't delete everything, just show/hide based on the visible map

xzyfer:
>For the duration that the marker is alive, preferably.

Please no. The bouncing is wildly distracting. At the very least we should be able to disable the bounce. The desktop notification is plenty for a lot of people.

Slivo-fr:
+1

Maybe give the user the choice of bounce or not, and duration.
I like to have the marker bounce it's whole lifetime on my side.

Agree with OP on clearOutOfBoundsMarkers side. It sounds like a broken feature.

PiaFraus:
@AHAAAAAAA 
I've just noticed your comments in #1322 and #1415 :
Wildcard('*') is essentially "All". I can rename it if you think it will be more clear.
And I've added it to both selectors and to just Notify as I think it's good if similar UI elements are as similar to each other as it could be.
Plus there are a couple of request to invert this fields and additional feature of wildcard allows that automatically.

PiaFraus:
@xzyfer @Slivo-fr Well currently bouncing is stopped if:
a) You click on any notification
b) You click on bouncing marker
c) You change zoom to remove bouncing for all of them
I don't really want to introduce new UI elements so may be we can think of something? But if you think it will really help -I'll do it.

Slivo-fr:
Current behavior is fine for me, was just throwing some ideas :)

invisiblek:
moving this to 2.2.0 as its now become a feature and we're trying to get 2.1.0 finished

jnovack:
> That's quite a number of commits. Should I rebase?

@PiaFraus always rebase and squash.

hhu94:
Can't wait for this to be merged. It's essential to the notification system.

PiaFraus:
@hhu94 Feel free to take ownership of it. I am not dealing with grunt.

invisiblek:
Rebased here: https://github.com/invisiblek/PokemonGo-Map/commit/4acc1bd47d89bf4c748609113731970ef4dd4c83

But I must admit, I still don't get the point of the wildcards. What possible reason could you have to hide ALL the Pokemon at once or display all notifications at once?

hhu94:
@invisiblek How I use it is I hide all the pokemon I don't want to see and just enter the star in the notifications list. That way I get notified of any pokemon that appears as long as it's not in my ignore list.

invisiblek:
Wouldn't it make more sense to have it be a toggle instead of adding it to the list? My thought would be toggling it on would grey out the list box (like it does when you turn the "Follow location" toggle on) and then act like the behavior you're describing. I still don't see the point of having it in the "Hide Pokemon" box though...

PiaFraus:
@invisiblek Main reason for me - simpler UI. 
At home I want to see everything around and be notified about everything but a few common ones.
Exclude pokemons=empty; Notify = * + Zubat/Goblat/...
At work I want to see only a bunch of rare, if I look at the map when I have time. I want to be notified only for super rare, since it can interrupt my workflow
Exclude pokemons = * + Eevee/...; Notify = Dratini/...

All of this options and more with just two UI elements.
Without wildcard there would probably be twice as many ui elements.

Slivo-fr:
The idea of toggles for wildcard make more sense imo. 
If this particular feature is to be discuted maybe you should split it in a second pull request ?

Slivo-fr:
Why the hell closing it ? it's still bug atm and I'm waiting for those improvements for a week now !

invisiblek:
PiaFraus has no intention of rebasing the patch. If you'd like to do so, feel free to submit a new pull request.

PR #2136, opened 2016-07-26
Fix multiple lured pokemon notification bug

hhu94:
## Description
This patch fixes a bug where notifications of lured pokemon would appear to spam the user, even after the lured pokemon disappears.

## Motivation and Context
Notifications of lured pokemon would repeatedly be triggered every updateMap() cycle (5 seconds), because processLuredPokemons() did not check if the lured pokemon passed in was expired.

To reproduce the bug, it's easier if you disable processPokemons() by commenting out the following line from updateMap():
`  $.each(result.pokemons, processPokemons);`

This way you only lured pokemon are scanned, narrowing down test subjects.

## How Has This Been Tested?
Tested by running the server, expired lured pokemon do not trigger notifications anymore in each updateMap() cycle, and new lured pokemon still trigger notifications as intended.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -782,13 +782,16 @@ function processLuredPokemon(i, item) {
         return;
     }
 
-    if (map_data.lure_pokemons[item2.pokestop_id] == null && item2.lure_expiration) {
+    if (map_data.lure_pokemons[item2.pokestop_id] == null
+            && item2.lure_expiration > new Date().getTime()) {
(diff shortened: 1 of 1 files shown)


PiaFraus:
I also fixed it in #2131 and also added a general function to check if pokemon should be excluded or not.

hhu94:
Closing, #2131 is a lot more complete.

PR #1960, opened 2016-07-25
Fixed absolute path in geolocation call

georgo:
## Description
Again, absolute path was asserted in javascript code. For instance, I'm running map behind apache proxy in subdirectory. This commit fixes issue, when "Geolocation" switch is enabled.

## How Has This Been Tested?
Tested locally on Google Chrome and on iOS.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2137, opened 2016-07-26
use relative location for AJAX requests

ToadKing:
## Description
Use relative path for `next_loc` calls in map.js

## Motivation and Context
For sites that are behind a reverse proxy, it's possible for them to be hosted in a subdirectory on a website: ex. example.com/map. To make sure everything works in that case, try to use relative paths for requests when possible.

## How Has This Been Tested?
Tested on my own server, using nginx as a reverse-proxy. Approximate server settings are as follows:
```
server {
   listen 80;
   server_name example.com;

   location /map/ {
      proxy_pass http://localhost:5000/;
      proxy_set_header X-Real-IP $remote_addr;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header Host $http_host;
   }
}
```

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -1088,14 +1088,13 @@ $(function () {
     window.setInterval(function() {
       if(navigator.geolocation && Store.get('geoLocate')) {
         navigator.geolocation.getCurrentPosition(function (position){
-          var baseURL = location.protocol + "//" + location.hostname + (location.port ? ":"+location.port: "");
           lat = position.coords.latitude;
           lon = position.coords.longitude;
 
           //the search function makes any small movements cause a loop. Need to increase resolution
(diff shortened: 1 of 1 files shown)


andrewda:
Actually, could you just change that all to the [`changeLocation(lat, lng)`](https://github.com/ToadKing/PokemonGo-Map/blob/4828428d0d892e7ee756ee1ed5ddb9612d514a5d/static/map.js#L1008-L1014) function?

ToadKing:
Updated. I kept in the anti-jitter filter since it seems like it's useful.

georgo:
This is already submitted in PL #1960

PR #2150, opened 2016-07-26
Search multiple locations from single instance

adamhwang:
## Description

- Allows users to add multiple map pins
- Lists current map pins in side bar
- Move map pins to change location
- Delete map pins from side bar
- Locations will be searched inside-out sequentially by step count (i.e. all center locations will be searched, then all step 1, then all step 2, etc.)
- Command line -l support with pipe delimits (i.e. "-l Location 1|Location 2|Location 3")

## Motivation and Context

Gives you the ability to monitor multiple locations without needing to run several servers.  Allows a single server to serve multiple users.  Will be even more powerful once threading issue is fixed.  Somewhat depends on #2128 when using a lower step count.

## How Has This Been Tested?

Regular multi-user usage with Chrome+Windows

image

## Types of changes

## Checklist:


tkSimon:
atm you can run multiple servers in "server mode", basically without the web server, writing to the same DB, while the web server provides clients with data from all the servers instances. The upside to this is that you can define different step count for different locations. I guess that's not possible with your PR?

adamhwang:
The draw back for that method is that you can only have a fixed number of points and the points aren't changeable.  I like to scan areas I'm traveling through, or check out some new locations that could require a detour.  That wouldn't be possible with -ns unless you remote back to the server, figure out the nearest landmarks or lat/lon, and then deploy another server.   You'd also need separate accounts for each location unless you want simultaneous logins from a single account, and this is all ignoring the fact running an entirely new instance is an incredibly heavy thing for something this trivial.

Agreed about the step counts.  But arguably, that (and location) shouldn't be strictly a server setting to begin with.  This PR isn't perfect, but moves us one step closer to removing those limitations.

adamhwang:
I'd also like to point out this shouldn't preclude anybody from using the map in exactly the same way as before.  The only reason I labeled it breaking was that instead of changing locations from the search box, it adds a new one.

n3cr0cr0w:
Adam, I see we are scanning close to the same area, but I'm planning on scanning the whole area beehive style.

n3cr0cr0w:
image

vinsilsim:
Traceback (most recent call last):
  File "runserver.py", line 13, in <module>
    from pogom.app import Pogom
  File "/root/Map/pogom/app.py", line 13, in <module>
    from pogom.utils import get_args, get_location_key
ImportError: cannot import name get_location_key

Anyone can help solve this issue?

adamhwang:
@vinsilsim do you have def get_location_key...  in your utils.py file?
https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2150/commits/b2ae4f0342ab724cc09a960831b4f6f6f4a90963#diff-6c84a64c790b52a5ff2656d6bb388fefR147

vinsilsim:
@adamhwang now the issue is another haha sorry =/ 

Exception in thread search_thread:
Traceback (most recent call last):
File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
self.run()
File "/usr/lib/python2.7/threading.py", line 754, in run
self.__target(self.__args, *self.__kwargs)
File "/root/Map/pogom/search.py", line 179, in search_loop
if args.thread_delay > 0:
AttributeError: 'Namespace' object has no attribute 'thread_delay'

adamhwang:
@vinsilsim it appears that are missing the latest for many of your files

vinsilsim:
@adamhwang I have the newest version along with this change

vinsilsim:
@adamhwang can you help me? Your change is very important to me

adamhwang:
@vinsilsim please try to update the whole project from the feature branch.  I cannot help you figure out why you don't have the latest files.

https://github.com/adamhwang/PokemonGo-Map/tree/multisearch2

vinsilsim:
@adamhwang solved! thank you!

adamhwang:
based on @hgentry comments and work on the beehive generator (#2153), I've updated this PR to include support for the drawbacks he outlined.  multiple locations are supported pipe-delimited and the algorithm to process multiple locations works outside-in so that the centers of each location is processed first, then the first steps of each location, etc:

image

hgentry:
This is really cool, multiple users is a big deal and it looks like it's algorithmically sound

adamhwang:
@vinsilsim no, sorry.  please try whatever you did for https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2150#issuecomment-235683284 again.  I do not know how or why your environment is incapable of getting the latest updates.

adamhwang:
@vinsilsim you seem to be missing https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2150/commits/1cfb0cd46e37e8ccbbe16c686b245cec3fe05f58

vinsilsim:
@adamhwang fixed bro, sorry, im noob.

AHAAAAAAA:
I do appreciate the effort being put in here, but I prefer that this project not go down the path of multi-user authentication, sessions, and the like right now. If we were to implement this, then we'd have to implement it truly instead of hacking together a patch that allows it. We're already able to scan large amounts of area, with the honeycombing approach covering entire metropolitan areas without being too resource intensive. We can even run scanners on any number of towns on the same map. This has given rise to a lot of community maps all over the world, which helps maintain the community vibe. I understand many users are running this for their friends, I am too, but the server should be able to cover the whole area needed. Of course we can compromise on this, but that's my position on these PRs. Closing this for now.

vinsilsim:
@AHAAAAAAA show me a version that covers the entire world map. Why all the versions I've seen so far has problems with many instances. The @adamhwang project is the best I've seen so far for multiple users at the same time.

adamhwang:
@AHAAAAAAA with all due respect, that is a very strange position, and I hope you reconsider.  why add threads at all if you could have just as easily run multiple scanners?  obviously because running multiple scanners is a huge waste of resources.  imagine every project that could have been improved with threading, only to have been told to instead run multiple copies of itself.

while an beneficial side effect, the main motivation for this PR wasn't for multiple users, but to be able to search multiple locations, which is clearly a valid use case, without having the need to run additional scanners.  it's a PR that also allows those multiple searches to be added, removed, and changed dynamically without needing access to your server.  this is a mobile game that requires you to move around and explore new places.  why force people to have to remote into their server to be able to make any changes?

adamhwang:
Also, I would like to point out that there is no multi-user authentication or sessions in this PR.

chuyskywalker:
I pushed for @AHAAAAAAA to make a decision about some of these "multi" efforts, so I want to expand on this a bit and not leave you hanging.

I know "slippery slope" is a logical fallacy, but not so much in this case I think. With a feature like this, people would see it, see the "follow me" feature and file about 10 bugs a day about how the "follow me" feature doesn't work for them and their friends. To make that the case, we'd need to implement _some_ kind of multi-user system, which most of the devs have agreed isn't something we want to support.

Additionally, this is kind of a half-baked "manage search threads via ui" setup. For example, how would people manage their step values? Shouldn't they be able to have one marker with `step=15` and another with `step=2` for a very specific extra location somewhere. Also, they can't assign different logins to each search thread, which is super useful for being able to increase scan speed and reliability.

I think that, ultimately, we want an application that can be _used by_ multiple people for a given, known area but _managed_ by a single person.

adamhwang:
@chuyskywalker thanks for the follow up.  I appreciate it.  

if that's your concern, the "follow me" could be a named pin defined by the client.  I think that'd be trivial to implement actually.  no state or auth changes to the server would be required

step values are a different story.  I don't think it should be a server variable, but at the same time, I don't see why it matters if step # is fixed either

finally, I don't really buy the login argument.  sure it's fun to stockpile hundreds of logins, but I've run 100+ threads against the same ptc account without issue.  any reasonably intelligent engineer could just as easily detect an account scanning once every 5s as they can an account scanning once every 5ms.  in fact, if you had 100+ accounts, it'd probably last longer to only scrape with one account at a time until it's banned, instead of getting all of them banned at once

any ways, I appreciate the work you guys are all putting in here.  I know it's changed the way my friends and I have played the game for the better.  my branch is stable and has everything I need right now, but I'll try to continue to rebase with master if you guys change your mind.

PR #2179, opened 2016-07-26
Multiple search locations

cichacz:
Allows users to scan several areas (instead of creating one huge area)

## Description
Location argument is parsed to return more than one result. 
Locations should be separated by |

## Motivation and Context
I have few areas where i spend a day, and creating one area to cover them all would be pointless

## How Has This Been Tested?
Tested on Debian and Win10, just observing map behavior, and checking server for errors

pgomap

## Types of changes

## Checklist:
- notify users that they can specify multiple locations


Pedro-vk:
+1
And how could I set multiple locations?

cichacz:
Via command line or in config, eg. "38.891206, -77.036644|39.271855, -76.602401"

tomdupont:
python runserver.py -l "50.823327, 3.227706|50.843981, 3.204446" <other arguments> gives this error: [  ERROR] Could not get a position by name, aborting.

Viralpoke:
Yeah, I tried altering the config for the deployment on heroku.com and get an error also
can't believe there's not option an option for multiple points

cichacz:
@tomdupont you have to checkout my branch if you want to test it
@Viralpoke from what i saw in source (during my development) there was ability to specify only 1 location at a time

chuyskywalker:
This is already possible by running multiple instances with different `-l` flags right now. Has some file locking contention problems at the moment, but is relatively functional.

cichacz:
@chuyskywalker you're right, but then you're unable to alter other locations using map, am i right?
My version allows you to move any marker, only search field in "filters" tab moves first one

pogom/pgoapi/utilities.py
```diff
@@ -50,18 +50,22 @@ def camelcase():
...
+        locations = []
+        locations_names = location_name.split("|")
+        for location in locations_names:
+            geolocator = GoogleV3()
+            loc = geolocator.geocode(location)
+            if loc:
+                locations.append((float(loc.latitude), float(loc.longitude)))
+        return locations
+
+    return None
```
kinncj:
```python

def get_pos_by_name(location_name):
    locations = []
    locations_names = location_name.split("|")
    for location in locations_names:
        geolocator = GoogleV3()
        loc = geolocator.geocode(location)
        if loc:
            locations.append((float(loc.latitude), float(loc.longitude)))
    return locations
```

This should be enough.

jnovack:
http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

adamhwang:
it seems we have duplicated our efforts? #2150

AHAAAAAAA:
I do appreciate the effort being put in here, but I prefer that this project not go down the path of multi-user authentication, sessions, and the like right now. If we were to implement this, then we'd have to implement it truly instead of hacking together a patch that allows it. We're already able to scan large amounts of area, with the honeycombing approach covering entire metropolitan areas without being too resource intensive. We can even run scanners on any number of towns on the same map. This has given rise to a lot of community maps all over the world, which helps maintain the community vibe. I understand many users are running this for their friends, I am too, but the server should be able to cover the whole area needed. Of course we can compromise on this, but that's my position on these PRs. Closing this for now.

cichacz:
@adamhwang you're right i didn't notice it :)
We posted it on the same day so i had this ready anyway.. :)

@AHAAAAAAA thank you for your opinion, although i have different approach to this. I'm using 2 markers, the second one is always placed in the area where i spend the most of my day, while the first one (which is somewhat "main" marker) is used to scan some new locations where I'm moving during the day.
This gives me an ability to keep radius of reasonable size (which improves "freshness"), and to be up to date with pokemons close to my home/work and also while traveling.

chuyskywalker:
That's possible right now by running a second copy of the application in `-ns` mode ("no webserver") by setting the location of the second instance to the static location you're interested in. The main application's search thread can follow you as normal.

adamhwang:
@chuyskywalker copying my comment from #2150 for you here: with all due respect, that is a very strange position, and I hope you reconsider. why add threads at all if you could have just as easily run multiple scanners? why add steps when you could have left it to a single step and run honeycombed 1-step scanners for your entire search area? obviously because running multiple scanners is a huge waste of resources. imagine every project that could have been improved with a simple loop, only to have been told to instead run multiple copies of itself.

while a beneficial (or negative?) side effect, the main motivation for this PR wasn't for multiple users, but to be able to search multiple locations, which is clearly a valid use case, without having the need to run additional scanners. it's a PR that also allows those multiple searches to be added, removed, and changed dynamically without needing access to your server. this is a mobile game that requires you to move around and explore new places. why force people to have to remote into their server to be able to make any changes?

Serzh90:
Is there any chance this will be implemented ?

PR #2209, opened 2016-07-26
Only notify if pokemon is within current map view bounds.

xorrbit:
## Description
This will only fire a notification if the pokemon is in the current map view bounds.

## Motivation and Context
When you are running a very large map with multiple scanners, you don't want to notify on a pokemon that isn't within a distance you can reasonably get to before it despawns. If you zoom your view to things you can reasonably get to in time, this will only notify when things spawn inside this view.

## How Has This Been Tested?
Tested on a local install I'm using to map a view large area. Works properly.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -531,7 +531,7 @@ function setupPokemonMarker(item, skipNotification, isBounceDisabled) {
         disableAutoPan: true
     });
 
-    if (notifiedPokemon.indexOf(item.pokemon_id) > -1) {
+    if (notifiedPokemon.indexOf(item.pokemon_id) > -1 && map.getBounds().contains(marker.getPosition())) {
         if (!skipNotification) {
             if (Store.get('playSound')) {
               audio.play();


invisiblek:
clearOutOfBoundsMarkers should prevent this from happening...

xorrbit:
@invisiblek the notifications are fired in setupPokemonMarker() which is called from processPokemons() which is run before clearOutOfBoundsMarkers().

PiaFraus:
> When you are running a very large map with multiple scanners, you don't want to notify on a pokemon that isn't within a distance you can reasonably get to before it despawns.
Reasoning is good. But implementation is not!
I especially want a notification when it's out of bounds, because I don't see that pokemon and want to be notified about it.

xorrbit:
@PiaFraus yeah, coupling the notification area to the current view was a compromise I felt was reasonable. Any ideas how to implement (or what the UI would be) for separation of "area of notifications" vs "current view of the map"? Maybe draw a circle on the map or something? What would be ideal for your use?

PiaFraus:
@xorrbit For me the ideal is how it's working right now. (Well plus my changes in #2131, which group notification to prevent spam) , My area of interest - is the zone I am in and 5 steps around. I want to see pokemons in this area of interest + I want to get notifications about them.  I don't really need a feature to restrict notification to a smaller sub area. But if I would implement it, I would probably go with either circle or square.
BTW, I actually think that out  of bounds pokemons are not even loaded on the map with mapUpdate, so they shouldn't show unless they are from lured stops. I fixed this particular issue in my change.

xorrbit:
@PiaFraus it does look like this might not be an issue anymore in #2131, looks like some good improvements in there, I'll test it out in a bit.

PR #2214, opened 2016-07-26
Only notify once

invisiblek:
## Types of changes

## Checklist:

* Maintain an array of markers that notifications have been sent for
  (based on encounter_id)

### DIFF
--- a/static/dist/js/map.js
+++ b/static/dist/js/map.js
@@ -17,6 +17,7 @@ var idToPokemon = {};
 
 var excludedPokemon = [];
 var notifiedPokemon = [];
+var notifiedMarkers = [];
 
 var map;
 var rawDataIsLoading = false;
@@ -443,7 +444,9 @@ function setupPokemonMarker(item, skipNotification, isBounceDisabled) {
         disableAutoPan: true
     });
 
-    if (notifiedPokemon.indexOf(item.pokemon_id) > -1) {
+    if (notifiedPokemon.indexOf(item.pokemon_id) > -1 && notifiedMarkers.indexOf(item) < 0) {
+        notifiedMarkers.push(item);
+
         if (!skipNotification) {
             if (Store.get('playSound')) {
                 audio.play();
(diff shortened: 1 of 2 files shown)


invisiblek:
i'm not sure i like this all that much to be honest

the array will pretty much grow forever until the page is refreshed, the only spot it worked to actually prune from the array is when you hit the remove link in the popup box, otherwise the array was getting wiped constantly when i put it in the clearStaleMarkers method

PiaFraus:
@invisiblek I think most of the problems with notifications comes from the fact that they are based on visibility of markers. Well I kinda fixed current symptoms in #2131 by moving all checks if notification should be shown or not, after clearStaleMarkers;

invisiblek:
oh bleh, we don't even get ALL the pokemon, just the ones that would show up on our screen...ffs

invisiblek:
Latest patch set is the best i've been able to come up with (aside from a major refactor as @PiaFraus has done)

It still builds the array, but now with a timestamp of when that notification fired. It prunes the array based on a 30 minute window, which theoretically should be long enough to assume that pokemon is gone.

BillyWM:
FYI splice modifies arrays in-place

BillyWM:
The marker code is kind of a tangled mess that could use a good refactor but as it stands now I see this as a more sensible solution - storing a notified list off to the side.

invisiblek:
> FYI splice modifies arrays in-place. You can operate directly on notifiedMarkers.

doesn't that fuck with the $.each loop? for whatever reason i thought it does, so i did the copy/copy situation here

invisiblek:
ok, i'm about as happy as i can be with this commit

PiaFraus:
@invisiblek Would it still be relevant with my changes (#2131) after recent change to clearOutOfBouns? Since now we always have data and we hide just the markers - actual pokemon data lives from loading till exclusion. So notification will show up only once.

invisiblek:
Ha, yea I think you're right. We no longer call setupPokemonMarker() for the same marker (ever) so it should never kick a notification anymore.

PR #2263, opened 2016-07-26
Feature: Templates refaktoring :-)

sdhormess:
This pull request includes a

- [ ] Bug fix
- [x] New feature
- [ ] Translation

- Little Template structure
- Usage of (Jinja-)Blocks, results a way smaller map.html
- :-)


AHAAAAAAA:
Could we get some discussion in here over whether to do this or not? I'm leaning towards no because we're not at the size where this makes sense yet.

PR #2282, opened 2016-07-27
Instead of deleting out of bound data, only show/hide markers

jjbott:
## Description
clearOutOfBoundsMarkers deletes valuable data, which causes problems like duplicate notifications and markers coming back after they've been hidden. Refactor it so is only responsible for showing/hiding markers, and doesn't destroy data. 

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -686,22 +686,29 @@ function clearStaleMarkers() {
     });
 };
 
-function clearOutOfBoundsMarkers(markers) {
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
Tested, works. I like this approach, it makes more sense than other suggestions.

medghaim:
Testing

medghaim:
+1

PR #2291, opened 2016-07-27
Lock sqlite db writes

adamhwang:
## Description

Lock sqlite db writes to prevent "OperationalError: database is locked" error when running multiple threads.  (Will not prevent same error caused when running multiple servers.)

## Motivation and Context

With threading hopefully soon-to-be fixed, this PR will help alleviate some of the multithreading issues sqlite users will face.

## How Has This Been Tested?

Tested code against #1910 branch with 20 threads and sqlite.

Before after a few seconds "database is locked" happens on almost every single db write:
image

After, "database is locked" only happens during db reads (/raw_data):
image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -9,6 +9,7 @@ from peewee import Model, MySQLDatabase, SqliteDatabase, InsertQuery, IntegerFie
                    OperationalError
 from datetime import datetime, timedelta
 from base64 import b64encode
+from threading import Lock
 
 from . import config
 from .utils import get_pokemon_name, get_args, send_to_webhook
@@ -21,9 +22,11 @@ log = logging.getLogger(__name__)
 
 args = get_args()
 db = None
+db_lock = None
 
 def init_database():
     global db
+    global db_lock
     if db is not None:
         return db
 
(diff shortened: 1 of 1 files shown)


slonce:
Not working. I play in 3 instanse in 30 thread and "Database is locked"

adamhwang:
Like I said, this will do nothing for multiple instances.  Only for multiple threads within a single instance.

slonce:
aaa ok thanks. thats sorry. But for multiple instanse that connected for 1 db have fix?!)

Pedrock:
@slonce You should be using MySQL.

marauder37:
1. have you tested this on Windows? It's an alternate universe there, where locking doesn't work reliably
2. don't use acquire/release, just write `with lock:`

adamhwang:
yes, only tested with Windows. somebody will need to check Linux compatability. 

"with lock:" will require duplicating line 332 in an if statement. I'm fine with it either way if there's something wrong with acquire/release

wcastello:
Wouldn't this just introduce a new lock? There's already one in search_thread guarding the entry to parse_map().

adamhwang:
@wcastello you're right.  closing PR

PR #2338, opened 2016-07-27
added indexes for faster searching of the database

polygamma:
## Description

I added indexes to some of the columns in the tables of the database.

## Motivation and Context

Many people want to use the data they have crawled with the PokemonGo-Map to search for specific spawnpoints etc. After letting the PokemonGo-Map run for days, searching through the data gets really, really slow. Adding indexes to the columns, which are used for scraping the most, has changed this drastically. You can read more about it, for example, here: http://stackoverflow.com/a/1130

## How Has This Been Tested?

- Checked it with 'old' databases and can confirm that nothing bricks.
- Checked it also with 'new' (empty) databases and can confirm that the indexes are being added.

## Types of changes

If the database has already been created and is not empty, this commit changes nothing.
It bricks nothing, but also doesn't add anything.
But if one starts with a new, clean database, the indexes are being added.

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -56,11 +56,11 @@ class Pokemon(BaseModel):
     # We are base64 encoding the ids delivered by the api
     # because they are too big for sqlite to handle
     encounter_id = CharField(primary_key=True, max_length=50)
-    spawnpoint_id = CharField()
-    pokemon_id = IntegerField()
+    spawnpoint_id = CharField(index=True)
+    pokemon_id = IntegerField(index=True)
     latitude = DoubleField()
     longitude = DoubleField()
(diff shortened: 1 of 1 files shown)


zeryl:
May also want to do an index on lat/lon in the tables.  A combined index since the searches do a window (lat > xxx lon > yyyy).

polygamma:
Good idea @zeryl ! Added it.

marauder37:
I have confirmed that the pokemon_latitude_longitude index is being hit for the raw_data query with mysql so I'm +1 on this

marauder37:
+1 from e-wold

Issue #1879, opened 2016-07-24
Negative Time Left


calebbait:
**## Current Behavior**
I turned on the -dc option and noticed some Pokemon appear with negative time. There isn't any pattern to what Pokemon are effected by this, it could be specific spawns but I'm not sure. I'll collect more data and update when/if I can. The negative time Pokemon also do not show on the map.

**## Steps to Reproduce (for bugs)**
1. Run map with -dc (or output Pokemon data in another way)
2. Wait for a bugged Pokemon to appear

**## Additional Information**
In my feature request #1709 I suggested a data collection script as an alternative to the database -D option. I threw together a quick edit of customLog.py and models.py to dump the data in a comma delimited text file as described. One such piece (or, rather pieces) of data it dumps are spawn time and date. Because of this I decided to check the time/date for all the Pokemon with negative times and notices a very distinct pattern. Every Pokemon with a negative time, when the math is done, "spawns" on Monday, July 18th at 17:38 (5:38 PM) EST. What does this mean exactly? I have no clue. But the negative times it gives definitely aren't random.

**## Logs**
In console with -dc
```
======================================
 Name: grimer
 Coord: [REDACTED]
 ID: 88
 Remaining Time: -6 days, 2:45:27.921000
======================================
```
In my data dump
`Grimer,[REDACTED],[REDACTED],17:38,Mon,Jul 18,2016`

**## Your Environment**
* Version used: Commit [0eef97fccbaf7c7f5d97ea518cafe97072f54e19](https://github.com/AHAAAAAAA/PokemonGo-Map/commit/0eef97fccbaf7c7f5d97ea518cafe97072f54e19)
* Environment name: Python 2.7
* Operating System: Windows 10

MonjoNL:
Can we merge this with #1621 ? Same issue. I posted a work around, not completely accurate though but it shows the pokemon.

calebbait:
I'll look at the fix you posted and see if I can splice something together myself.

Issue #1621, opened 2016-07-23
Some pokemons don't show up / show up with a delay


SuperMartas:
Hello,
I noticed that some pokemons don't show up or show up after tens of minutes. I have 2 accounts - one for a game in my phone and the second for searching. The game showed me up 4 pokemons:
![screenshot_20160723-192849](https://cloud.githubusercontent.com/assets/7558984/17079875/c0837cc0-511d-11e6-8663-6f7ecb625c28.png)
Your app printed on a screen that there are 3 of them (`Upserting 3 pokemon`) but showed me up on the map just 1 of them:
![2016-07-23 19_29_40](https://cloud.githubusercontent.com/assets/7558984/17079898/3a9bd3b8-511e-11e6-94f7-d4c2c913719a.png)
I could see the rest of pokemons after about 15 minutes:
![2016-07-23 19_46_12](https://cloud.githubusercontent.com/assets/7558984/17079912/8abdde40-511e-11e6-907b-670e09f54671.png)
Is it a bug or it's common that there may be a delay between 2 accounts? I didn't see it for the first time and I'm pretty sure that there isn't any another spawn point...

OS: Ubuntu 16.04.1 LTS
Python 2.7
Commit: caf4ee8b9bf186af038057894c5cce2a2d70894c

Lenny188:
same problems over here

QuadTog:
Add "-t 1" to your CMD line. Currently threading doesn't log correctly and pokemon are not registering.

SuperMartas:
This happened to me many times when the threading didn't exist and if I use -t 1, nothing will change and the bug is still there, some pokemons are still show up with the delay.

QuadTog:
Threading has existed for a while. It's just normally set to ~5 cores. Last night we were having issues with upsetting Pokemon related to threading. 

Double checking now with an older version to see what happens.

SuperMartas:
I watch almost every commit so I'm pretty sure it were doing in a particular way before the threading existed. But never mind. I tested it again with 1 thread and -dc switch and it luckily did again.
The game in the phone:
![screenshot_20160724-004340](https://cloud.githubusercontent.com/assets/7558984/17080796/d650ff58-513d-11e6-9ead-bd3ca5c22933.png)
But look at these next two screenshots:
![2016-07-24 00_41_33](https://cloud.githubusercontent.com/assets/7558984/17080800/148b053e-513e-11e6-9c06-9f7341f04435.png)
[...]

PR #2346, opened 2016-07-27
Show pokemon with negative disappear time

MonjoNL:
Display special spawns that have a very big negative disappear time but are in reality in the game for a long time.

## Description
Adds a correction that is bigger than the negative time rounded up till the next full hour.
Still trying to find the exact disappear time. In case it is still there it will be automatically added again after the next scan for almost one hour.

## Motivation and Context
We don't want to miss a Pokemon? Do we?

## How Has This Been Tested?
I happen to live close to a spawn point that has this behavior, and Pokemon in that point are present for about one hour. I changed my copy of costomLog.py to keep track of the negative disappear times.

## Types of changes

## Checklist:

See issues #1621 and and #1879

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -3,6 +3,7 @@
 
 import logging
 import os
+import math
 from peewee import Model, MySQLDatabase, SqliteDatabase, InsertQuery, IntegerField,\
                    CharField, DoubleField, BooleanField, DateTimeField,\
                    OperationalError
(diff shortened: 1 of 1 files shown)


hgentry:
This may be of use https://www.reddit.com/r/pokemongodev/comments/4uuwx0/spawn_myths_debunked_aka_the_truth_about_spawn/

AHAAAAAAA:
Is this still an issue? I can't replicate.

PR #2375, opened 2016-07-27
Fixes scan location for high -st

### LINKED ISSUE #2187: Scan locations far from the center are increasingly off
Close to the center the circles of scanned locations align and cover all the ground. But further outside (I tried it with -st 100 and waited 20 minutes) you will get this picture:

screenshot from 2016-07-26 14-42-44

This is likely a small error that gets added again and again.

reyqn:
## Description

I changed how the latitude longitude are calculated. It should be more accurate now.
I also used more accurate values for some constants.

## Motivation and Context

At high -st, longitude was becoming very inaccurate. This change will help with that.
This should fix #2187 

## How Has This Been Tested?

I tested pretty quickly where I live (45° lat), next to the equator (0° lat) and north pole (90° lat). Seems to work pretty well.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -55,6 +55,8 @@ def generate_location_steps(initial_location, num_steps):
  
     # Calculate the length of a degree of latitude and longitude in meters
     lat = math.radians(initial_location[0])
+    # All those numbers are not random. They are often used for calculations like this one and you can understand more on them here:
+    # https://gis.stackexchange.com/questions/75528/length-of-a-degree-where-do-the-terms-in-this-formula-come-from#75535
     meters_per_degree_lat = 111132.92 + (-559.82 * math.cos(2 * lat)) + (1.175 * math.cos(4 * lat)) + (-0.0023 * math.cos(6 * lat))
     meters_per_degree_lng  = (111412.84 * math.cos(lat)) + (-93.5 * math.cos(3 * lat)) + (0.118 * math.cos(5 * lat))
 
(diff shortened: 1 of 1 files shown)


medghaim:
Far too many magic numbers for my liking. I have the fix, just testing it in all quadrants of the earth.

reyqn:
@medghaim https://gis.stackexchange.com/questions/75528/length-of-a-degree-where-do-the-terms-in-this-formula-come-from#75535

reyqn:
But I guess meters_per_degree_lng should be recalculated every 100 steps or so (if you want to scan huge zones)

reyqn:
Seems to be more accurate than #2382

medghaim:
Yeah, I'm not sure what the hell happened, that was the edge of `120` - and a few other tests too, but that GIS page you linked is 100% accurate.

I'm still not a fan of the magic numbers though, they need their own to be their own variables with comments. 

Since GIS capabilities are becoming more and more required evidently, I believe we should just write a GIS module.

reyqn:
Yeah I have no idea why yours is not accurate. It definitely looks better. I made 

I am not really proficient in python, so I have no idea how you would like to implement this GIS module.

marauder37:
Had a look at this but it needs rebasing

AHAAAAAAA:
Is this still necessary?

reyqn:
@AHAAAAAAA It could if people still want (and are able) to scan huge areas. It's still not perfect though. Finding the inaccuracies in #2382 would probably be better.

PR #2382, opened 2016-07-27
Improving step misalignment after huge step numbers

medghaim:
Using a more precise method to calculate _new_ coordinates for usage in the hex generator. 

## Description
New function to handle the calculation of a new pair of coordinates based on an original location, a distance, and a bearing - `get_new_coords(from_loc, distance, bearing)`

## Motivation and Context
This seriously reduces the misalignment in huge step numbers.

## How Has This Been Tested?
With `-st 120` and `150`. In all quadrants of the earth. This doesn't 100% solve things, but seriously reduces the issue (as seen at the edge of this `-st 120` hex):
image

As opposed to the edge of this `-st 100` hex:
image

Note, the warping only happens in 1 direction, due to map projection. 

The remaining tiny gaps are (_i believe_) floating point imprecision, which can be cleaned up.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -31,15 +31,6 @@ log = logging.getLogger(__name__)
 TIMESTAMP = '\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000'
 api = PGoApi()
 
-# Constants for Hex Grid
-# Gap between vertical and horzonal "rows"
-lat_gap_meters = 150
-lng_gap_meters = 86.6
-
-# 111111m is approx 1 degree Lat, which is close enough for this
-meters_per_degree = 111111
-lat_gap_degrees = float(lat_gap_meters) / meters_per_degree
-
 search_queue = Queue()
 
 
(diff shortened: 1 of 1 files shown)


reyqn:
Yeah this looks way better code wise. I'm closing mine.

AHAAAAAAA:
+1 tested, gaps improved as reported.

reyqn:
@medghaim Actually here is mine at -st 100. (I can't really try more because it's taking very long on my laptop) Does it seem better than yours when you are only at 100?
screenfetch-2016-07-28_00-52-19

mwillems1994:
Perhaps you could change the long if statements for a switch, which is generally faster in comparing integers and makes the code a bit more maintainable.

reyqn:
@AHAAAAAAA Why did you merge this? Here are the results of #2375 for 120 steps.

It's near the north pole and at the most inaccurate place (big latitude and longitude difference from origin)

screenfetch-2016-07-28_01-59-50

We can start seeing little holes but nowhere as important as in this PR.

pogom/search.py
```diff
@@ -61,47 +52,61 @@ def send_map_request(api, position):
         log.warning("Uncaught exception when downloading map " + str(e))
         return False
 
+def get_new_coords(init_loc, distance, bearing):
+    """ Given an initial lat/lng, a distance(in kms), and a bearing (degrees),
+    this will calculate the resulting lat/lng coordinates.
+    """ 
+    R = 6378.1 #km radius of the earth
```
reyqn: Try 6378.137 maybe?

PR #2355, opened 2016-07-27
Fix babel transpilation & clean up grunt flow

parkershepherd:
## Description
A previous commit disabled babel, uglify wasn't being applied to the main map.js file, and the grunt flow was inconsistent (some files went to an intermediate `/css` folder, not all source maps were present, etc)

## Motivation and Context

* Makes it easier to keep track of which files are build vs source (not ALL built files are in `/dest`)
* Fixes babel
* Minifies map.js
* Makes sourcemaps work all the way down to original source

## How Has This Been Tested?
Chrome/OSX, Firefox/OSX, Safari/OSX, Chrome/Android

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2274, opened 2016-07-27
Example Docker-Compose w/ Let's Encrypt Proxy

hemsleyk:
## Description
Adds an example docker-compose and Caddyfile for one server, one SQL writer, a MySQL DB, and Caddy for HTTPS termination w/ Let's Encrypt.

$EMAIL must be specified (e.g. when using docker-compose up) elsewise Caddy will prompt for it and stall.

Run commands in docker-compose.yml must receive google maps API key, username, pw.

Nothing is externally exposed port-wise except the proxy. Both locations appear on map as expected.

## Motivation and Context
Compose is good. Multi-instance is good. MySQL is good. Let's Encrypt is good.

## How Has This Been Tested?
Everything EXCEPT https is working. ACME seems to work but chrome refuses to load the page (possibly due to me using free, long scaleway.com subdomain, I have a similar setup on servers at work which functions fine)

If I disable HTTPS in the Caddyfile everything works.

MySQL Performance with Miami / Orlando as test vectors is quite good, lots of lines like "upserted 100 pokemon" and virtually null load.

## Screenshots (if appropriate):
http://imgur.com/a/Xg6uN

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2384, opened 2016-07-27
Removed dist files from Git

Chlodochar:
Another attempt, see #2034

## Description
The files under `static/dist` are being generated by Grunt. They are in gitignore, but are still in git. This is confusing. It adds unnecessary rebasing for PRs. Generated files should generally not be checked into git.

There will be backlash, as this has been tried before in #2034, but was later reverted because people weren't used to running Grunt. Additionally, the Dockerfile was not yet updated.

In this PR the Dockerfile is updated, however I still had trouble with this, so HELP WANTED ON DOCKERFILE.

## How Has This Been Tested?

`docker build .`

## Types of changes

## Checklist:


Chlodochar:
Docker has 2 problems:

* The local installation of grunt cannot be found somehow.
* The Dockerfile isn't as space/buildtime efficient as it should be.

Pedrock:
There is a problem in the gruntfile:
https://github.com/AHAAAAAAA/PokemonGo-Map/blob/6e01bb7189a3f3076b6037e2681d78b14454580b/Gruntfile.js#L43
It should be instead:
'static/dist/js/app.min.js': 'static/js/app.js'

chuyskywalker:
I'm going to kick in with this one as a general grunt cleanup as well -- there's a number of organizational tasks that would make the `static` folder much friendlier and easier to use that I'd like to put into place, if that's alright.

chuyskywalker:
Nevermind me, #2355 specifically address my above

invisiblek:
+1000000000000000000000000

andrewda:
+1000000000000000000000001

chuyskywalker:
https://github.com/Chlodochar/PokemonGo-Map/pull/1 <-- working docker image builds from this

Chlodochar:
@chuyskywalker I've integrated your changes to the Dockerfile. I'll leave cleaning up Gruntfile and others to another PR ;)

chuyskywalker:
LGTM, @andrewda or @invisiblek if you can give that update for the docker file a check over, we can get this merged.

> I'll leave cleaning up Gruntfile and others to another PR ;)
Yeah, #2355

Chlodochar:
Also, @FrostTheFox do you know what else needs to be done to get this change in with minimal backlash?

invisiblek:
@chuyskywalker I dont use docker (never have), can't weigh in on that at all.
@Chlodochar backlash will happen. There's no way around it.

hemsleyk:
+1 for revising the Dockerfile.

@invisiblek you should give it a shot, maybe start with compose if you speak yml config files. My compose at #2274 lets you make a few text edits (credentials), run docker-compose up, and have a two-location MySQL set up with an HTTPS reverse proxy and an A- on qualys. Adding workers is as simple as copy/pasting the whole worker block with different service name, credentials and location.

Ashex:
The docker file throws away layers which is going to result in a bloated image. Take a look at my dockerfile used for building the image `ashex/pokemongo-map:latest`:

https://github.com/Ashex/pokemongo-map-docker/blob/master/tagged/Dockerfile

I spent way too much time on that Dockerfile so it should cover what's needed. You'll need to add in the run-script bit in the RUN command at line 29.

hgentry:
Has anything been done about how the babel section of the gruntification takes half an hour?

Chlodochar:
@Ashex That's what I had first, but I was having trouble with Grunt/jshint not finding the right files. I'll take a look to split things up some more.

@hgentry That's probably sass you're talking about. Installing sass takes a long time because it needs to compile libsass :-/ I might be able  to put that as one of the first steps in the Dockerfile though.

Dockerfile
```diff
@@ -17,16 +17,20 @@ ENTRYPOINT ["python", "./runserver.py", "--host", "0.0.0.0"]
 # Set default options when container is run without any command line arguments
 CMD ["-h"]
 
-# add certificates to talk to the internets
+# Install required system packages
 RUN apk add --no-cache ca-certificates
 
-# Copy Python requirements so we only rebuild deps if they have changed
-COPY requirements.txt /usr/src/app/
```
tkSimon: Removing this means that we wont be able to profit from caching mechanisms provided by docker. It'll need to install all python pavkages on each and every image build.

Chlodochar: Yes, but both steps require build-base and afterwards it needs to be removed. That's why everything 'needs' to be in a single step. We can split it up, but it'll install and remove build-base twice.

Chlodochar:
It's now split into multiple steps. I'm uncertain whether this is better or more efficient. `build-base` needs to be installed and removed twice. There are multiple `COPY` steps which result in multiple caches being made (that fine and dandy, but it seems creating caches takes time as well). There doesn't seem to be a 'best' Dockerfile for this.

Let me know if you have concrete changes that can be made.

invisiblek:
we really need to get this shipped. more and more pull requests are coming in where people are updating static/dist/js/map.js or whatever and not updating the source version

Chlodochar:
You can merge it without me. I won't be at home for a day or so.

invisiblek:
well, if it breaks something (docker) we can't ship it

chuyskywalker:
The docker container for this works fine, the commit for that came from me and was tested.

chuyskywalker:
I can confirm that the docker file produces a working map.

chuyskywalker:
> @hgentry That's probably sass you're talking about. Installing sass takes a long time because it needs to compile libsass :-/ I might be able to put that as one of the first steps in the Dockerfile though.

I was looking around and alpine-edge has libsass as a package. I may try that out later.

chuyskywalker:
As a quick follow up; no alpine-edge with that libsass package did not help. But maybe it wasn't looking in the right place or...something like that. I dunno.

Ashex:
@chuyskywalker edge is the development build of alpine so it's best to avoid using that. The libsaas compiling is incredibly irritating for me too.

We could make building libsaas a separate step at the beginning so that it's cached and not built by `npm install`.

Base it off this image: https://github.com/xzyfer/docker-libsass

Chlodochar:
Using a different image specifically for sass will suck, as we need Python to be a specific version and preferably node too. We can so npm install as a separate step and cache it. It'll take up more space but I think the space is less important than our time.

Ashex:
@Chlodochar Sorry I meant to say incorporate the libsaas build process used in that image. I'll review the Dockerfile and make a PR with my recommendations later on if nobody else does it first.

marauder37:
@Chlodochar the fallout from this PR is huge because the documentation updates were really poorly handled. We've got people running this in Windows, OSX, their android phones, Heroku, EC2, Bluemix, a bunch of other environments. This has screwed a lot of people.

invisiblek:
Then they should use the master branch, not the develop branch. Tracking these files is both confusing and pointless.

Chlodochar:
@marauder37 I'm sorry, but no one was coming up with actual solutions. I consider keeping the dist files a very very poor solution and the dist files were still in `.gitignore`, which doesn't make sense since contributors are required to commit their dist files.

In addition, the number of PRs are growing, partly because the PRs aren't being merged. They aren't being merged mostly because of rebase-issues multiple times per day. Those rebase-issues are in large part due to dist files.

There are many ways in which this project can be installed from develop, but as @invisiblek mentioned most people shouldn't be using develop, as the name most certainly implies. Develop is for people who change the code. Changing JS or CSS will require Grunt anyways.

Instead of relying on develop, releases should be made. Releases should be as-far-as-possible prebuild. That means releases should have dist files. We can release it as zip or .tar.gz and can even make Travis build these automatically.

I agree that running Grunt (and therefore needing to install Grunt) is annoying. We knew there would be some fallout (apparently the Heroku file was forgotten), but that can be fixed with additional PRs. Keeping the files is not an option: the longer we keep it, the more people will rely on it. Hence the fallout.

PR #2397, opened 2016-07-28
Disable caching on GET request to /raw_data

Scyntrus:
Some browsers, prominently on mobile, will cache the request to /raw_data and future requests will not actually query the server. This results in stale data and new Pokemon not being added.

## Description
Added a cache: false flag to the jQuery ajax call.

## Motivation and Context
I frequently encounter a problem where new Pokemon spawns are not being updated when viewing the map using a mobile browser and on data.

## How Has This Been Tested?
Using the map now always requests new data, even on mobile browsers.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -754,6 +754,7 @@ function loadRawData() {
       'neLng': neLng
     },
     dataType: "json",
+    cache: false,
     beforeSend: function() {
       if (rawDataIsLoading) {
         return false;


marauder37:
I'll give it a test, but how could this be wrong?

wcastello:
Works, no cache.

PR #2399, opened 2016-07-28
Preserve animations when marker goes off screen 

jjbott:
## Description
Marker animations are lost when they go off screen. When they're removed from the map, they lose their animation setting. Save their animation setting and restore it when they're redisplayed

## How Has This Been Tested?

Chrome on Windows, scrolled bouncers off screen and back on screen. Should keep bouncing

## Types of changes

## Checklist:


AHAAAAAAA:
+1, good catch.

hhu94:
I am getting `map.built.js:673 Uncaught TypeError: a[b].marker.getAnimation is not a function` with this change.

PR #2400, opened 2016-07-28
Persist location changes across sessions when using search box

sf302:
## Description
Adds remember_location to local storage and defaults to start location passed via HTML. Changing your location using the location search box will persist across sessions and open map up to last searched location. Using the default value functionality from Store ensures behavior does not change if the remember_location variable is unset.

## Motivation and Context
I have multiple searchers running in the background for friends' locations and wanted to be able to have the page automatically open up to their location instead of mine.

## How Has This Been Tested?
Chrome on Win10, tested from clean state (no remember_location var set), verified var was set properly via search box, and fresh pageload worked with saved var.

## Types of changes

## Checklist:

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -116,6 +116,10 @@ var StoreOptions = {
     default: [],
     type: StoreTypes.JSON
   },
+  remember_location: {
+    default: {lat: center_lat, lon: center_lng},
+    type: StoreTypes.JSON
+  },
   showGyms: {
     default: false,
     type: StoreTypes.Boolean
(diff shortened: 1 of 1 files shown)


FlarumHostingCo:
Map doesn't load after patching this in

sf302:
@FlarumHostingCo Thanks for the catch -- looks like I committed an older version that still had lon instead of lng.

FlarumHostingCo:
Love this addition but just found out it broke in a recent develop update.  Could you please fix as some lines no longer match like..

changeSearchLocation(lat, lng).done(function() {
      map.setCenter(loc);
      marker.setPosition(loc);

PR #2401, opened 2016-07-28
Removed locationButton under certain conditions.

daika7ana:
## Description
* Removed locationButton when Geolocation is toggled on, because current
position = worker position anyhow.

## How Has This Been Tested?
Tested on Win10: Edge, Chrome, Firefox.
Tested on Android 6.0.1: Chrome.

before
after

## Types of changes

## Checklist:

### DIFF
--- a/static/dist/js/map.js
+++ b/static/dist/js/map.js
@@ -882,6 +882,10 @@ function myLocationButton(map, marker) {
   locationButton.style.marginRight = '10px';
   locationButton.style.padding = '0px';
   locationButton.title = 'Your Location';
+  locationButton.id = 'location-button';
+  if (Store.get('geoLocate')) {
+    locationButton.style.display = 'none';
+  }
   locationContainer.appendChild(locationButton);
 
   var locationIcon = document.createElement('div');
@@ -1128,6 +1132,11 @@ $(function () {
   });
 
   $('#geoloc-switch').change(function () {
+    if (this.checked) {
+      $("#location-button").hide();
+    } else {
+      $("#location-button").show();
+    }
     $("#next-location").prop("disabled", this.checked);
     $("#next-location").css("background-color", this.checked ? "#e0e0e0" : "#ffffff");
     if (!navigator.geolocation) this.checked = false;else Store.set('geoLocate', this.checked);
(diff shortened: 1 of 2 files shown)


static/dist/js/map.js
```diff
@@ -1237,6 +1240,10 @@ $(function() {
   });
 
   $('#geoloc-switch').change(function() {
+    if (!Store.get('geoLocate'))
+		document.getElementById("location-button").style.display = 'none';
```
georgo: I would recommend use $("#location-button").hide() instead of this.

static/dist/js/map.js
```diff
@@ -1237,6 +1240,10 @@ $(function() {
   });
 
   $('#geoloc-switch').change(function() {
+    if (!Store.get('geoLocate'))
+		document.getElementById("location-button").style.display = 'none'; 
+	else 
+		document.getElementById("location-button").style.display = 'inline';
```
georgo: I would recommend use $("#location-button").show() instead of this.

andrewda: Use the standard coding conventions, please. Use jQuery, as @georgo pointed out, and add curly braces. Also, indent with two spaces. We are going to be strictly enforcing these rules.

willgrnr:
+1

daika7ana:
Cheers for everything that you've said and for the input :+1:  Striving to be better at this. Fixed everything that you've pointed out, retested and rebased. How does it look?

static/dist/js/map.js
```diff
@@ -1237,6 +1241,12 @@ $(function() {
   });
 
   $('#geoloc-switch').change(function() {
+    if (!Store.get('geoLocate')) {
+      $("#location-button").hide();
+    }
+    else {
```
andrewda: Hate to be so nit-picky, but please make the closing curly brace and the `else` statement on the same line. We're trying to keep and maintain a code standard.

static/dist/js/map.js
```diff
@@ -1128,6 +1132,11 @@ $(function () {
   });
 
   $('#geoloc-switch').change(function () {
+    if (!Store.get('geoLocate')) {
```
georgo: This negative condition doesn't make sense (because on the end of this function it'll probably change and it will be confusing.
Use ```this.checked``` instead. ;)

daika7ana: I initially did this, however for some reason this was not working on android default browser. However that being said, now the map doesn't work at all in the default browser but it works perfectly fine in Chrome/Firefox. So I guess the problem lies elsewhere.

AHAAAAAAA:
This has already been merged in it seems, but thanks anyways!

PR #2403, opened 2016-07-28
Mobile Page Part III

chuyskywalker:
## Description

- Fix list so each numeral is read aloud
- New innerText method reads correctly (doesn't get stuck on repeating count downs), have it always on
- Use button as "msg" instead of dedicate element (which wasn't working)
- Also fixed overflow on pokemons in desktop browser that added scrollbars (yuck!)

## Motivation and Context

Realized some of the changes had an unintended effect on the mobile page for VoiceOver usage

## How Has This Been Tested?

Turned on VoiceOver, interacted with page.
Test page on Safari and on desktop chrome

Looks no different, yay!

## Types of changes


chuyskywalker:
@parkershepherd :/ I should have tested your branch better.

parkershepherd:
my bad, didn't have anything with voiceover specifically to test with :/

chuyskywalker:
Not at all, you actually fixed the stupid repeating voice issue by using innerText instead. Totes my fault for not testing it better.

static/js/mobile.js
```diff
@@ -1,39 +1,30 @@
...
+		navBtn.innerText = 'Locating...';
+
+		// Get location and use it!
 		navigator.geolocation.getCurrentPosition(function(p) {
-			loadingMessage.innerText='Reloading...';
+			navBtn.innerText = 'Reloading...';
 			location.href='/mobile?lat='+p.coords.latitude+'&lon='+p.coords.longitude;
 
 		}, function(err) {
 			loadingMessage.innerText='';
```
parkershepherd: Leftover `loadingMessage` reference here causes an error (if a legitimate error occurs)

static/js/mobile.js
```diff
@@ -53,20 +44,19 @@ function updateTimes() {
 	// Yes, this could be a smidge innaccurate, but not by
 	// more than 1 second or so which doesn't matter.
 	// And now we don't have to deal with timestamps and dates!
-	document.querySelectorAll('div.remain').forEach(function(element){
```
chuyskywalker: `forEach()` did not work in mobile safari. I have no idea why, `caniuse` shows it as supported, but...

parkershepherd: hmm, that's odd... I did check the browser support before skipping the traditional `for` loop

chuyskywalker: Yup, I dunno what it was, but good 'ol `alert('wtf, say something');` inside the loop did nothing. ^shrug^

parkershepherd: huh, my first guess would be that the js was just cached or still compiling, but mobile devices do weird stuff. Wish I had an iOS device to mess around with haha

static/js/mobile.js
chuyskywalker: Thanks, fixed.

parkershepherd:
Might want to bump the `li` padding-left from 50 to 55px
image


otherwise everything looks good to me

chuyskywalker:
Quite a few of the evolved creatures were clipped -- this is what happens when I live around nothing buy zubats and pidgeys.

```
{% for i in range(1,152) %}
		<li style="background-image: url('static/pixel_icons/{{i}}.png')">
			<div class="name">{{i}}</div>
		</li>
{% endfor %}
```

To the rescue for testing :)

AHAAAAAAA:
Tested on mobile and several different mobile resolutions, works great for its usecase. I'll note that it looks broken on larger laptop wide-resolutions like this:

screenshot 2016-07-30 11 09 53

Otherwise, it's got my +1

chuyskywalker:
@AHAAAAAAA you have old css in that screenshot, it should look like this:

2016-07-30 10_51_50-nearby pokemon

PR #2498, opened 2016-07-28
Add backoff to login logic

diamondo25:
## Description

The Pokemon login servers are not so fast. When tokens are generated, they are not directly available in other services.

## Motivation and Context

This changes adds backoff to the login logic, improving overall stability and speed when starting up.

## How Has This Been Tested?

On a fast connection and PC, python is too fast for the pokemon servers. Without this change, it would take up to a minute before every part of the login process succeeded. Not anymore, because 99 of 100 times, the first login statement (getting token from PTC & use that) fails and is now succeeding the second try.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/pgoapi/auth_ptc.py
+++ b/pogom/pgoapi/auth_ptc.py
@@ -26,6 +26,7 @@ Author: tjado <https://github.com/tejado>
 import re
 import json
 import requests
+import time
 
 from auth import Auth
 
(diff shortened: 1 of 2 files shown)


diamondo25:
I'm not sure about the code style, however. Most of this was written in a terminal session

andrewda:
Please make this PR upstream in the pgoapi repository. I don't like the idea of changing a lot of the pgoapi logic unless it's done upstream: https://github.com/tejado/pgoapi

PR #2501, opened 2016-07-28
Create robots.txt

avdg:
## Description
Block search engines

## Motivation and Context
Security

## Types of changes
Added http://www.robotstxt.org/robotstxt.html

## Checklist:
- [x] Create robots.txt

### DIFF
--- a/pogom/robots.txt
+++ b/pogom/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /


andrewda:
First of all, robots.txt won't be accessible from that folder without additions to the python code. Second of all, this is unintended for commercial/production use, only as a personal map, so I find this to be useless.

avdg:
Feel free to do what you want, although I have to say I already see a lot of clones popping up on search engines.

avdg:
Not to mention that all these servers randomly makes connection to the internet and are able to be operated remotely without permission at all.

avdg:
Also, I appreciate additional fixes

avdg:
(uh yeah and I didn't take the time to fully checkout this project yet)

avdg:
Actually this problem is partially caused by clones of this project or implementing it and have the intent of sharing information with others.

avdg:
(yeah I had to fix it in their repo, although I prefer to have an alternative fix here too but I don't use python)

sebastienvercammen:
:-1: Clones popping up are not the responsibility of this repo. We have nothing to do with, and don't support, any public scanners, especially ones that use the code on this repository.

chuyskywalker:
There is no reason to have a robots.txt here. If a search-engine (or other crawler) wishes to hit the site, they won't find much and will probably just move on.

PR #2390, opened 2016-07-28
Italian Localization

hilycker:
Actually in italian pokemon names are the same as in english. It's just to let you know as soon you'll decide to make other strings localized, you'll have italian translations too :)

PR #2540, opened 2016-07-29
Italian Localization

hilycker:
fixed #2390 

This time I actually got the original en file. Also saw it was encoded with UTF-8. maybe that's why mine wasn't good.


AHAAAAAAA:
These are the Pokemon names in English. Why do we need this?

callmebackdraft:
@AHAAAAAAA adding this also enables more project wide localization, for instance my PR #2667 which moves the mapstyle selector and implements localization in a custom select bar.

If people/contributors would somehow change the menu options to different languages through localization files it would not be able to show pokemon pokemon names if the pokemon file isnt available for that specific language.

andrewda:
@callmebackdraft or default to English if it's not available :>

Each localization file is ~12kb, sometimes more depending on the language, but that adds up quickly if we were to add a bunch that are literally the same as the English localization.

hilycker:
@AHAAAAAAA  quoting myself from #2390 "[...]It's just to let you know as soon you'll decide to make other strings localized, you'll have italian translations too :)"

and It's to have something to start, to let other people know there's also an ongoing italian localization .

on another repo, and one widely known, weeks without the italian localization passed, as soon as i started it, the following day other 2 translators joined :)

and of course what @callmebackdraft said

tim95030:
This is in the old language format. I switched to using an i8ln function that does word lookup. As such all you need to do is create a dictionary of any words you want translated and if it isn't found it resorts to english. Meaning don't translate Mew -> Mew etc. just translate actual things. For format it is a simple json file:

```
{
  "english word": "translated word",
  "another english word": "another translated word"
}
```

I did this for one to limit space, and for 2 to be able to separate translation from data. There is no more pokemon.{lang}.json just {lang}.json in locales. Also don't bother to minify as a grunt task does this. You can copy one of the other files for reference. But my above example is exactly how it works. It just maps words to words. It can also do phrases as well, but currently only rarity, types, and names are being translated.

AHAAAAAAA:
Update this to the new localization format and I'll merge

hilycker:
@tim95030 @AHAAAAAAA 

will do, thanks

Issue #2202, opened 2016-07-26
map not showing up


Lirang6:
I have done exerything according the Turtuiol and the map still not showing up not mater which browser i choose...

## Your Environment
* Version used: master version
* Environment name and version:  Python 2.7
* Operating System and version: desktop and mobil

GreenflyM:
Having the same problem on mobile, but loading fine in firefox and chrome on desktop, did you try those?

nepomuk13:
I think I have the same problem. Pokemon seems to load according to the console, interface is showing up (filtering options and so on) but not the map in itself

haqthat:
Confirmed in chrome, and firefox, screenshot from chrome.

![image](https://cloud.githubusercontent.com/assets/1305090/17147801/54a4fef4-5332-11e6-8e47-58692b62cf33.png)

Lirang6:
![image](https://cloud.githubusercontent.com/assets/20610887/17149021/18f6c618-5372-11e6-9f06-a240264b2cc9.png)
 
this is my problem... no map at all! what can i do?

ricohumme:
Did you get your own Google Maps API key set up ?
[https://github.com/AHAAAAAAA/PokemonGo-Map/wiki/Google-Maps-API:-a-brief-guide-to-your-own-key](url)

Gaudon:
I frequently have this issue as well, have to ctrl+f5 a lot to get it to work but then no pokemon actually show on the map heh.

devinjparsons:
yeah looks like you need to make sure you have your gmaps api key in there

GreenflyM:
I set up my own API key and even generated a second one with a separate account to test this. I get a blank screen with the options menu no matter what browser I use on ios. It works with chrome and firefox on desktop, but safari displays a blank screen with the options menu also.

ricohumme:
I also have it working on Chrome, but IE is a no go. Only displays the black header and that is as far as it goes. Didn't check the console though to check if any errors are shown.

Gaudon:
I have my own maps api key yeah, and chrome / firefox do not produce errors.. Just get a title menu bar at times and nothing else.. It's very unreliable.

ricohumme:
In IE it shows the error InvalidValueError: initMap is not a function
Probably due to the Google maps API not being correctly loaded somehow

haqthat:
Might have found the issue.

The error we were getting was: 
[...]

Issue #2190, opened 2016-07-26
Mobile site not showing all pokemons


Hoekynl:
In the mobile version of the map */mobile not all pokemons are showing, I think this has to do with the recent addons of lured pokemons showing, etc.

## Expected Behavior
That all pokemons show at the mobile site

## Current Behavior
Now they don't

## Possible Solution
Some sort of code? I personally do not code.

## Steps to Reproduce (for bugs)
Setup dev version, browse to */mobile and see that not all pokemons are showing.

## Context
When im walking, I like to use the mobile version since its click&go to the location via Google maps.

## Your Environment
Rpi 3b
Python 2.7
Mysql 5.5
Latest dev version available with git clone
Proxy via nginx for authentication and HTTPS

Issue #2161, opened 2016-07-26
Pokemon no longer show up when accessing map from Web


Valareos:
## Expected Behavior
Should be able to see pokemon on the screen. Server shows pokemon properly using localhost:5000, my internal server's LAN address, and my routers WAN address (with proper port forwarding set up) but ONLY if i use a browser from within my LAN. Port fowarding is set up correctly, as it was working earlier.

## Current Behavior
accessing my map from an external location does not show pokemon

## Steps to Reproduce (for bugs)
http://120.146.39.206:5000/
http://120.146.39.206:5001/  (This one has a custom edit to the web template to remove the custom location option)

## Context
users in the Jimboomba Rural area do not have the density of pokemon that other places has. this service is to allow them to access pokemon information, either as a small scan, or requesting a bigger scan.

## Your Environment
* Version used: Master d8cdab2
* Environment name and version (e.g. Python 2.7): Python 2.7
* Operating System and version (desktop or mobile):  Server: Windows 7 Desktop

Valareos:
This issue does not seem to appear with every device. Unsure of pattern, but I know a desktop unit earlier  had issue.

Those who have reported the issue has stated it did eventually show up, but only after leaving screen open for 10-15 minutes.

hook54321a:
When you setup your API key, did you restrict what IP addresses could use it?

florian-184-de:
The first map works fine (port 5000)
The second map doesn't work (port 5001). See https://www.dropbox.com/s/ph9igq7wneuc4jg/Screenshot_2016-07-26-11-55-59.png?dl=0
I tested it with Google Chrome 51 (Android; OnePlus One)

monu11296:
port 5000 shows pokemons but takes a lot of time. 
port 5001 didn't show any pokemons.

Plop:
Got the same issue with almost every of my maps.Been stopping/launching the servers, same problem. It just doesn't appear.

![](http://i.imgur.com/suaC62U.jpg)

Valareos:
@hook54321a No I did not limit the api.

The links i gave works sometimes, then dont work, even without restarting server, so I am really unsure of what is goign on. Much rather it happen all the time, lot easier to troubleshoot

Issue #2151, opened 2016-07-26
Downs work with safari


Nuffials:
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
* Environment name and version (e.g. Python 2.7):
* Operating System and version (desktop or mobile):

123456abcdef:
Do you use Safari on desktop or mobile. Both works for me, check that you disabled private mode. Otherwise no Pokemon will show up, because locale storage is blocked.

GreenflyM:
Map isn't working in safari on both mobile and desktop for me. Works fine in firefox and chrome on desktop. Doesn't work in any browser on mobile.

Nuffials:
Same as you

straypccat:
Same:/
Please help me

123456abcdef:
You could check your browser console for errors. It is also possible on mobile devices [link](https://blog.idrsolutions.com/2015/02/remote-debugging-ios-safari-on-os-x-windows-and-linux/).

coolaj86:
I'm working on a version that will work on older versions of Safari here:

* https://github.com/Daplie/node-pokemap

It's a node.js server that uses the same web ui, but is easy to install and none of the problems with installing python, compiling c code, dealing with threads, ports are already open, single-user, setting up https, etc. Super easy and simple.

Pre-Built Packages
------

It even has pre-built downloadable packages:

* https://github.com/Daplie/pokemap-webapp

(though windows is untested, but I'd love to know if the windows-pokemap.bat works)

Disclaimer
----

I haven't merged in all of the UI updates from this main repo yet, but they're coming. Also, my version still has a few things that may not work in Safari right now, but open up an issue and I'll fix it as soon as I can.

PR #2542, opened 2016-07-29
Readd async to Google Maps JS load

leftmostcat:
## Description

This sets the &lt;script&gt; tag for the Google Maps JS API to load asynchronously.

## Motivation and Context

Without the async attribute, there are a variety of timing issues with map loading which may prevent the map from being properly created, scanned locations from appearing, or markers (e.g., Pokémon) from showing. The following issues all appear to be related to this problem:

https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1905
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1907
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2046
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2110
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2121
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2151
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2161
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2190
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2202

## How Has This Been Tested?

This has been tested in Firefox for Linux and OSX and Chrome for OSX. Additionally, it has been tested in Chrome for iOS and Android, Firefox for Android, and Safari on iOS. The map has been heavily used over the course of two or three days with this change to verify that it resolves previously-seen timing problems.

## Types of changes

## Checklist:

### DIFF
--- a/templates/map.html
+++ b/templates/map.html
@@ -179,6 +179,6 @@
     </script>
     <script src="static/dist/js/map.js"></script>
     <script src="static/dist/js/stats.js"></script>
-    <script defer src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places,geometry"></script>
+    <script async defer src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;callback=initMap&amp;libraries=places,geometry"></script>
   </body>
 </html>


chuyskywalker:
This shouldn't, actually, fix anything. If anything, it'd make things worse. _(And, as a side note, several of those tickets are definitely not related to any script loading issues.)_

Why worse? The goal of `async` option in this tag, as stated on [their site](https://developers.google.com/maps/documentation/javascript/tutorial):

> The async attribute lets the browser render the rest of your website while the Maps JavaScript API loads. When the API is ready, it will call the function specified using the callback parameter.

is too let the browser load the maps api in the background because, usually, the map is a small portion of the page, but a heavily load. By using the `async` tag you inform the browser to carry on and load it when it can. Without the `async` tag, the browser says "HALT, DOWNLOAD THIS AND MAKE IT SO RIGHT HERE".

Adding the `async` tag would actually lead to *more* timing problems because the map download would cause the map to be available and much weirder times _and_ since we're not really using the callback to "start up" the map, functions which run on page load (ie, in `$(function()...`) could end up trying to use the map even though it's not ready. As it is right now, those methods won't trigger until the map _is_ loaded and ready.

jangrewe:
Considering that the map is the central element and pretty much only reason for the page, i'd agree that `async` is a rather bad idea.

invisiblek:
We're not going back to this. Removing this made my map load consistently and I haven't had the issue since.

invisiblek:
All the iOS issues linked above appear to be related to localStorage and incognito mode.

PR #2572, opened 2016-07-29
Added custom config file argument

Pho3nixHun:
Added new argument to be able to supply custom config files.

## Description
Added -cf and --config-file argument to supply custom config files.

## Motivation and Context
If you would like to run more searchers, it is handy to provide a custom config file for each.

## How Has This Been Tested?
Supplied a config file by this argument and checked if it's used by the server. (It does)

image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -31,9 +31,8 @@ def verify_config_file_exists(filename):
 
 def get_args():
     # fuck PEP8
-    configpath = os.path.join(os.path.dirname(__file__), '../config/config.ini')
-    parser = configargparse.ArgParser(default_config_files=[configpath])
-    parser.add_argument('-cf', '--config-file', is_config_file=True, help='Config file path')
+    parser = configargparse.ArgParser()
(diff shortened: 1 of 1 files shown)


invisiblek:
Nice find, I was actually thinking about writing this today.

One observation I did have was that the way this is, the user-supplied config file works as a supplement to the standard config/config.ini. In my opinion it should replace it wholesale. Thoughts?

I think this could be achieved by removing line 34, the param in line 35, and setting os.path.join(os.path.dirname(__file__), '../config/config.ini') as the default for this new config option.

Pho3nixHun:
I like your solution too it's way clearer, but this way if you modify only certain values - eg. location - extending a default config file is much more efficient :)

invisiblek:
Right, but that can get pretty confusing. Which file is taking precedence? If you've specified a file, why is it still reading the default (I'm not sure of any other program that does this)

chuyskywalker:
Do not have configs merge, that's super confusing.

AHAAAAAAA:
not merging against master. pr to dev

PR #2596, opened 2016-07-29
Removed beehivebuilder.sh

commanderjsoc:
## Description

I removed beehive builder.sh

## Motivation and Context

This script is useless and I want to keep the codebase clean

## How Has This Been Tested?

N/A

N/A


marauder37:
The use of that script is to help people navigate a refactoring that's less than an hour old, so that when they're following outdated documentation they can figure things out. Or when they do the thing that worked this morning and doesn't work now, they don't get stuck. I'm all for a tidy codebase but leave this in place for a few days.

commanderjsoc:
Sounds good!

Issue #2121, opened 2016-07-26
Can't View Map on Mobile Devices


GreenflyM:
I'm running the master version on Ubuntu 16.04.1 x64 on a digital ocean droplet. Everything works perfectly on desktop, but I'm unable to load the map when accessing the site from ios or android. Is this a known issue or do I have a configuration error? 

You can view the map here: http://0pokescript.findpoke.net:5000/

I'm entering the following in the console:

python runserver.py -u Username -p password -st 13 -k googleapikeyhere -l "53.32686446 -6.24349594" -H 104.131.7.30

rneu31:
+1

rneu31:
I clicked "options" and it eventually loaded. Not sure if it was just taking a while to load images or whatever.

aanc:
Map load without problem on my Nexus 5 and on my wife's iPhone 6.
Deployed on my home server in Docker.

Could it be related to a connectivity issue of your device ?

bdeschut:
Works fine on my samsung s5

domingguss:
the `Options` side menu does open, but doesn't close anymore (in an iOS app inside a UIWebView (Safari))

GreenflyM:
@aanc Any thoughts on how I could troubleshoot the iphone?

Lops88:
I have the same problem. Tried to run the server from my Ubuntu 15.10 VM. Everything looked fine. Could access the site via wifi from my phone using a dyndns-service. However I was not able to connect when I was using my mobile internet. Running the server now from my Host Win10 system and everything is working. Performance is quite a bit worse on Win10 compared to an Ubuntu VM.

aanc:
@GreenflyM, is your server currently running ? I cannot load http://0pokescript.findpoke.net:5000/ on the desktop, i get connection refused.

Can you specify if it is only the map you can't see (the Option bar displays correctly, but the map stay blank), or if you can't connect at all (total blank page, or error message) ?

GreenflyM:
@aanc Thanks for getting back to me. The server was offline, but is live again if you want to take a look. 

You are correct in understanding that the option bar displays correctly, but the map stays blank.

aanc:
@GreenflyM, I can load the map using your server on my Nexus 5. I'll give it a try tonight with my wife's iPhone (if i don't forget).

Can you load http://maps.google.com successfully on your phone ?

GreenflyM:
[...]

Issue #1907, opened 2016-07-24
No markers appear on Mobile browsers.


tashonleo:
## Expected Behavior
Pokemon, stops, gyms etc should appear on the map as they do on desktop. It worked on my early versions!

## Current Behavior
Everything works fine on desktop but the map is completely empty on my iPhone. No markers or anything.

## Steps to Reproduce (for bugs)
Setup instances as per instructions.
Visit url on mobile device.

https://tashon-pgo.herokuapp.com/ <--- New version, empty map on mobile.
http://e2de9901.ngrok.io/ <--- Old version, works fine.

## Context
I'd like to be able to use this while out and about. Yes, the old version works fine but the new version doesn't.

## Your Environment
* Version used: Master as of 25/07/16 00:48 GMT (19:48 EST)
* Environment name and version (e.g. Python 2.7): Heroku Deployment as well as a Python 2.7 setup on Win 7 locally.
* Operating System and version (desktop or mobile): iOS 9.3

Alok-Sharma:
Your link works fine for me, on mobile. There's a switch for displaying Pokemons in the options menu, are you sure you activated that?

Aphit:
@tashonleo In my experience yesterday and today using this application, I had to use Chrome application on my iphone 6 in order to see any pokemon on the map.  For some reason, Safari wouldn't show them.

tashonleo:
I've found out what's going on! The newer versions don't work well with incognito mode/in private browsing on mobile browsers it seems. Switched back to normal browsing and all is well!

I'll take a look at the html tomorrow and see if I can get it to work on both modes.

Aphit:
@tashonleo I had that problem on 1.0 as well and it fixed it, but it seems like Safari still does not work well on this version on my iPhone 6.

kgurnani96:
@tashonleo this happens because iOS safari (or well any other safari for that matter) doesn't allow localstorage to be used in private/incognito mode. A quick look at the console in the safari debugger shows this `[Error] QuotaExceededError: DOM Exception 22: An attempt was made to add something to storage that exceeded the quota.` thrown.

invisiblek:
that patch should allow the javascript to continue if localStorage isn't supported in your browser (which apparently is the case with safari + incognito?)

please test

PR #2599, opened 2016-07-29
Don't attempt to utilize local storage if its not supported

invisiblek:
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/1907
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2121

### DIFF
--- a/static/map.js
+++ b/static/map.js
@@ -27,6 +27,8 @@ var pGoStyleNoLabels=[{"featureType":"landscape.man_made","elementType":"geometr
 
 var selectedStyle = 'light';
 
+var localStorageSupported = true;
+
 var map_data = {
   pokemons: {},
   gyms: {},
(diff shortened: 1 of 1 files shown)


marauder37:
What platforms need to test this?

invisiblek:
iOS safari in incognito mode apparently? The js bails without this and you get no map or no markers or something (see the linked issues above)

kgurnani96:
just tried it, its not working.

static/map.js
```diff
@@ -1248,3 +1260,12 @@ $(function() {
       Store.set('geoLocate', this.checked);
   });
 });
+
+function isLocalStorageSupported()  {
+  try {
+    window.localStorage.setItem("test", "1");
```
codeniko: Why don't you remove the item right after incase local storage is supported?

invisiblek: done

kgurnani96:
@invisiblek i finally got time to check this better, first of all, you have initialised `localStorageSupported` to `true`, which actually should be the value that the function `isLocalStorageSupported()` returns. This causes that `localStorageSupported` is `true` when `initMap()` is called. I've tried setting it to false by default to know what happens, and the map doesn't even load, also `map.js:1076 TypeError: null is not an object (evaluating 'currentLocation.style')` is thrown.

Chlodochar:
Sorry for the late review, but this doesnt make use of the default values of options at all. In addition it should save the value for the current session (locally in an object). It might be best to save values inside the Option object upon `set` and retrieve first from Option and if that's empty, retrieve and parse from localstorage, if that fails return default on `get`.

PR #2616, opened 2016-07-29
Handle InternalError, log it and retry

invisiblek:
potential fix for the deadlock issue: catch and retry on _any_ exception

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -13,6 +13,8 @@ from playhouse.shortcuts import RetryOperationalError
 from datetime import datetime, timedelta
 from base64 import b64encode
 
+from pymysql.err import InternalError
+
 from . import config
 from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, get_args, send_to_webhook
 from .transform import transform_from_wgs_to_gcj
@@ -354,7 +356,7 @@ def bulk_upsert(cls, data):
         log.debug("Inserting items {} to {}".format(i, min(i+step, num_rows)))
         try:
             InsertQuery(cls, rows=data.values()[i:min(i+step, num_rows)]).upsert().execute()
-        except OperationalError as e:
+        except (OperationalError, InternalError) as e:
             log.warning("%s... Retrying", e)
             continue
 


Chlodochar:
Why not one or more specific excepts? Log not only retrying, but also why it is retrying. It'll otherwise be hard to debug such issues.

Chlodochar:
Nice. Didn't know you could catch both of them in one line. This looks good :+1:

andrewda:
:+1: LGTM

Chlodochar:
Landscape.io says it cannot find `InternalError`. I looked online, but it seems that's one from MySQL?
From what I could gather this might need:
```
from mysql.connector.errors import InternalError
```

invisiblek:
you'd think it wouldn't run if it couldn't fin InternalError =/

Pedrock:
continue goes to the next iteration. This does not retry the query.

invisiblek:
@Pedrock i never gets incremented

parrots:
Running this locally also says that it cannot find InternalError, but that message doesn't appear until the actual crash happens. +1 on the import likely being needed.

Pedrock:
@invisiblek Sorry, I didn't notice it was being incremented at the end of the loop.

invisiblek:
@parrots weird that I'm not seeing this.

Does `from mysql.connector.errors import InternalError` fix it?

parrots:
@invisiblek doing that yields `ImportError: No module named mysql.connector.errors` on initial run.

Chlodochar:
I was apparently looking at the wrong library earlier.

Figured it out:
```
import pymysql
```
and
```
except pymysql.err.InternalError:
```

invisiblek:
but...that still doesn't answer why i dont need it =/

AHAAAAAAA:
Same errors as above, we will need to include the necessary libraries. Strange that you don't need it invisiblek.

invisiblek:
I did end up getting that error when it threw the exception. Odd that it wouldn't complain about it to me by simply running the script.

The above patch should fix it, but I'll need to wait around until it happens to be sure.

Chlodochar:
@invisiblek 

Try:
```
try:
    print('This results in no error')
except NonExistingError:
    print('error')
```
and:
```
try:
    print('This results in an error' + (0/0))
except NonExistingError:
    print('error')
```

invisiblek:
current patch didnt catch the error

i'm inclined to catch all exceptions, print them, and retry like i did before. tired of dealing with this

Chlodochar:
@invisiblek It seems like the only option atm. At least it would be a step forwards.

`InternalError` was probably not from MySQL, but something else :/ I'm not sure how the actual type of an error can be retrieved in Python.

tkSimon:
not a python dev, but found this: http://stackoverflow.com/a/9824050

Chlodochar:
@tkSimon Yes, catching exceptions non-explicitly is bad. However, how do we know which specific exception we need to catch? We know the exception has the name `InternalError`, but we don't know how to refer to that specific exception. Python doesn't seem to have a concept of namespacing, so the full-name of `InternalError` cannot be retrieved. Looking online where `InternalError` was used directed me to MySQL, which was why I suggested using `pymysql.err.InternalError`. It seems now though that `pymysql.err.InternalError` is a different `InternalError` than the one that is thrown.

invisiblek:
https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2687

PR #2620, opened 2016-07-29
Added password protection to map.html Develop

rezzj:
Password protection to prevent excessive traffic or unauthorized access.

## Description

Basic HTTP authentication was added to flask.
In contrast to PR #2356 the password is optional and is not hardcoded Username is admin by default, but changeable with -pu. Password is optional and can be set from the command line as an argument with -pp.
## Motivation and Context
Several persons have shown interest to have a simple way to protect their deployment. This is by no means meant for production, the authentication is done over http and IS NOT ENCRYPTED.

I run this app on an exposed raspberry Pi and did want unauthorized users to be able to access the page.

I access the Pokemap from my mobile not just at home and I need it to be reachable from internet not just from private LAN.

## How Has This Been Tested?
I have tested several cases:
case 1:
A new password is passed as an argument to python runserver.py and then checked, then:
no username no password check
correct password only, correct username
wrong password, correct username
correct username and password

case 2: 
A new password and username are passed as command line arguments, then:
no username no password check
correct password only, correct username
wrong password, correct username
correct username and password

Case 3:
Only a new password was passed to the command line then:
no username no password check
correct password only, correct username
wrong password, correct username
correct username and password

Case 4:
No new username or password passed to the command line,then:
simply opening the url 

screen shot 2016-07-27 at 21 12 45
screen shot 2016-07-27 at 21 13 00
## Types of changes

## Checklist:

- [ 
screen shot 2016-07-27 at 21 12 45
] I have updated the documentation accordingly.

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -4,15 +4,38 @@
 import calendar
 import logging
 
-from flask import Flask, jsonify, render_template, request
+from flask import Flask, jsonify, render_template, request, Response
 from flask.json import JSONEncoder
 from flask_compress import Compress
 from datetime import datetime
 from s2sphere import *
 from pogom.utils import get_args
+from functools import wraps
 
 from . import config
 from .models import Pokemon, Gym, Pokestop, ScannedLocation
+
+def check_auth(username, password):
+    """This function is called to check if a username /
+    password combination is valid.
+    """
+    return username == 'admin' and password == 'secret'
+
+def authenticate():
+    """Sends a 401 response that enables basic auth"""
+    return Response(
+    'Could not verify your access level for that URL.\n'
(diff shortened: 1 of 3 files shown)


MRokas:
It should be optional.

andrewda:
This has been mentioned numerous times before. If you would like to add a password, I believe it should be done through another service to prevent this code from becoming too cluttered.

chuyskywalker:
Authentication should be layered on top of the application in much the same way as SSL. Thank your for the effort, but @AHAAAAAAA 's [last statement](https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2356#issuecomment-235732500) on not wanting auth in the application is quite clear and has not changed since.

PR #2637, opened 2016-07-29
Fixed webhook to push time as UTC object

kvangent:
## Description

Changed line 270 to `'disappear_time': calendar.timegm(d_t.timetuple())` - time.mktime() sends the object as a local timestamp, even though it was UTC time. this is causing issues with the times sent at webhook when attempting to convert it to a UTC time.

## Motivation and Context

I have a 3rd party app that is having issues.

## How Has This Been Tested?

Ran it locally.

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -3,7 +3,7 @@
 
 import logging
 import os
-import time
+import calendar
 from peewee import Model, SqliteDatabase, InsertQuery,\
                    IntegerField, CharField, DoubleField, BooleanField,\
                    DateTimeField, OperationalError, create_model_tables
@@ -270,7 +270,7 @@ def parse_map(map_dict, step_location):
                     'pokemon_id': p['pokemon_data']['pokemon_id'],
                     'latitude': p['latitude'],
                     'longitude': p['longitude'],
-                    'disappear_time': time.mktime(d_t.timetuple()),
+                    'disappear_time': calendar.timegm(d_t.timetuple()),
                     'last_modified_time': p['last_modified_timestamp_ms'],
                     'time_until_hidden_ms': p['time_till_hidden_ms']
                 }


pogom/utils.py
```diff
@@ -207,7 +207,7 @@ def send_to_webhook(message_type, message):
 
         for w in webhooks:
             try:
-                requests.post(w, json=data, timeout=(None, 1))
+                requests.post(w, json=data, timeout=(None, 5))
```
andrewda: Why does this need to be changed? That's the timeout for the response, which we're not doing anything with. A value of `1` means it won't hold up the application from sending out more webhooks when it's not necessary because, again, we don't need to wait for the response.

pogom/models.py
```diff
@@ -4,6 +4,7 @@
 import logging
 import os
 import time
```
andrewda: We can also remove `import time` as it's no longer being used.

xxzone:
Hey, thank you so much for your work on this. I wouldn't have figured out the webhook stuff on my own without a lot of work. Most appreciated!

kvangent:
@andrewda I changed the socket time to 5 because 1 was causing socket problems when my application attempted to send a response back. I am using a flask microframework (just like the map), which doesn't allow control over the server stack. Other developers will also have this problem, and I don't feel that it is unreasonable to up to 5. If something happens and the server can't respond, it will only hold up 1 hook (because the hooks after that should be denied upon trying to establish a new connection). 

Another possible solution would be to spin off a new thread to send the request instead. I can make this change if you prefer.

kvangent:
@andrewda Landscape should be happy now, I've removed the time import. If possible I would appreciate a correction for at the very least the time issue quickly, as it is causing issues with other applications.

hhu94:
Can we get this merged please?

kvangent:
@andrewda I've removed the change to the post request timeout.

andrewda:
:+1: from me - need one more +1 from another collab.

PR #2638, opened 2016-07-29
Update map.html

Draco18s:
Tweak to sidebar HTML

## Description
Tweak to the HTML to make the "change location" input field line wrap nicely to fit inside the sidebar when running in Firefox.

## Motivation and Context
Change Location input field was nearly invisible, as it was pushed out of the sidebar with overflow:hidden, preventing access to the feature's UI.

## How Has This Been Tested?
Used Firebug to modify the HTML on the fly, double-checked results in Chrome.

http://s31.postimg.org/4z4lzaj9n/original.png
http://s31.postimg.org/oyz08tv63/fix.png

## Types of changes

## Checklist:

### DIFF
--- a/templates/map.html
+++ b/templates/map.html
@@ -43,8 +43,8 @@
         <div class="form-control switch-container" style="display:{{is_fixed}}" >
           <label for="next-location">
             <h3>Change your location</h3>
-            <input id="next-location" type="text" name="next-location" placeholder="Change your location">
           </label>
+          <input id="next-location" type="text" name="next-location" placeholder="Change your location">
         </div>
         <div class="form-control switch-container">
           <h3>Follow location</h3>


templates/map.html
```diff
@@ -43,8 +43,8 @@
         <div class="form-control switch-container" style="display:{{is_fixed}}" >
           <label for="next-location">
             <h3>Change your location</h3>
```
brycedorn: why not just move the `h3` above the label like the other controls?

Draco18s:
The <label> tag's purpose is to say "this other piece of HTML, when clicked, puts the focus on the named Input element."  so the <h3> is that other element.  The other controls have additional related elements.

AHAAAAAAA:
Tested on a few browsers and mobile screens, +1. Merging with just 1 +1 as this is a small change.

templates/map.html
chavochavo2015: Am0

PR #2665, opened 2016-07-29
Support for spreadsheet-compatible data output file via -dt argument

justanironic:
## Description
I added a command-line argument for runserver.py to output data into a text file in the format of PokemonName,Lattitude,Longitude,SpawnTime,SpawnDayOfTheWeek,SpawnDate,SpawnYear,DespawnTime
The argument is as follows: `-dt filename.extension`
An example of this would be `-dt data.txt`

## Motivation and Context
I am personally not very fond of databases, and wanted to have an output format using comma delimiters I could easily copy and paste into a spreadsheet and get various pieces of data from it, such as distance of the Pokemon from the scanning point and percentage of spawns.

## How Has This Been Tested?
I've been running this code and variations of it for about a week now, and it hasn't given any issues. I run it alongside the `-dc` argument to make sure it outputs the proper Pokemon, and indeed it does.

## Screenshots:
An example of the output of the command `python runserver.py -st 7 -l "washington dc" -dc -t 3 -sd 1 -dt DC.txt`
Example output

## Types of changes

## Checklist:

**If someone could provide clarification on each option within the checklist I would appreciate it**

### DIFF
--- a/pogom/customLog.py
+++ b/pogom/customLog.py
@@ -11,17 +11,17 @@ args = get_args()
 #elif args.only:
 #    ONLY = [i.lower().strip() for i in args.only.split(',')]
 
-def pokeData(id,lat,lng,despawntime):
+def pokeData(pokemon,lat,lng,despawntime):
     if args.datatext:
-        pokemon_name = get_pokemon_name(id).lower()
-        pokemon_id = str(id)
+        pokemon_name = get_pokemon_name(pokemon).lower()
         data = True
         if data:
             spawntime = datetime.datetime.now() - datetime.timedelta(seconds=(900 - despawntime))
             desptime = datetime.datetime.now() + datetime.timedelta(seconds=despawntime)
             datafile = args.datatext
             with open(datafile, "a") as myfile:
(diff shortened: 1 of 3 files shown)


chuyskywalker:
I can almost guarantee this won't get merged -- it's far to specific and narrow a use case, plus should be readily available by writing an SQL script.

justanironic:
@chuyskywalker But what's wrong with being specific and narrow? Not everyone has the knowledge (myself included) to write an SQL script, nor do many people have the patience and motivation to learn it. I figured that for people who use spreadsheets it would be a nice and simple addition.

Ashex:
Consider that this isn't very "scalable", this only writes the data for the worker running. If you've got multiple workers you'll be generating multiple files that would have to be merged together (I'm running 19 right now). This data is already being collected in a database that can be queried. We're also talking about a basic select statement which is not difficult to figure out.

If you're not familiar with SQL scripting you can also simply look up how to do it.

http://lmgtfy.com/?q=mysql+script+generate+csv&l=1

chuyskywalker:
As another option to sql, there is a webhook for pokemon notifications (which is scalable across multiple threads) and you could make a very simple nodejs/python endpoint to receive the WH post and console.log the pokemon info.

The primary thing about adding it is that then we have to keep supporting it. Testing it, making sure we're not breaking the additional feature which few people know about or use. This is why the webhook or sql is much better -- if a small subset of users want this functionality, it can be added on much easier and without making the primary application development harder.

justanironic:
@Ashex You can make multiple instances output to the same file, you just have to make sure they are all set to the same name. No merging required.
Guess it doesn't really matter though, as it's been denied.

Issue #2628, opened 2016-07-29
[Request] Modify map.js to load current user location instead of mapped location


pir2:
## Expected Behavior
Not expected behaviour, but potential enhancement.  Have the option to start the default view at the user's current location

## Current Behavior
The map currently loads at whatever the starting location of the search bot is set at

## Possible Solution
Modify `initMap` in `maps. js` with the following:

```
function initMap() {
  map = new google.maps.Map(document.getElementById('map'), {
    center: {
      lat: center_lat,
      lng: center_lng
    },
    zoom: 16,
    fullscreenControl: true,
    streetViewControl: false,
    mapTypeControl: true,
    mapTypeControlOptions: {
      style: google.maps.MapTypeControlStyle.DROPDOWN_MENU,
      position: google.maps.ControlPosition.RIGHT_TOP,
      mapTypeIds: [
        google.maps.MapTypeId.ROADMAP,
        google.maps.MapTypeId.SATELLITE,
        'nolabels_style',
        'dark_style',
        'style_light2',
        'style_pgo',
        'dark_style_nl',
        'style_light2_nl',
        'style_pgo_nl'
      ]
    },
  });

// Additional Code for current location stolen from map.js
 if (navigator.geolocation) {
      navigator.geolocation.getCurrentPosition(function(position) {
        var latlng = new google.maps.LatLng(position.coords.latitude, position.coords.longitude);
        locationMarker.setVisible(true);
        locationMarker.setOptions({
          'opacity': 1
        });
        locationMarker.setPosition(latlng);
        map.setCenter(latlng);
      });
    } 
```

## Context
Currently running a server for multiple users and most searchers are run in `-ns` mode.  The searchers currently map the whole city, but some of the designated locations are actually the user's marker so that they can map their own/current location in case it's not mapped already, or one of the searchers is down.

Users are clicking the `follow location` button to get to their current location.  If the map started at their current location instead, there would be less user interaction required.

(yes, the user could just press the little circle in the bottom right, but if it just loaded in the current location, most end users would be happy!)

invisiblek:
i'm not even sure we'd add something like this as an option. if it can detect a user's whereabouts, it should go there in my opinion

jangrewe:
[...]

PR #2668, opened 2016-07-29
Open map at user's location

invisiblek:
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2628

### DIFF
--- a/templates/map.html
+++ b/templates/map.html
@@ -43,12 +43,12 @@
       <nav id="nav">
         <div class="form-control switch-container" style="display:{{is_fixed}}" >
           <label for="next-location">
-            <h3>Change your location</h3>
+            <h3>Change scan location</h3>
             <input id="next-location" type="text" name="next-location" placeholder="Change your location">
           </label>
         </div>
         <div class="form-control switch-container" style="display:{{is_fixed}}" >
-          <h3>Follow location</h3>
+          <h3>Scan follows location</h3>
 	  <div class="onoffswitch">
 	    <input id="geoloc-switch" type="checkbox" name="geoloc-switch" class="onoffswitch-checkbox" checked/>
 	    <label class="onoffswitch-label" for="geoloc-switch">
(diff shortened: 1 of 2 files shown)


invisiblek:
toggle added

templates/map.html
```diff
@@ -43,12 +43,12 @@
...
         <div class="form-control switch-container" style="display:{{is_fixed}}" >
           <label for="next-location">
-            <h3>Change your location</h3>
+            <h3>Change scan location</h3>
             <input id="next-location" type="text" name="next-location" placeholder="Change your location">
           </label>
         </div>
         <div class="form-control switch-container" style="display:{{is_fixed}}" >
-          <h3>Follow location</h3>
+          <h3>Scan follow location</h3>
```
Bitti09: doesn't

> Scan follows location


sound better?

invisiblek: Nothing sind good here... Yours is better though

templates/map.html
```diff
@@ -57,6 +57,16 @@
 	    </label>
 	  </div>
 	</div>
+        <div class="form-control switch-container" >
+          <h3>Start at user's location</h3>
+	  <div class="onoffswitch">
```
andrewda: Small indentation issue here and on lines 65/66. Other than that LGTM :+1:

invisiblek: omg

not tabs OR spaces
tabs AND spaces

i have lost the will to live

Slivo-fr:
You should remove the toggle if the -fl flag is used

invisiblek:
@Slivo-fr this doesn't actually change the scanner location (which is what -fl is there to prevent). This simply adjusts the starting point of the map if the user chooses to have it on.

Slivo-fr:
What do you mean starting point ? Does it change the marker location ?
Does the scan happens elsewhere than the location used when starting the python script ?

PR #2610, opened 2016-07-29
Fix scan radius to be 70 metres (adapt to API change)

EarthlingKira:
API changed scan radius from 100m to 70. This commit adapts the code so that all pokemon show up again. Fixes several issues I think.

At least this one:
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2569

regards, Kira

PR #2682, opened 2016-07-30
Change to 70m cells

marauder37:
You will need to run grunt build again 

This is @earthlingkira 's #2610 but consistent with new js file locations

### DIFF
--- a/Tools/Hex-Beehive-Generator/location_generator.py
+++ b/Tools/Hex-Beehive-Generator/location_generator.py
@@ -2,6 +2,7 @@ import math
 import argparse
 import LatLon
 import itertools
+import os
 
 parser = argparse.ArgumentParser()
 parser.add_argument("-lat", "--lat", help="latitude", type=float, required=True)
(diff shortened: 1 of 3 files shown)


chuyskywalker:
I'm giving this a try; the map part anyway

chuyskywalker:
I think this is working. If you look at this setup, The lower right you can see at that park there is a paras, ekans, and ratta that *should* be there, but got missed because the scan shows 100m, but really only covered 70.

2016-07-29 21_39_18-pokemon go map
 

LGTM

Tools/Hex-Beehive-Generator/location_generator.py
```diff
@@ -15,18 +16,17 @@
...
-preamble = ""
+preamble = "#!/usr/bin/env bash"
 server_template = "nohup python runserver.py -os -l '{lat} {lon}' &\n"
 worker_template = "sleep 0.5; nohup python runserver.py -ns -l '{lat} {lon}' -st {steps} {auth} &\n"
 auth_template = "-a {} -u {} -p '{}'"  # unix people want single-quoted passwords
 
 R = 6378137.0
-
-r_hex = 75.0
+r_hex = 52.5  # probably not correct
```
reyqn: You can't really do more correct though.

PR #2687, opened 2016-07-30
Fix #2686

### LINKED ISSUE #2686: MySQL transaction error perma-crashes scanner
When running 7 scanners using `-sd 0.75 -st 8 -t 2` into a MySQL database, occasionally there will be a race condition in which there is an error obtaining a lock on the DB. When this happens the instance that experiences the exception stops scanning.

This is using the current development branch.

## Expected Behavior
The instance should either retry the database update, or drop the update and just move on to the next location. Either way, it should not prevent the next step in the scan hex.

## Current Behavior
An exception happens, killing future scans for that instance.

## Possible Solution
#2616 is close. Replacing that line with `except Exception as e:` fixes it.

## Steps to Reproduce (for bugs)
1. Set up multiple workers (~7) using a shared MySQL database
2. Not sure what exact timing profile is required, but launching them all at once using `-sd 0.75 -st 8 -t 2` options triggers it reliably for me within 10 min
3. Start the worker instances and let them run
4. Watch as eventually a worker instance hits the transaction error and eventually stops updating the DB

## Context
Trying to scan a region using multiple worker threads

## Your Environment
* Version used: Dev c5fe7d6fc0e0e4c5015b29d9d0006a47ab3d57db
* Environment name and version (e.g. Python 2.7): Python 2.7.11, Ubuntu instance on DigitalOean
* Operating System and version (desktop or mobile): Desktop

parrots:
Expand the scope of exception handling to include transactional error cases.

## Description
Sometimes (as per #2686) a transaction lock error can occur. Instead of failing gracefully, the current implementation just crashes and prevents future updates from the scan instance.

## Motivation and Context
Solves a crash.

## How Has This Been Tested?
Tested via running on my local instances for half a day to ensure the crash (which normally happens w/ 6 scanners in 10 min) doesn't happen again.

## Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [ x] My code follows the code style of this project.

Include additional exception types to prevent search-thread death on a transaction lock issue.

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -355,7 +355,7 @@ def bulk_upsert(cls, data):
         log.debug("Inserting items {} to {}".format(i, min(i+step, num_rows)))
         try:
             InsertQuery(cls, rows=data.values()[i:min(i+step, num_rows)]).upsert().execute()
-        except OperationalError as e:
+        except Exception as e:
             log.warning("%s... Retrying", e)
             continue
 


chuyskywalker:
Looks sane to me, don't know how to replicate it, but the logic itself is sound.

invisiblek:
This is kind of a shotgun approach. 
I would prefer we catch the specific exception. 

https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2616

invisiblek:
Haven't been able to catch this specific exception. Shotgun approach wins

PR #2350, opened 2016-07-27
Fix gym InfoWindow issues. (#2349)

tkSimon:
see Gym InfoWindow bug #2349

PR #2688, opened 2016-07-30
Only allow single info window at a time.

joshluongo:
## Description
I've made it so you can only open one info window at a time if you're on a compact device. All of the info windows are stored in a hash map within `map_data`.

I've also added a convince function `clearInfoWindows()` to close all open info windows.

This is an optimised version based on the suggestions by @brycedorn from my original PR #2124.

A compact device is defined as a viewport that is less than 768px.

## Motivation and Context
On a phone the screen get's very crowded with multiple info windows open. If the user keeps tapping the marker multiple times then it can be frustrating to close all of the windows.

## How Has This Been Tested?
This is a minor change but I've tested it on iOS and Safari on Mac OS X.

_Can't show the change in a screenshot._

## Types of changes

## Checklist:


chuyskywalker:
The gym near me continues to get stuck open even with this PR:

2016-07-29 22_03_51-pokemon go map

I did not click anything here, just hovered over each with my mouse, gym first.

chuyskywalker:
Might actually be related to #2350

joshluongo:
Thats what it seems like to me as well, Its something to do with how it gets presented. I'll take a look in to it in few hours.

tkSimon:
It gets stuck because the info window is re-created on every update, without closing it prior to recreating it. There's really no point in recreating them if you can just update an existing one, like in my PR mentioned above.

chuyskywalker:
@tkSimon's fix for the gym info has been merged in, which is probably what collided this PR.

sorokyl:
So... you can't have two pokemon info windows up at once? That doesn't sound like a very good change, esp for desktop use.

MRokas:
This should definitely be toogleable option

joshluongo:
I'll rebase and see if that fixes the issue. I'll add a toggle to disable this behaviour.

What if we have that behaviour on by default if their using a mobile device? This was the main motivation behind making this change.

joshluongo:
I've rebased this pull request. The issue where gyms windows wouldn't disappear has been fixed thanks to #2350. 

I'll update when I've got a toggle setting for this feature working.

joshluongo:
I've added a switch to disable the single window behaviour. Its only on default if you're on a compact device.

I've also update the main PR description with this information.

AHAAAAAAA:
I'd rather not add a toggle for this as this is typically not a problem for larger screens. Make it the default for mobile screen resolutions (smaller than a table), and remove the toggle.

joshluongo:
I've also rebased the PR with the develop branch again.

AHAAAAAAA:
lgtm, +1

Issue #2527, opened 2016-07-29
First search pass finds no pokestops/gyms, doesn't keep using same login creds


chuyskywalker:
When running, the first search loop is not finding any pokestops nor gyms. When the second loop kicks off, it does not use the same login, but grabs a new session. This loop finds all the things. The 3rd loop will, correctly, keep using the session from the 2nd loop.

```
# python ./runserver.py \
  --host 0.0.0.0 \
  -a xxx -u xxxx -p xxxx \
  -l "Central Park" \
  -st 3 -P 80 \
  -k 'mapkeyhere' \
  -t 1 -ld 15 -td 3 -sd 3
2016-07-29 02:38:16,854 [     runserver] [   INFO] Parsed location is: xxxxx/xxxxx/0.0000 (lat/lng/alt)
2016-07-29 02:38:16,856 [        search] [   INFO] Search loop 0 starting
2016-07-29 02:38:16,857 [        search] [   INFO] Attempting login to Pokemon Go.
2016-07-29 02:38:16,858 [      auth_ptc] [   INFO] PTC login for: xxxxxxx
2016-07-29 02:38:17,713 [      auth_ptc] [   INFO] PTC Login successful
2016-07-29 02:38:18,079 [        search] [   INFO] Login to Pokemon Go successful.
2016-07-29 02:38:18,498 [        models] [   INFO] Upserted 1 pokemon, 0 pokestops, and 0 gyms
2016-07-29 02:38:21,767 [        models] [   INFO] Upserted 0 pokemon, 0 pokestops, and 0 gyms
2016-07-29 02:38:25,058 [        models] [   INFO] Upserted 3 pokemon, 0 pokestops, and 0 gyms
2016-07-29 02:38:28,371 [        models] [   INFO] Upserted 2 pokemon, 0 pokestops, and 0 gyms
2016-07-29 02:38:31,680 [        models] [   INFO] Upserted 3 pokemon, 0 pokestops, and 0 gyms
2016-07-29 02:38:34,974 [        models] [   INFO] Upserted 1 pokemon, 0 pokestops, and 0 gyms
2016-07-29 02:38:38,286 [        models] [   INFO] Upserted 2 pokemon, 0 pokestops, and 0 gyms
2016-07-29 02:38:41,553 [        models] [   INFO] Upserted 0 pokemon, 0 pokestops, and 0 gyms
[...]

PR #2691, opened 2016-07-30
Parse pokestops/gyms on first search loop

chuyskywalker:
Pretty much what the title says, in response to #2527.

With the rate limiting in place, it seems wasteful to wait an entire loop to parse these, and (as far as I can tell) doesn't seem to dramatically impede speed.


AHAAAAAAA:
This was more necessary when our search was much slower, but now it's useless I agree!

+1, merging

PR #2702, opened 2016-07-30
Update runserver.py

thimslugga:
Added env after /usr/bin/ to allow for starting app via ./runserver.py when using virtualenv.


Chlodochar:
Yes, some distros require running python through env. It's a better approach than directly referring to `/usr/bin/python` (essentially hardcoding where python must be located), so :+1:.

chuyskywalker:
Yup, good practice confirmed.

PR #1795, opened 2016-07-24
Adding start/stop functionality from the UI

JamesWP:
## Description

adds the start stop functionality requested/mentioned  in issue #1586

adds the -nsc --no-search-control option to disable this feature (on by default)

## Motivation and Context

allows user to start stop the service without accessing the server

addresses issue: #1586

## How Has This Been Tested?
click the stop button, service stops cleanly
click the start button, starts again
clicking either button when already started / stopped does nothing

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

added new endpoint in app.py to accept status update requests
added methods to control the search thread and to support restarting
once stopped
added option to sidebar for user controls

Issue #1586, opened 2016-07-23
[Question] Stop current search progress via button


youlllovegithubde:
Hello Trainers,
is it possible to stop a current search progress via button on client side?

JamesWP:
i have proposed a change that will allow this functionality

PR #2722, opened 2016-07-30
Adding start/stop functionality from the UI (Take two)

JamesWP:
## Description

adds the start stop functionality requested/mentioned  in issue #1586

adds the -nsc --no-search-control option to disable this feature (on by default)

(this is the second pr for this feature, because github) previous discussion here #1795

## Motivation and Context

allows user to start stop the service without accessing the server

addresses issue: #1586

## How Has This Been Tested?
click the stop button, service stops cleanly
click the start button, starts again
clicking either button when already started / stopped does nothing

## Types of changes

## Checklist:

added new endpoint in app.py to accept status update requests
added methods to control the search thread and to support restarting
once stopped
added option to sidebar for user controls

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -71,6 +71,10 @@ def get_args():
     parser.add_argument('-ns', '--no-server',
                         help='No-Server Mode. Starts the searcher but not the Webserver.',
                         action='store_true', default=False)
+    parser.add_argument('-nsc','--no-search-control',
+                        help='Disables search control',
+                        action='store_false', dest='search_control')
+    parser.set_defaults(search_control=True)
     parser.add_argument('-os', '--only-server',
                         help='Server-Only Mode. Starts only the Webserver without the searcher.',
                         action='store_true', default=False)
--- a/runserver.py
+++ b/runserver.py
@@ -25,6 +25,7 @@ from pogom.pgoapi.utilities import get_pos_by_name
 
 def main():
     global args
+    global control
     args = get_args()
 
     if args.debug:
--- a/runserver.py
+++ b/runserver.py
@@ -101,15 +101,18 @@ def main():
     config['REQ_SLEEP'] = args.scan_delay
 
     if args.no_server:
-        # This loop allows for ctrl-c interupts to work since flask won't be holding the program open
-        while True:
-            try:
-                time.sleep(60)
-            except KeyboardInterrupt:
-                control.stop()
-                break
+        wait_for_exit()
     else:
         app.run(threaded=True, use_reloader=False, debug=args.debug, host=args.host, port=args.port)
+
+# This loop allows for ctrl-c interupts to work since flask won't be holding the program open
+def wait_for_exit():
+    while True:
+        try:
+            time.sleep(60)
+        except KeyboardInterrupt:
+            control.stop()
+            break
 
 #method to start (or restart the search thread)
 def start_search_thread():
--- a/runserver.py
+++ b/runserver.py
@@ -105,7 +105,7 @@ def main():
         while True:
             try:
                 time.sleep(60)
-            except:
+            except KeyboardInterrupt:
                 control.stop()
                 break
     else:
@@ -113,7 +113,6 @@ def main():
 
 #method to start (or restart the search thread)
 def start_search_thread():
-    global args
     global search_thread
     search_loop_start()
     if not args.mock:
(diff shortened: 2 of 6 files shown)


ottowayne:
I would make this opt-in because the user frontend shouldn't have this much impact on the server by default.

JamesWP:
well, its enabled by default now so whoever is going to merge this into develop can choose, my thoughts are that it should be enabled by defaut

JamesWP:
paging @mackhankins, @chuyskywalker, @joshslark for comments etc...

mackhankins:
@JamesWP I absolutely love the feature and the toggle is much better.  It behaves a tad bit different for me on this push.  When I toggle it off it stops immediately instead of finishing the current loop.  Sorry, not sure what to call it.

I'm biased on the default thing cause I always need this visible.

pogom/utils.py
```diff
@@ -71,6 +71,10 @@ def get_args():
     parser.add_argument('-ns', '--no-server',
                         help='No-Server Mode. Starts the searcher but not the Webserver.',
                         action='store_true', default=False)
+    parser.add_argument('-nsc','--no-search-control',
+                        help='Disables search control',
+                        action='store_false', dest='search_control')
+    parser.set_defaults(search_control=True)
```
AHAAAAAAA: why not just `default=True` in the add_argument function?

runserver.py
```diff
@@ -16,14 +16,16 @@
...
+from pogom.search import search_loop_stop, search_loop_start
+from pogom.models import init_database, create_tables, drop_tables, Pokemon, Pokestop
 
 from pogom.pgoapi.utilities import get_pos_by_name
 
 
-if __name__ == '__main__':
+def main():
+    global args
+    global control
```
AHAAAAAAA: eww. globals. Why do args have to be global? There must be a cleaner way to refactor this.

AHAAAAAAA:
Tested, works well and I agree it should be available by default. Have you updated the usage documentation? Need more eyes on the code itself, feel like the implementation might introduce a few issues.

runserver.py
JamesWP: Start_search_thread needs args too :/ simple solution was make args global

The Def main with the call at the bottom was just so I could define other functions below

JamesWP:
@AHAAAAAAA I need to look up what I need to change in the documentation as I've not even read it.
Also final decision if this should be an opt in feature or opt out (as it is now)

AHAAAAAAA:
It should be opt out. This project targets the personal user and not people running servers. People running servers can opt out, but as long as this is primarily for personal use then it will be opt out.

JamesWP:
OK, it's ready for merge minus the documentation, I'm not available till after tomorrow (UK) so I can give you commit access to finish that off so there need not be any more rebasing.

AHAAAAAAA:
Sounds good, yeah I'll wait till another contributor looks it over and then we'll merge.

chuyskywalker:
I'll give this a go.

chuyskywalker:
When you move the marker while the searcher is stopped, then start it again, it doesn't relocate the search thread.

runserver.py
```diff
@@ -85,18 +87,12 @@
...
-        else:
-            log.debug('Starting a fake search thread')
-            insert_mock_data()
-            search_thread = Thread(target=fake_search_loop)
+        start_search_thread()
 
-        search_thread.daemon = True
-        search_thread.name = 'search_thread'
-        search_thread.start()
+    app = Pogom(__name__)
```
chuyskywalker: If you look at the full file, `app = Pogom(__name__)` is being initialized twice.

chuyskywalker:
When the `stopping` is set, if the queue isn't empty yet, it causes the `search_thread`(s) to go into an infinite loop and doesn't clear the (global, I know) queue out. 

This code isn't acting so much like a "pause" button, but it's trying to straddle the gap between a pause button (like a global "pause_thread" boolean that causes an infinite loop in the `search_thread`s) but also is trying to stop, halt, and recreate the `search_loop` thread itself. The two are a bit at odds and not fully resolved with the way this is coded right now.

JamesWP:
Might just need to stop the addition of items into the queue instead then? I wrote this before the change to how searching worked. 
The intended effect is to stop each thread after it completes it's current action.
Again I'm not in a position to work on this for a few days so if someone knows a better way and wants commit access to my repo, please ask


Sent from my Samsung device

-------- Original message --------

chuyskywalker:
I took this work and branched off it (via copy paste, really) to address most of the concerns: #2756

JamesWP:
Fine by me :) though next time look into using somthing like git cherry-pick to maintain history etc...



Sent from my Samsung device

-------- Original message --------

chuyskywalker:
#2756

PR #2725, opened 2016-07-30
add option to select all in "notify of pokemon"

twistednipple:
Select all feature to improve ui. Right now selecting numerous pokemons form drop down list is pain, it will be much easier to click to remove on unwanted rather than manually selecting all.


marauder37:
@twistednipple thanks for the PR, I would like this feature. But the PR is set up wrong. You need to compare our repo:develop to your repo:your branch. Thanks!

PR #2728, opened 2016-07-30
add webhook notification for location changes

blaulan:
send a webhook notification every time the location is changed, useful when receiver needs origin coordinates to calculate distance.


andrewda:
Will test it in a bit, but looks good!

PR #2749, opened 2016-07-30
Refer to Store and external properties as strings

Chlodochar:
## Description
To eventually comply with linting rules an automated formatter needs to handle JS. It needs to camelCase some of the properties, except the ones that are referred by as strings. These include the properties from JSON and the ones from localStorage. This PR will change all of these references to use strings instead of normal properties.

`hidden` and `marker` are not referred to as string, because they do not originate from the server (it isn't part of JSON).

## How Has This Been Tested?
Going through the different features of the GUI and seeing whether they still work.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -108,59 +108,59 @@ var StoreTypes = {
 };
 
 var StoreOptions = {
-  map_style: {
+  'map_style': {
     default: 'roadmap',
     type: StoreTypes.String
   },
-  remember_select_exclude: {
+  'remember_select_exclude': {
     default: [],
     type: StoreTypes.JSON
   },
-  remember_select_notify: {
+  'remember_select_notify': {
     default: [],
     type: StoreTypes.JSON
   },
-  showGyms: {
+  'showGyms': {
     default: false,
     type: StoreTypes.Boolean
   },
-  showPokemon: {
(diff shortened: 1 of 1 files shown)


static/js/map.js
```diff
@@ -108,59 +108,59 @@ var StoreTypes = {
...
-  geoLocate: {
+  'geoLocate': {
     default: false,
     type: StoreTypes.Boolean
   },
-  playSound: {
+  'playSound': {
     default: false,
     type: StoreTypes.Boolean
   },
```
andrewda: why is `playSound` defined twice in this object? (here and on line 147)

Chlodochar: It is incorrect. It already was incorrect, but because there was no linting it was not caught by anyone. #2058 did catch that error. I can create a separate PR for this problem, but I'd like to do that after this one.

Chlodochar: Never mind, I removed it in this PR. Kind of cheating, but ya.

Chlodochar:
Never mind this PR for now. I'll just apply JS formatting without renaming variables. Renaming variables wasn't working consistently, so we'll have to do that partly manually anyways. That can be done after automatic formatting.

PR #2756, opened 2016-07-30
Allow search to be toggled on/off from the UI

chuyskywalker:
See: #1586, #1795, #2722

Cleaned up, no globals, doesn't "restart" threads, just pauses the overall process; handles relocated marker during "paused" state.

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -28,19 +28,39 @@ class Pogom(Flask):
         self.route("/loc", methods=['GET'])(self.loc)
         self.route("/next_loc", methods=['POST'])(self.next_loc)
         self.route("/mobile", methods=['GET'])(self.list_pokemon)
+        self.route("/search_control", methods=['GET'])(self.get_search_control)
+        self.route("/search_control", methods=['POST'])(self.post_search_control)
+
+    def set_search_control(self, control):
+        self.search_control = control
+
+    def get_search_control(self):
+        return jsonify({'status': self.search_control.is_set()})
+
+    def post_search_control(self):
+        action = request.args.get('action','none')
+        if action == 'on':
+            self.search_control.set()
+            log.info('Search thread turned resumed')
+        elif action == 'off':
+            self.search_control.clear()
+            log.info('Search thread turned paused')
+        else:
+            return jsonify({'message':'invalid use of api'})
+        return self.get_search_control()
 
     def fullmap(self):
         args = get_args()
-        display = "inline"
-        if args.fixed_location:
-            display = "none"
+        fixed_display = "none" if args.fixed_location else "inline"
+        search_display = "inline" if args.search_control else "none"
 
         return render_template('map.html',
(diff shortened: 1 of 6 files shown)


Chlodochar:
Tested and working. This is a very welcome addition. Code looks good, except that Landscape.io found one unused parameter, could you remove that one?

mackhankins:
Very clean.  I tested the following on my mac.  I'm going to push it out to Heroku for some "field testing"

* Follow Location with Search on & off
* Drag Marker with Search on & off

@chuyskywalker Holy branches 👍

pogom/app.py
```diff
@@ -28,19 +28,39 @@ def __init__(self, import_name, **kwargs):
...
+        self.route("/search_control", methods=['POST'])(self.post_search_control)
+
+    def set_search_control(self, control):
+        self.search_control = control
+
+    def get_search_control(self):
+        return jsonify({'status': self.search_control.is_set()})
+
+    def post_search_control(self):
+        action = request.args.get('action','none')
```
chuyskywalker: I just realized I've missed something here wherein this endpoint is still available even when the flag `-nsc` is turned on. Need to detect that flag and disable this endpoint.

chuyskywalker: Fixed

magimat:
been testing this feature tonight, works great

medghaim:
+1

mackhankins:
Thanks :)

magimat:
hmm.. I'm running the develop branch including this PR since this morning and it seems each time I resume the search after pausing for a long while (maybe an hour), it doesn't resume correctly. I'm getting the "Niantic server under heavy load" message after resuming and the only way to make it search again is to kill the server and start it back up manually.

chuyskywalker:
Ah dang, I think I see the issue here. Since this just "pauses" all the threads, the login token can will go invalid over time. And when we "unpause" it doesn't validate or re-login, so it gets stuck trying to use the same broken token in an infinite loop. (This is probably even happening _regardless_ of this feature, like if the token expiration rolls around in the middle of a search loop.)

I'll look into a fix for this tonight -- it'll probably require a lot of logical re-writing though :|

magimat:
thanks!  I'd help, but I don't really have the skills nor time to really be helpful other than reporting some test feedback.

thanks for your work.. it's really appreciated!

invisiblek:
I notice search_control gets called, even when I'm not dicking around with the toggle. Might be worth looking in to, seems to be unnecessary calls to this going on.

JamesWP:
If this is the same as my initial version it is called to keep the toggle In sync with the backend.
I.e. if you toggle on one page the other is also updated


Sent from my Samsung device

-------- Original message --------

invisiblek:
Hmm, yea I didn't think of that. I suppose that makes sense. Seems a bit costly to me though...

chuyskywalker:
You're not wrong, but we've got a polling architecture, so it's what we have to work with.

Websockets now... ;D

PR #2770, opened 2016-07-30
Fix absolute paths in stats.js.

tkSimon:
When running this behind a reverse proxy and not in the document root, img tags are broken because they are referencing files from the document root instead of using relative paths.

## Description
Removed the leading slash in src attributes of img tags in `static/js/stats.js`.

## Motivation and Context
Motivation? I'd like it to work on my server, obviously :P
Also most of the ppl running this behind a rev proxy will benefit from this.

## How Has This Been Tested?
Tested on my root server with rev proxy.

## Screenshots (if appropriate):
Before: http://imgur.com/a/QuFUs

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [X] My code follows the code style of this project.

Issue #2712, opened 2016-07-30
Travis crashes on npm install


andrewda:
Known issue, probably should be looked at.

Example: https://travis-ci.org/AHAAAAAAA/PokemonGo-Map/builds/148433058

efiku:
Maybe this will help:
https://docs.travis-ci.com/user/trusty-ci-environment/

tkSimon:
someone filed this way back with travis-ci (travis-ci/travis-ci#3130) and they referred them to upstream (npm/npm#6309) which is outdated also... but i found this on SO: http://stackoverflow.com/questions/26474640/why-travisci-builds-fails-during-npm-install-due-to-file-lock

Can anyone reproduce this with their local setup?

Chlodochar:
I have not been able to reproduce this yet, but it might be possible to fix the issue by installing a new version of node using `nvm install 4.4.7` in Travis. It should also result in a newer npm version.

PR #2772, opened 2016-07-31
Fix absolute paths in mobile.js

tkSimon:
Similar to #2770, but for /mobile in `static/js/mobile.js`.
Would have submitted as one PR if I had noticed earlier.

When running this behind a reverse proxy and not in the document root, the refresh button is broken because javascript is setting `document.location.href` to `/mobile` instead of just `mobile`.

## Description
Removed the leading slash of href in `static/js/mobile.js`.

## Motivation and Context
Motivation? I'd like it to work on my server, obviously :P
Also most of the ppl running this behind a rev proxy will benefit from this.

## How Has This Been Tested?
Tested on my root server with rev proxy.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/mobile.js
+++ b/static/js/mobile.js
@@ -9,7 +9,7 @@ var navBtn = document.querySelector('#nav button');
 navBtn.onclick = function() {
   if (localStorage.useLoc !== 'true') {
     navBtn.disabled = true;
-    return (location.href="/mobile");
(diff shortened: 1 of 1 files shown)


tkSimon:
Check failed on npm install, see #2712

chuyskywalker:
Lgtm

Issue #2810, opened 2016-07-31
Hex Builder location broken 


Quiv:
Hello, i'm trying to set multi thread but it seems i have a problem with hex builder ...
Using the up to date dev branch.
All center of each scan are not at the correct place.
I test all my location with google map and http://www.hamstermap.com/quickmap.php and nothing showing at the correct place.

hex generator : 

python location_generator.py -lat 18.024281 -lon -63.047378 -st 5 -lp 2 --account accounts.csv --auth google

What it is suppose to look like with coords : 

![2](https://cloud.githubusercontent.com/assets/6039307/17275199/cbe6da86-56d5-11e6-858a-dc9dc9f4925a.jpg)

What it does : 

![1](https://cloud.githubusercontent.com/assets/6039307/17275201/d0c83ac2-56d5-11e6-88fb-aeb3ca28aa28.jpg)

## Steps to Reproduce (for bugs)
1. clean install w/ dev branch
2. locationgenerator.py

## Your Environment

* Version used:
Dev Branch 
Ubuntu server 16.04

Please can anyone help me to understand.
Thanks in advance

fourmansyah:
can some one help me with this problem ?
<div class="separator" style="clear: both; text-align: center;">
<a href="https://2.bp.blogspot.com/-eS6n-0OEL0c/V52odIf46ZI/AAAAAAAAKx4/UvuMdrVigVsSIlUnsWuRgXx43vJvFzkUgCLcB/s1600/1.jpg" imageanchor="1" style="margin-left: 1em; margin-right: 1em;"><img border="0" height="316" src="https://2.bp.blogspot.com/-eS6n-0OEL0c/V52odIf46ZI/AAAAAAAAKx4/UvuMdrVigVsSIlUnsWuRgXx43vJvFzkUgCLcB/s640/1.jpg" width="640" /></a></div>

Quiv:
your answer has nothing to do in this issue .... thanks

NoMotion:
I think this is a duplicate issue people have been posting. Take a search on the dev branch for this issue

Issue #2796, opened 2016-07-31
Gaps/holes in beehive


elenaran:
Running on py2.7 on Win10 - getting gaps when I try to run a beehive. See image

![screenshot](https://i.snag.gy/lzGnci.jpg)

I'm thinking maybe it's a rounding error or something in the "mod" calculation

Epion003:
yea i get this also. Wondering of possible to add a marker so we can just move that worker or have to look for the lat,lon. I cant find the post that had the numbers for the cells if u have or anyone has incase i have to edit the workers individually thank you

moggd:
Any luck finding the fix for this? Check out my image below.

![real](https://cloud.githubusercontent.com/assets/20671849/17282353/e428b888-5772-11e6-951a-289e1e146776.png)

Issue #2693, opened 2016-07-30
Beehive's location generator isn't what's breaking beehive, with proof


hgentry:
## Expected Behavior
Beehive should make a bunch of hex cells

## Current Behavior
Beehive makes utter nonsense.

I've been working on trying to fix this issue for a few hours and everyone seems to think it's the locations, but it's not. Let's plot all the points given to us by running with --verbose:

`python location_generator.py -st 2 -lp 2 -lat 38 -lon -86 --verbose
38.0, -86.0,
38.0040541813, -86.0,
38.0020270073, -85.9955628915,
37.9979728246, -85.9955631358,
37.9959458159, -86.0,
37.9979728246, -86.0044368642,
38.0020270073, -86.0044371085
`

I've been using the [cesiumjs demo](http://cesiumjs.org/Cesium/Apps/Sandcastle/index.html?src=Hello%20World.html&label=Showcases&gist=b847e8040846b9819ebf7e2329c495e4) to plot points, using this [gist](https://gist.github.com/hgentry/b847e8040846b9819ebf7e2329c495e4). Let's see how these points look there:

![map](http://i.imgur.com/hZ5zFv1.png)

Looks good, right?

Let's run `./beehive.sh`:

![map](http://i.imgur.com/Bdx7kyT.png)

These worker locations are complete nonsense. I've tested this at much higher levels and they plot just fine, too. I don't know what's actually wrong, but we can look elsewhere.

## Steps to Reproduce (for bugs)

Detailed above

## Your Environment

* Version used: latest dev
* Environment name and version (e.g. Python 2.7): Python 2.7
* Operating System and version (desktop or mobile): raspbian

marauder37:
https://i.snag.gy/tPOrga.jpg

see how the points are on a gentle lean vertically? that's not going to tesselate correctly, is it?

hgentry:
Forgot to mention there's an issue with the `mod` variable. That should be set to 0 until someone can work out a version that works spherically.

marauder37:
Also just use http://www.hamstermap.com/quickmap.php for plotting points

hgentry:
I use cesium so I don't have to deal with projection-based issues near the poles

Dagothur:
Is there any website that will allow you to lift coordinates out of the plotted map?

I like to trim the sides of my beehive as they often cover forests, lowly populated regions, etc. and it's proved cumbersome and almost a guess and check exercise to figure out which ones I should be dropping.

Issue #2621, opened 2016-07-29
Hex Generator Not Evenly Generating?


DeviantEng:
Latest dev build on Linux.  

Used this to generate my beehive.sh:
`python location_generator.py -lat "37.674221" -lon "-82.777588" -st 3 -lp 5 --accounts=/opt/PokemonGo-Map/accounts.csv`

Here is what I'm getting:
http://imgur.com/a/avNdT

I'm using 10 PTC accounts to distribute the load.  Using `-st 5 -lp 3` is still not showing up as expected.  Did I do something wrong?

hgentry:
Are you sure you're fully updated? A fix for this went out three hours ago.

DeviantEng:
Yes, I am running commit `0eb3c9bc0afcefc829d8d50ce3ab1200bfd9ce64`, which is currently the latest, with PR #2597.

DeviantEng:
Updated to the latest develop build, `681580a`, and still seeing weird behavior.

`-st 3 -lp 3` is giving me this:
http://imgur.com/a/RqHR2

hgentry:
I'm having issues, too. I think I've figured out the cause and will be updating it soon

ChopSu3y:
Has a method for adding multiple accounts to the bash script been found? If so, how is it done?

PR #2818, opened 2016-07-31
Fix broken regex when parsing coordinates with no comma

Scyntrus:
There is a typo in the regex that detects the location argument as coordinates.
Also fixes misaligned hex beehive generator grid.

## Description

The question mark makes the preceding expression optional. It is clear it is meant for the comma, not the latitude value.
Since it is not detected as coordinate values, the argument is passed to the GoogleV3 geolocator api, which then returns the value but with a lower precision. This is why the beehive seems misaligned.

## Motivation and Context

When using the hex beehive generator, the generated grid is misaligned and there are gaps between the hexes. 
Fixes:
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2810
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2796
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2693
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2552
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2621

## How Has This Been Tested?

Ran using the coordinates produced by the hex beehive generator and there are no more gaps.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/pgoapi/utilities.py
+++ b/pogom/pgoapi/utilities.py
@@ -50,7 +50,7 @@ def to_camel_case(value):
     return "".join(c.next()(x) if x else '_' for x in value.split("_"))
 
 def get_pos_by_name(location_name):
-    prog = re.compile("^(\-?\d+\.\d+)?,\s*(\-?\d+\.\d+?)$")
+    prog = re.compile("^(\-?\d+\.\d+)[,\s]\s*(\-?\d+\.\d+?)$")
     res = prog.match(location_name)
     latitude, longitude, altitude = None, None, None
     if res:


Scyntrus:
CI build server is broken. Not due to my change..

Scyntrus:
Rerunning the build was sucessful.

Chlodochar:
You probably want `(,|\s)` instead of `,?`. That way you're sure a single number is not being interpreted as 2 numbers.

Scyntrus:
Somewhat of a technicality but Python regex is greedy, so the first expression matching the first number will match the entire thing, leaving nothing for the last expression to match. This will result in the regex properly failing. Also note that the way the regex is written means that both values must have a decimal point.

It is true your suggestion is a bit more readable.

Chlodochar:
I think it will backtrack one character and still parse the second number with one digit before the `,`.

Scyntrus:
```
>>> prog = re.compile("^(\-?\d+\.\d+)?,\s*(\-?\d+\.\d+?)$")
>>> res = prog.match("0.00.0")
>>> print res
None
>>> res = prog.match("0.0,0.0")
>>> print res
<_sre.SRE_Match object at 0x0000000002370140>
>>>
```
Not according to my Python shell

Scyntrus:
Fine, I guess it will handle the case where someone forgets to separate their latitude and longitude.

Chlodochar:
This looks good. :+1:

hkme:
Hey you have to change line 57 as well, as you added one more capture group now the longitude is on the third (3) capture group instead of the second group (2)

Scyntrus:
Fixed by changing it to use square brackets.

chuyskywalker:
Yay, now I'll be able to copy Lat/Lon directly from google maps 😃

PR #2863, opened 2016-07-31
Python Code Quality

chuyskywalker:
Tons of little things; see the commit list since I tried to make each one self-explanitory

## Motivation and Context
Clean up the codebase to be more consistent and apply best practices

## How Has This Been Tested?
Ran it.

## Types of changes


Ashex:
Why did you switch all the string formatters to the old style? It's good that it's now consistent but it seems an odd choice to use the old instead of new.


https://pyformat.info/#simple

chuyskywalker:
It's not actually the old style, that would be:

```
log.info("string %s" % variable)
```

This is
```
log.info("string %s", variable)
```

It's a very subtle difference, but the theory is that `variable` resolution is deferred in case that log level isn't actually rendered. That way you're don't doing a whole bunch of `.format()` calls when you're not actually logging the data.

Ashex:
I see now, I completely missed that these changed were only to logging. 

https://docs.python.org/3/howto/logging.html#logging-variable-data

AHAAAAAAA:
rebase. +1 for merge

pogom/models.py
```diff
@@ -279,60 +279,57 @@ def parse_map(map_dict, iteration_num, step, step_location):
...
         bulk_upsert(Pokemon, pokemons)
 
     if pokestops and config['parse_pokestops']:
         pokestops_upserted = len(pokestops)
-        log.debug("Upserting {} pokestops".format(len(pokestops)))
         bulk_upsert(Pokestop, pokestops)
 
     if gyms and config['parse_gyms']:
         gyms_upserted = len(gyms)
-        log.debug("Upserting {} gyms".format(len(gyms)))
```
Chlodochar: Why remove these? It seems confusing when Pokemon's are logged.

chuyskywalker:
Kinda wish landscape.io would fucking work for this branch...

Chlodochar:
OT, but we might want to switch to SideCI.

chuyskywalker:
As long as you can turn off it's "here, let me leave you 5000 github notifications" feature ;D

Chlodochar:
If the current problems are solved it seems alright to have a comment for each new problem.

PR #2888, opened 2016-07-31
Clean scanned locations

ingwarsw:
## Description
Clean scanned location table out od old data,,
Its important when you have low space on db.

## How Has This Been Tested?
Working on my server from few days..

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -341,6 +341,16 @@ def parse_map(map_dict, iteration_num, step, step_location):
 
     bulk_upsert(ScannedLocation, scanned)
 
+    clean_database()
+
+def clean_database():
+    flaskDb.connect_db()
+    query = (ScannedLocation
+            .delete()
+            .where((ScannedLocation.last_modified <
+                (datetime.utcnow() - timedelta(minutes=30)))))
+    query.execute()
+    flaskDb.close_db(None)
 
 
 def bulk_upsert(cls, data):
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
Good idea, I was going over the DB structure the other day and this solution makes sense!

warrussell:
Im not too familiar with python, but if this feature could be switched on/off in the config, that would be really cool. Im working on something and I dont want the database to be periodically cleared.

sfxer0:
Is there a way to add a button on the map UI that could "wipe" the scanned area's with this?    Or maybe when you change the scan location manually it could clear all the scanned area rings that "new" scan rings might touch.   For example if you are scanning your home and then want to scan 5 steps away overlapping but your new location doesn't "match up" with the other rings, instead of getting rings that overlap, the older rings could be erased if they interfere with new scanned ones?

Chlodochar:
Are the changes to `@classmethod` actually needed? What is the impact on other code? Without those I'd say :+1:.

Ashex:
if the implicit class argument isn't used then it's good practice to decorate it as a `@staticmethod`. I think PEP8 bitches about it.

chuyskywalker:
I haven't tested this, but the idea is a :+1: from me -- there's little to no value for old scan location data to stick around in the database.

Malouff:
@chuyskywalker couldn't the old database data sticking around be used to look for trends or create heat maps of Pokémon spawning locations (Finding Hotspots in your local area)?

chuyskywalker:
This is only cleaning up the "scanned locations" table; the pokemon information stays around.

Issue #2277, opened 2016-07-27
Web UI - Can't add enough "Hide Pokemon" or "Notify of Pokemon"


BullHorn7:
When you add a lot of Pokemon into these lists, they become as long as the entire browser window, eventually not allowing you to add more.

## Expected Behavior
Each of these specific 'windows' should have its own scrollbar, rather than sliding the entire UI up and down

## Current Behavior
The 'windows' elongate until no more Pokemon can be added.

## Steps to Reproduce (for bugs)
http://i.imgur.com/qwrHoCh.png

## Context
Trying to create a 'white-list' 

## Your Environment
Today's dev build.

tuommii:
Currently i use developer tools for that.
![screenshot from 2016-07-27 10-32-52](https://cloud.githubusercontent.com/assets/8975117/17167418/c321bd78-53e5-11e6-9d84-1d7bf2e0fd01.png)

jabberjabber:
Experience also sucks on mobile.

PR #2912, opened 2016-07-31
Fixed display issue with "Notify of Pokemon" select list

grjkz:
## Description
Fixed the "Notify of Pokemon" list that causes the selector menu to go off the screen when too many Pokemon are added into the list.
The changes also affect the css of "Hide Pokemon" as well with no visual or negative changes from what I've seen.

Firefox has a separate issue where the Sound slider/button vanishes off to the bottom of the viewport but that is a preexisting issue.

## Motivation and Context
Added because it becomes annoying to add a lot of pokemon.
Open Issue:
https://github.com/AHAAAAAAA/PokemonGo-Map/issues/2277

## How Has This Been Tested?
Tested on Chrome since that's where I saw the issue.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1272,6 +1272,10 @@ $(function() {
     // recall saved lists
     $selectExclude.val(Store.get('remember_select_exclude')).trigger("change");
     $selectNotify.val(Store.get('remember_select_notify')).trigger("change");
+
+    // set the height and overflow to prevent selector from going off the page
+    $select2RenderedLists = $('.select2-selection__rendered');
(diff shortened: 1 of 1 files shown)


invisiblek:
Works, but does kick an error for me when I first load the page: `Uncaught ReferenceError: $select2RenderedLists is not defined`

Presumably since these haven't been rendered yet. What about something like [this](https://gist.github.com/invisiblek/50ff9e4ff5ae12fa2bafb70ee719faad)? They really only need to be updated when one of these get changed.

Issue #2731, opened 2016-07-30
[Feature Request] "select all" in "notify on pokemon" menu


twistednipple:
Right now selecting multiple pokemons to get notifications about is pain, selecting all with a button and then manually deselecting unnecessary is much easier way.
plus potentially feature of inverting selection between hide pokemon/notify of pokemon

Issue #1429, opened 2016-07-23
The current process for filtering is laborious.


Pixeltramp:
Any chance you could restore the comma-delineated flag, or allow a quicker input method like a frontend-based comma-delineated input?

2-m:
Agreed

talentedmrripley:
I would like to see this return also.

FaustKnight:
Agreed, comma-delineated would be amazing. Best if it can also be put in a config file and not just the map.

PR #2942, opened 2016-08-01
Add a way to add all non-excluded Pokemon to the notify list.

JackDoan:
## Description

- Added two buttons in the sidebar, Add All and Remove All
- Add All automatically adds all non-excluded Pokemon to the notify list
- This can sometimes be a lot of Pokemon, so Remove All clears the notify list
- 'notifyAboutPokemon(id)' was updated to silently ignore requests to notify about excluded Pokemon

## Motivation and Context

In suburban areas like my own, anything that isn't a Pidgey or Rattata or the like is notable. These buttons make it quick and easy to set up notifications for when something interesting spawns.

Issues: 
#2731 
#1961
#1609
#1555 
#1415

Also kinda helps with:
#1429
#1228

## How Has This Been Tested?

This works correctly in Chrome 50 on Arch Linux. I haven't checked anywhere else yet.
The server was running in Docker, but all changes are client side.

image

## Types of changes

## Checklist:


JackDoan:
I apologize for my messy commits. If there's a way to clean them up, I'd love to know.

invisiblek:
I thought about implementing this awhile back, my thought was to have a switch next to the "Hide Pokemon" label that would enable/disable notifications on all. If you enabled that switch, it'd grey out and disable the box for adding/removing these and just notify on all pokemon (of course, not notifying on ignored ones). Seems it might be better UX since if you added a new one to the ignore list, you wouldn't have to search for it and remove it from the notify list.

JackDoan:
@invisiblek The only issue I see there is what if you want to display a Pokemon, but not notify about it? I think Eevee is a pretty good example of this.

What if instead adding something to exclude also removed it from notify automatically?

AHAAAAAAA:
This has been brought up a few times, but implementation always seemed confusing. We essentially want a Notify selector that's 'All except hidden'. It solves 95% of the issue.

Issue #2941, opened 2016-08-01
Notifications FIX


cicklow:
Notifications appear infinitely, even when a pokémon has disappeared from the map!

## Possible Solution

in \static\dist\js\map.js
search:
     ` sendNotification('A wild ' + item.pokemon_name + ' appeared!', 'Click to load map', 'static/icons/' + item.pokemon_id + '.png', item.latitude, item.longitude);`
and change by
      `sendNotification('A wild ' + item.pokemon_name + ' appeared!', 'Click to load map', 'static/icons/' + item.pokemon_id + '.png', item.latitude, item.longitude, item.disappear_time);`

search:
`function sendNotification(title, text, icon, lat, lng) {
  if (!("Notification" in window)) {
    return false; // Notifications are not present in browser
  }`
change by
`function sendNotification(title, text, icon, lat, lng, ttt) {
  if (!("Notification" in window)) {
    return false; // Notifications are not present in browser
  }
	var now = new Date();
	if (ttt < now) return false;`

JackDoan:
I have not experienced this issue. Would you happen to have a screenshot of a notification appearing for a Pokemon that is gone?

cicklow:
![sin titulo](https://cloud.githubusercontent.com/assets/933689/17283546/0eadae3c-5788-11e6-82c8-2248466d8f69.png)

Briones:
It happens to me also, when the pokemon disappers the page sends infinite notifications of that pokémon 😕

adcurtin:
Thanks for the fix!

Here's a better formatted code block for the second one (and I fixed the style :p) :

    function sendNotification(title, text, icon, lat, lng, ttt) {
      if (!("Notification" in window)) {
        return false; // Notifications are not present in browser
      }

      var now = new Date();
      if (ttt < now) {
        return false;
      }

datbomb:
May be similar to #2930 and/or #2131.

BaIthamel:
ahaha i had this when i had it hidden AND notify! so if you add the same pokemon to the hide and notify list you get incessant alerts

PR #2950, opened 2016-08-01
Do not send a notification if a Pokemon has expired

JackDoan:
## Description

Sometimes a notification can persist/exist even if the Pokemon in question has despawned. This helps prevent that.

## Motivation and Context

Addresses issue #2941 

## How Has This Been Tested?

Notifications for Rattata near me are still sent.
I'm not entirely sure what causes phantom notifications, but this should help prevent them.
## Types of changes

## Checklist:

     Caveat: 'sendNotification()' is now coupled with the Pokemon marker. It probably should be refactored into something like sendPokemonNotification, but I haven't seen a style guide for this project yet.  ¯\_(ツ)_/¯ 


PathToLife:
Does this solve the notification spamming issue?
Case: When pokemon expires, chrome spams the pokemon found notification over 30 times till page refreshed.
Thanks, =) I don't wanna delve into it if you've fixed it!

JackDoan:
@PathToLife It should! I haven't proven it yet, and that spamming may actually indicate a deeper underlying issue, but this should _probably_ fix it.

MRokas:
There's also spam whenever Lured Pokemon spawns.

Scyntrus:
I feel like this check should happen earlier in the code. The way you have implemented right now means that the audio.play() might still occur, and that's much more annoying than the notification.

JackDoan:
@Scyntrus Nice catch! I've moved the check to account for this.

invisiblek:
Would it make more sense to short circuit setupPokemonMarker entirely if its expired? I see no reason in doing anything in this method if the thing is expired.

JackDoan:
@invisiblek Good thinking! I've adjusted my change to account for this.

static/js/map.js
```diff
@@ -602,9 +602,9 @@ function setupPokemonMarker(item, skipNotification, isBounceDisabled) {
   if (notifiedPokemon.indexOf(item.pokemon_id) > -1) {
     if (!skipNotification) {
       if (Store.get('playSound')) {
-        audio.play();
```
invisiblek: hmm, the logic here seems off now. Its only going to notify if playSound is enabled

JackDoan: oops! I've fixed this.

static/js/map.js
```diff
@@ -602,7 +602,7 @@ function setupPokemonMarker(item, skipNotification, isBounceDisabled) {
...
+      audio.play();
```
andrewda: Why did you remove the indentation? Please re-add 2 spaces.

PR #2987, opened 2016-08-01
Implement lured pokemon webhook

hhu94:
## Description
Implemented a webhook for lured pokemon that is parallel to the normal pokemon webhook.
Also increased the timeout for the webhook post to allow more wiggle room for the listening server. This is a known fix that is documented in PokeAlarm's README. https://github.com/kvangent/PokeAlarm

## Motivation and Context
This allows third party apps like PokeAlarm to notify when a lured pokemon appears.

## How Has This Been Tested?
I implemented this while simultaneously working on implementing lured pokemon notifications for PokeAlarm, see this pull request https://github.com/kvangent/PokeAlarm/pull/16.
Lured pokemon notifications work fine under the normal use cases.

## Types of changes

## Checklist:


pogom/models.py
```diff
@@ -283,6 +283,16 @@ def parse_map(map_dict, iteration_num, step, step_location):
                         lure_expiration = datetime.utcfromtimestamp(
                             f['lure_info']['lure_expires_timestamp_ms'] / 1000.0)
                         active_pokemon_id = f['lure_info']['active_pokemon_id']
+
+                        webhook_data = {
+                            'pokestop_id': b64encode(str(f['id'])),
```
hkme: for type 'pokemon' i think you need a "encounter_id" and "spawnpoint_id", not 'pokestop_id'
e.g. (have not tested)
`
                    'encounter_id': b64encode(str(f['lure_info']['encounter_id'])),
                    'spawnpoint_id': f['id'],
`

hhu94: Lured pokemon don't have an encounter_id nor a spawnpoint_id. Though you are right in that the pokemon webhook should have a fixed structure.

I'll change it so that it sends a lured_pokemon webhook instead.

hkme: there is a encounter id as I check the pgoapi
https://github.com/tejado/pgoapi/blob/81e786cabf027a1c8fbd1e9a07e1c11aa3d8ee8b/pgoapi/protos/POGOProtos_src/Map/Fort/FortLureInfo.proto

viperfx:
@hhu94  I am curious, why was the encounterId b64encode'd ?

pogom/models.py
hhu94: Man this was an interesting one to figure out. Yes, there is an encounter_id in the link you sent, but the proto files in our repo are outdated and that field is called unknown2 instead of encounter_id, see here https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/pogom/pgoapi/protos/RpcSub_pb2.py#L1750.

Let me work out a solution using the newly found encounter_id now.

hhu94:
@viperfx I'm copying what @andrewda used in the 'pokemon' type message already implemented. https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2019/

AHAAAAAAA:
Can you rebase, it's changing 6 files when it only really needs to be changing 1

hhu94:
I'll open a new PR when this is ready.

PR #2997, opened 2016-08-01
Explicitly use http to avoid ssl certificate issues

springjools:
## Types of changes

## Checklist:

### DIFF
--- a/pogom/pgoapi/utilities.py
+++ b/pogom/pgoapi/utilities.py
@@ -56,7 +56,7 @@ def get_pos_by_name(location_name):
     if res:
         latitude, longitude, altitude = float(res.group(1)), float(res.group(2)), 0
     elif location_name:
-        geolocator = GoogleV3()
+        geolocator = GoogleV3(scheme='http')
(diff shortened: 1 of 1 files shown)


Ashex:
What sort of certificate issues does this avoid?

Certificates aren't validated.

https://github.com/geopy/geopy/blob/master/geopy/geocoders/googlev3.py#L66

springjools:
It's the same issues as listed here: https://github.com/Mila432/Pokemon_Go_API/pull/115/commits

I don't know more exactly what's the issue but with this fix it started working for me. I assume there are other users that have this issue.

Ashex:
I'd elect for updating the library used by the project rather than applying a single change from upstream.

Mr-Groch:
This will have an impact if I'm using nigix reverse proxy for https access to Pogomap?

jimdrie:
@Mr-Groch nope, as you can see in the commit this only concerns the Geolocation module, which is used to determine coordinates from a location string.

AHAAAAAAA:
Not touching the API.

PR #3050, opened 2016-08-01
Add an automatic retry if MySQL "goes away"

basepi:
## Description

Sometimes some of my workers would start failing in the upsert with "MySQL server has gone away". If this happens, they will forever try to upsert, throw the same error, and then continue. This pull request forces them to reinitialize the database connection which gives them a chance of recovering from temporary disconnects.

Working so far for me.

## Motivation and Context

Fixes intermittent database issues, keeps the workers alive

## How Has This Been Tested?

I have tested it with a series of workers which would hit this error about once a day. Since the change they have not died because of MySQL "going away".

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -347,7 +347,6 @@ def parse_map(map_dict, step_location):
 def bulk_upsert(cls, data):
     num_rows = len(data.values())
     i = 0
-    db_retry = 0
     step = 120
 
     flaskDb.connect_db()
@@ -357,16 +356,11 @@ def bulk_upsert(cls, data):
         try:
             InsertQuery(cls, rows=data.values()[i:min(i+step, num_rows)]).upsert().execute()
         except Exception as e:
-            # TODO: make the db_retry threshold configurable
-            if 'MySQL server has gone away' in str(e) and db_retry < 25:
+            if 'MySQL server has gone away' in str(e):
                 init_database()
-                db_retry += 1
-            else:
-                log.warning("%s... Retrying", e)
+            log.warning('%s... Retrying', e)
             continue
 
-        # Successful upsert, reset db_retry
-        db_retry = 0
         i += step
 
     flaskDb.close_db(None)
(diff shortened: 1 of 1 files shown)


pogom/models.py
```diff
@@ -356,10 +356,12 @@ def bulk_upsert(cls, data):
         try:
             InsertQuery(cls, rows=data.values()[i:min(i+step, num_rows)]).upsert().execute()
         except Exception as e:
+            if 'MySQL server has gone away' in str(e):
+                init_database()
             log.warning('%s... Retrying', e)
             continue
 
-        i+=step
+        i += step
```
basepi: This line is just my OCD linting. I'll remove it if you don't want the code churn.

pogom/models.py
```diff
@@ -356,10 +356,12 @@ def bulk_upsert(cls, data):
         try:
             InsertQuery(cls, rows=data.values()[i:min(i+step, num_rows)]).upsert().execute()
         except Exception as e:
+            if 'MySQL server has gone away' in str(e):
```
Ashex: Is there a proper exception being thrown? I really hate seeing such a general exception in here so it'd be better to catch the exact exception and handle it rather than doing a content check.

basepi: Agreed, ideally we would be catching the exact exception. In previous versions, it was actually `except OperationalError as e` and was not a normal exception. When I rebased to submit the fix, the `except Exception as e` was new. Perhaps the errors are being passed through without modification now.

I'd love to just get this merged and stop worrying about it, but if it's unacceptable in its current form I can remove the fix from my workers and try to reproduce again, see what it's actually throwing these days.

Ashex: Yeah, I completely forgot handling this specific error is a pain in the but. I guess we leave it as is :/

basepi: I think I'll add some logging to my servers to see what the error is, and log it out to a file or something, come clean this up later.

DomGrieco:
We need this!

PR #3062, opened 2016-08-01
Get correct s2 cells

EarthlingKira:
I've fixed incorrect S2 cell retrieval, see:
https://www.reddit.com/r/pokemongodev/comments/4vo59c/psa_servers_block_visibility_of_mons_from_some/

In some cases we never got the neighboring cell by following next/prev, so I used the proper methods for this

### DIFF
--- a/pogom/pgoapi/utilities.py
+++ b/pogom/pgoapi/utilities.py
@@ -70,18 +70,15 @@ def get_class(cls):
     return class_
     
 def get_cellid(lat, long):
-    origin = CellId.from_lat_lng(LatLng.from_degrees(lat, long)).parent(15)
-    walk = [origin.id()]
+    origin_cell = CellId.from_lat_lng(LatLng.from_degrees(lat, long)).parent(15)
+    walk = [origin_cell.id()]
 
-    # 10 before and 10 after
-    next = origin.next()
-    prev = origin.prev()
-    for i in range(10):
-        walk.append(prev.id())
-        walk.append(next.id())
-        next = next.next()
-        prev = prev.prev()
-    return ''.join(map(encode, sorted(walk)))
+    for direct_neighbor in origin_cell.get_all_neighbors(15):
+        walk.append(direct_neighbor.id())
+        for indirect_neighbor in direct_neighbor.get_edge_neighbors():
+            walk.append(indirect_neighbor.id())
+
+    return ''.join(map(encode, sorted(set(walk))))
 
 def encode(cellid):
     output = []


Tr4sHCr4fT:
@tejado can you upstream this please?

medghaim:
The issue you describe _will_ indeed occur in edge cases. Here's a more extreme example:
3 10 steps front back

This issue rears itself prominently in `-st 1` situations but when building a hexagon, the ingenuity of the Hilbert curve comes into play. 10 backpath + center + 10 frontpath (21 lvl15 cells total) from each step_location builds pretty complete coverage of the desired search area. Also, using a hilbert crawl preserves the integrity of cell order, which we can definitely use to our advantage to speed things up. 

Lastly, I personally discourage changes to pogoapi from here directly. This could cause some issues down the line when Tejado inevitably adds some cool new stuff to Pogoapi, and we want to switch to a newer version -- but we've already built off of changes you made here. I suggest you submit a pull to https://github.com/tejado/pgoapi, where the changes can be discussed with the man himself :)

AHAAAAAAA:
It's our intention to use pgoapi as a package soon, changes to the api should be directed to tejado!

EarthlingKira:
It seems like tejado implemented a different (more complex, not necessarily better ;-) ) solution:

https://github.com/tejado/pgoapi/blob/master/pgoapi/utilities.py

EarthlingKira:
@medghaim 
Sadly it's not so easy, you always have the problem, even with -st 15... The problem is some Pokemon not appearing on the map due to this cell problem. The server will only send you Pokemon in the 70m radius if you send the correct s2 cell id in the request, so this is always a problem

medghaim:
image

EarthlingKira:
@medghaim In your case it works because the "hilbert walk" luckily walked over both required s2 cells. But I have 2 cells in front of my flat where I *always* missed the pokemons in one spot with PokemonGo-Map due to this. I mean *always*, since days. And then I found out after debugging that exactly that s2 cell ID is *always* missing due to the "hilbert walk"

medghaim:
Yeah, I'm not arguing the edge cases don't exist. They do and it is (mostly) the cause of the missing pokemon, due to the bizarre path of the hilbert curve:
image

The cleanest solution to this would be to grab radius, as @tejado is currently doing. 


As for the overlap I was mentioning, it really does generate a _lot_ of overlap, so you're more likely to grab a pokemon than miss it:

image
This image plots the level 15 s2cells that _each_ scan location pulls in (ie, this DOESN'T include the 10 forward/10backwards steps) - if this were the state of things it would be scary, but:

image
This is with all the +10/-10 paths plotted as well. As you can imagine, there's a fair bit of overlap, but definitely not enough to find all the pokemon. 

RegionCoverer is cool because we can specify min/max levels, and build an _almost_ perfect circle. outta different level cells (ie, bigger smaller).

Dagothur:
@EarthlingKira 

Should I apply this change to my develop branch or switch over to the pogoapi-purge branch instead?

PR #3130, opened 2016-08-02
Custom alerts

smbambling:
## Description
Added a new argument -ca that has a static set of choices for custom alerts that can be triggered.  The models.py was updated to trigger customAlerts.py which is modeled after the customLog.py to display messages in the console.  The customAlerts.py reads in config values from a config/alerts.yaml file to get the values for the alarm triggers and alert configurations.  

This then triggers one or more custom alerts based on the -ca argument.  Current I have slack alerts finished.  Which is triggered via the custom_alerts/slackAlert.py.  This uses the Slacker module to publish messages to a specified channel in your Slack team via API key.

## Motivation and Context
Adding ability to push alerts directly via the app to services like slack or email for alerting on Pokemon.  There is a yaml config file that gives you the option to ignore or only trigger on Pokemon by name ID or Rarity.

## How Has This Been Tested?
This was tested on Ubuntu and OS X by running the server locally and verify that Pokemon found with the -dc flag set that were displaying on the console were also correctly getting triggered on via the customAlerts and they were arriving in my Slack team.

screen shot 2016-08-02 at 6 21 47 am

## Types of changes

## Checklist:

  Not sure where the configuration should be documented via the wiki ?  How does one go about updating the wiki in the PR ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ static/dist/
 node_modules/
 venv/
 core
+.DS_Store
+.ropeproject
--- a/config/alerts.yaml
+++ b/config/alerts.yaml
@@ -1,0 +1,18 @@
+---
+# Pokemon (Name,ID or Rarity) you ONLY want to alert on
+# Pokemon that are ignored below can be overridden here
+# Ex: You can ignore Pokemon with a common rarity but
+# still alert on individual Pokemon from that rarity class
+only:
+  - pidgey
+  - rare
+  - uncommon
+
+# Pokemon (Name,ID or Rarity) you DON'T want to alert on
+ignore:
+  - common
+
+# Slack Configuration Settings
+slack:
+  api_key: 'XXXXXXX-XXXXXXX-XXXXXXXX'
+  channel: '#pokemongo'
--- a/pogom/customAlerts.py
+++ b/pogom/customAlerts.py
@@ -1,0 +1,47 @@
+from .utils import get_pokemon_rarity, get_pokemon_name
+from pogom.utils import get_args
+from datetime import datetime
+from .custom_alerts.slackAlert import slackAlert
+from slacker import Slacker
+import yaml
+import os
+import smtplib
+
+args = get_args()
+alertpath = os.path.join(os.path.dirname(__file__), '../config/alerts.yaml')
+with open(alertpath, 'r') as stream:
+    try:
+        doc = yaml.load(stream)
+    except yaml.YAMLError as err:
+        print(err)
+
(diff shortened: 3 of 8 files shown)


ToonBoon:
I believe with the implementation of webhooks this (and any other notification pull requests) were deemed unnecessary.

smbambling:
@ToonBoon:

While I agree that that this could be done via the web hooks it would require yet another application/service running that receive the posts from the web hooks and then parse the data for any logic that would require alerting.

With the app already requiring a post to a web URL this doesn't seem like it would be much less efficient to just post directly to a alerting service such as Slack or Email.

ToonBoon:
I agree wholeheartedly. I implemented pushbullet notifications without webhook for my own map instance.

Just saying it likely wont get merged because they wanted devs to code around webhooks.

smbambling:
If thats the direction that they are going to force devs then I can see where keeping the app 'light' can be a good thing.  Hopefully it will at least get considered for approval.

AHAAAAAAA:
I love this! I'll probably make use it of myself, but we have decided (like ToonBoon mentioned) to rely on webhooks. Reason being that we have so much bloat and, with the pace of development, we're unable to support all of these things simultaneously. Webhooks are easier to maintain, and leads to a lighter, cleaner app. We'd like to focus on being a map and separate all these cool things built around the map from the actual code. I believe we should start a wiki page with a collection of all the cool things people built with the webhooks, you're welcome to start one!

joberembt:
Trying to use this and when I try to install pyyaml, I get the following error:
`Command "c:\python27\python.exe -u -c "import setuptools, tokenize;__file__='c:\\users\\XXXXXXXX\\appdata\\local\\temp\\pip-build-XXXXXXXX\\pyyaml\\setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record c:\users\XXXXXXXX\appdata\local\temp\pip-XXXXXXXX-record\install-record.txt --single-version-externally-managed --compile" failed with error code 1 in c:\users\XXXXXXXX\appdata\local\temp\pip-build-XXXXXXXX\pyyaml\`

This is on windows with python 2.7

joberembt:
Got it figured out - installed https://www.microsoft.com/en-gb/download/details.aspx?id=44266 and then it worked.

smbambling:
Glad you got it working I'll try to keep this fork/branch up to date with dev while I find some time to work on a separate customs alerts app feed from the  rest posts out putted from the map.

PR #3179, opened 2016-08-02
Added supervisord files to repo

DirtyCajunRice:
## Description
Supervisor is a client/server system that allows its users to monitor and control a number of processes on UNIX-like operating systems. Implemented to manage multiple workers through one interface CLI/Web.

## Motivation and Context
Features: 

* Worker Job control
* Per worker logs
* Web interface

## How Has This Been Tested?
many times

## Types of changes

## Checklist:


Ashex:
Wouldn't something like this be better suited as a wiki entry?

DirtyCajunRice:
@Ashex It was asked in discord by mods to be put in the repo. So i obliged

Ashex:
If this is going to be merged in it probably should not go under `Tools`, that folder has a terrible name. Perhaps it should go under a new directory `contrib/supervisord` as that seems to be the standard for other open source projects.

https://github.com/haproxy/haproxy/tree/master/contrib

Also can you update the PR with actual details rather than assume people know what this is for/what is does? Also needs to be a readme thrown in there explaining the function and how to use it.

DirtyCajunRice:
@Ashex i agree i just was using what was existing. will move and add

contrib/supervisord/template.ini
```diff
@@ -0,0 +1,10 @@
+[program:PokeGoMapWorkerWRK]
+command=/usr/bin/python DIRECTORYrunserver.py -ns -l "LOC" -st ST -sd SD -ld LD -a AUTH USER -p PASS
```
invisiblek: need a -u in here before USER?

contrib/supervisord/gen-workers.sh
```diff
@@ -0,0 +1,68 @@
...
+# Webserver Location
+initloc="Dallas, TX"
+# Account name without numbers
+pre="accountname"
+
+# Variables
+worker=0
+acct1=0
+numacct=5
+pass="yourpasshere"
```
invisiblek: Would be really nice if auth/user/pass could be parsed from a file. Would be even cooler if the file could specify several auth/user/pass combinations per line too. Maybe even have them in the coordinates file. Something like this could work probably:

`54.00000000, -94.00000000|auth1,user1,pass1|auth2,user2,pass2`

Or something like that, make it easy to parse.

contrib/supervisord/template.ini
DirtyCajunRice: @invisiblek it is parsed in per user via lines 48-50 of gen-workers.sh
+  user=$(for (( i = 1; i <= numacct; i++ )) do
 +           echo -n "-u ACCT$i "
 +         done)

contrib/supervisord/gen-workers.sh
DirtyCajunRice: This seems possible in a coords file and user/pass file. i dont quite see the purpose of adding auth as i dont know of a single person mixing gmail and ptc accounts.

BaIthamel: some people do odd things

PR #3189, opened 2016-08-02
Fix requirements.txt for Mac OS X System Python

EdenGottlieb:
## Description

Mac OS X El Capitan ships with six version 1.4.1 installed to its Python. Users who use their system python (not recommended but done by many) will encounter an error while running `pip install -r requirements.txt` :

`OSError: [Errno 1] Operation not permitted`

This is because Mac OS X does not allow upgrade of included packages.
Changing six minimum version to 1.4.1 fixes this problem, and I did not seem to encounter any compatibility issues while using it.

## How Has This Been Tested?

It works

## Types of changes

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-six==1.10.0
+six>=1.4.1
 Flask==0.11.1
 Jinja2==2.8
 MarkupSafe==0.23


macgngsta:
it seems that protobuf-to-dict requires  `pkg_resources.VersionConflict: (six 1.8.0 (/usr/lib/python2.7/dist-packages), Requirement.parse('six>=1.9'))`

EdenGottlieb:
@macgngsta Never encountered that issue and didn't found any reference of it?... protobuf-to-dict source code contains no mention of six whatsoever.

EdenGottlieb:
That requirement is made from protobuf, not protobuf-to-dict. 
It is present in protobuf 3.0.0, but not in protobuf 2.6.1 that our project requires. So no compatibility issues with the old version we're using.

PR #3197, opened 2016-08-02
Hide "Start at user's location" when -fl is on

EdenGottlieb:
Hide "Start at user's location" when -fl is on.
## Description

Whoever created "Start at user's location" forgot to hide it when the fixed-location (-fl) flag is on.

## How Has This Been Tested?

It works

## Types of changes

## Checklist:

### DIFF
--- a/templates/map.html
+++ b/templates/map.html
@@ -57,7 +57,7 @@
 	    </label>
 	  </div>
 	</div>
-        <div class="form-control switch-container" >
+        <div class="form-control switch-container" style="display:{{is_fixed}}">
           <h3>Start at user's location</h3>
           <div class="onoffswitch">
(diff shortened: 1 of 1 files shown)


invisiblek:
I left it this way intentionally.

-fl is intended to not allow end users to change the scan location via the front end. This option doesn't change the scan location, but rather where the map starts. This is useful since I share my map with several people and some of them live on the other side of the city or miles away. Their map shouldn't start at my house (where my web instance begins) but rather wherever they are.

xrobau:
Confirmed that @invisiblek 's idea of the UI is correct. I share it with a bunch of people in the local area, they want it to start at their location.

EdenGottlieb:
Oh, I must have misunderstood the feature. It doesn't change the map's initial scan location but only focuses it on where the user is at (without starting a scan from there).
PR closed

PR #3208, opened 2016-08-02
command line for host have changed

Daniel-ltw:
Since the cmd line for host have changed from --host to -H, this has not been updated. 

## Description
Change entrypoint to use -H
Update Dockerfile with new host param key


invisiblek:
https://github.com/AHAAAAAAA/PokemonGo-Map/blob/develop/pogom/utils.py#L55

--host should work just fine...

Daniel-ltw:
Cool. 


Thought there was a change that was not reflected well in the Dockerfile.

PR #3241, opened 2016-08-03
Added encouter_ID and time_till_hidden

ferazambuja:
Added encouter_ID and time_till_hidden. It worked for me. Tried to add the least possible.
encounter_ID can be encoded as base64 like pokemongo-map.


AHAAAAAAA:
Not merging against master. PR against develop

PR #3262, opened 2016-08-03
Increase scan reliability

chuyskywalker:
- Handle exceptions better
 - Prevent search_workers from even fully exiting
 - High-load sleep timer is now backoff based instead of a randomly assigned absolute value
 - Logins will retry `login-retries` times before throwing an exception to the search_worker thus creating a new PgoAPI instance and starting over

### DIFF
--- a/pogom/__init__.py
+++ b/pogom/__init__.py
@@ -6,8 +6,5 @@ config = {
     'LOCALES_DIR': 'static/dist/locales',
     'ROOT_PATH': '',
     'DATA_DIR': 'static/dist/data',
-    'GMAPS_KEY': None,
-    'REQ_SLEEP': 5,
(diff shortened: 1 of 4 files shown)


DomGrieco:
What's the purpose of the `login-retries`? Setting it to a lower number would mean less chance of looking suspicious?

chuyskywalker:
Essentially, I've had instances where the PgoAPI `api.login` calls got "stuck". It would half login, fail, and never actually get logged back in. Because it's a single call, it would do this forever. Placing a limit on this allows the search_worker thread to bail out and start over -- creating a new `api` instance solves the problem*.

_\* I "turn it off and on again" essentially_

ELusiveLPFAN:
You want to say that I don't see message about 15minutes ban anymore? ^_^

PR #3106, opened 2016-08-02
Pgoapi purge

invisiblek:
Kill our local copy of pgoapi and instead pull in a fresh copy from tejado's repo. Fixups along the way to handle the api changes.

PR #3302, opened 2016-08-03
Lured pokemon webhook

hhu94:
## Description
Implemented a webhook for lured pokemon that is parallel to the normal pokemon webhook.

## Motivation and Context
This allows third party apps like PokeAlarm to notify when a lured pokemon appears.
The new webhook works directly with PokeAlarm's latest version without the need of any changes from their part because it follows the message follows the same basic structure as that of normal pokemon's.

One thing to keep in mind is the 'unknown2' field in the FortLureInfo proto. This is actually a 'encounter_id' field in the newer pgoapi version. So the 'unknown2' in this patch will have to be changed to 'encounter_id' when we eventually update our pgoapi package as per #3106 

## How Has This Been Tested?
Lured pokemon are displayed without a problem on PokeAlarm's latest version.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -331,6 +331,8 @@ def parse_map(map_dict, step_location):
                     lure_expiration = datetime.utcfromtimestamp(
                         f['lure_info']['lure_expires_timestamp_ms'] / 1000.0)
                     active_pokemon_id = f['lure_info']['active_pokemon_id']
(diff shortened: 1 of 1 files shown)


AHAAAAAAA:
Putting this on hold until pgoapi purge goes through, so we can update unknown2 prior to merging.

PR #3304, opened 2016-08-03
Revert "Add Proxy Support"

invisiblek:
Reverts AHAAAAAAA/PokemonGo-Map#2814

We shouldn't be touching pgoapi. If changes are needed the, submit them to upstream.


tim95030:
I have a pull request to pgoapi. I am working an the implementation there.

PR #2815, opened 2016-07-31
Beehive fix

BaIthamel:
## Description

Commas are really really important!
and some comments
## Motivation and Context

#2810 #2796 #2879 
## How Has This Been Tested?

extensivly #killmenow :(
## Screenshots (if appropriate):

image

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #3305, opened 2016-08-03
Fix beehive alignment

reyqn:
## Description

Beehive alignment is broken because when giving latitude / longitude to pgoapi, it returns the latitude and longitude of a place next to what we want. So it isn't very accurate.
I added a way to give exact latitude and longitude as a location. This should fix the issue.

## Motivation and Context

Beehive scan isn't optimal. Some people were complaining about it so here is a fix.

## How Has This Been Tested?

I tested it quickly so feel free to check it a little more. 

## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -70,25 +70,14 @@ if __name__ == '__main__':
         logging.getLogger('pgoapi').setLevel(logging.DEBUG)
         logging.getLogger('rpc_api').setLevel(logging.DEBUG)
 
-<<<<<<< f3f7dbaea4c210845b0eb96c8d1bbcd7bef5f48e
-<<<<<<< b87fd21b6aa240358ba00301f3bff93ed9781e93
-
-    position = util.get_pos_by_name(args.location)
-    if not any(position):
-        log.error('Could not get a position by name, aborting')
-        sys.exit()
-=======
-    if  args.latitude is not None and args.longitude is not None:                                                                                                    
-=======
+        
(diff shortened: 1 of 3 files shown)


runserver.py
```diff
@@ -55,10 +55,14 @@
         logging.getLogger('pgoapi').setLevel(logging.DEBUG)
         logging.getLogger('rpc_api').setLevel(logging.DEBUG)
 
-    position = get_pos_by_name(args.location)
-    if not any(position):
-        log.error('Could not get a position by name, aborting')
-        sys.exit()
+
+    if  (args.latitude is not None and args.longitude is not None):
```
Ashex: Is there any reason you can't just do a validation check here to determine if `--location` is a GPS coordinate and handle it here?

Adding extra arguments when location is available does make this a little more difficult to maintain.

reyqn: Cause I was lazy

BaIthamel:
What is wrong with beehive?

reyqn:
@BaIthamel https://github.com/AHAAAAAAA/PokemonGo-Map/pull/3106#issuecomment-236888039

BaIthamel:
I fixed that already #2815

reyqn:
@BaIthamel well yes but #2818 was reverted by #3106. However @invisiblek fixed it already in https://github.com/tejado/pgoapi/pull/171 5 hours ago. Closing.

PR #3356, opened 2016-08-03
Added STOPSIGNAL entry to Dockerfile

mrik974:
I noticed a long delay when using `docker stop`.
Read the documentation about this [and this blog post in particular](https://www.ctl.io/developers/blog/post/gracefully-stopping-docker-containers/) to discover Docker sends a `SIGTERM` first and then a deadly `SIGKILL` if the app doesn't respond to `SIGTERM`. Python by default only takes `SIGINT` to stop (the equivalent of a `Ctrl+C`).

## Description
Added a `STOPSIGNAL` configuration to the original `Dockerfile` so the `docker stop` command will send the right signal right away

## Motivation and Context
The `docker stop` command will respond with no delay and the server will stop gracefully

## How Has This Been Tested?
Tested by myself with the command `docker stop` :).

## Types of changes

## Checklist:


chuyskywalker:
lgtm

tkSimon:
Note that this requires docker 1.9 or higher. I'd love to see this in develop, though at the moment this isn't even supported by docker hub, see docker/hub-feedback#460.

Maybe add a comment saying that this should be commented out if running on docker<1.9?

mrik974:
@tkSimon you are right, thanks for reminding me. 
However I think it will work on Dockerhub, since the issue you referenced is 9 months old and is closed now :).

PR #3372, opened 2016-08-03
Allow alternate config file to be specified using an environment variable

cdhowie:
## Description
Allows using the `POGOMAP_CONFIG` environment variable to specify a different `config.ini` file than `$CODE_ROOT/config/config.ini`.

## Motivation and Context
This allows me to rather easily run multiple instances of the server with separate configuration options (including the database) against one copy of the codebase.  This is not strictly *necessary* to run multiple copies of the map, but the alternatives have downsides:

* You can use arguments to `runserver.py` instead.  A primary downside of this approach is that PTC/Google and possible also DB credentials are passed on the command-line.  On most *nix systems, this exposes them to all processes running on the same system -- even other users.  On a shared system, this is a grave security risk.
* You can make a complete copy of the PokemonGo-Map directory.  The rather obvious downside here is that each copy takes about half a gigabyte of space, by the time you add in dependencies.  Additionally, each copy has to be separately maintained for updates to PokemonGo-Map and all npm/pip dependencies.

## How Has This Been Tested?
I am currently running multiple instances using this approach, using the `db` option in each `config.ini` so that each instance has its own SQLite database.  I have had no issues so far.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -34,7 +34,7 @@ def verify_config_file_exists(filename):
 
 def get_args():
     # fuck PEP8
-    configpath = os.path.join(os.path.dirname(__file__), '../config/config.ini')
+    configpath = os.environ.get('POGOMAP_CONFIG', os.path.join(os.path.dirname(__file__), '../config/config.ini'))
     parser = configargparse.ArgParser(default_config_files=[configpath])
     parser.add_argument('-a', '--auth-service', type=str.lower, action='append',
                         help='Auth Services, either one for all accounts or one per account. \


invisiblek:
Why would you do something like this? If anything this should be a parameter, not a read from the environment.

cdhowie:
@invisiblek Because as it is right now, argument defaults are effectively read from the config file first, then the arguments are processed.  By the time we are parsing arguments, we've already read the config file.

I agree that specifying it as a config switch would be preferable, but that's a significantly more involved change.

Ashex:
This is a very uncommon usage, that is the primary issue.

Also, configargparse does support pulling settings from variables so if you add an argument you should be able to set it with the method described.

cdhowie:
> configargparse does support pulling settings from variables

I'm not sure exactly what this means and therefore how it's relevant.

Here's what I see happening.  On [this line](https://github.com/cdhowie/PokemonGo-Map/blob/a38976940cf913fdc3f6fd674b93b03326ce3142/pogom/utils.py#L38), the argument parser is created, specifying a config file from which to pull default values.  On [line 123](https://github.com/cdhowie/PokemonGo-Map/blob/a38976940cf913fdc3f6fd674b93b03326ce3142/pogom/utils.py#L123) the arguments are actually parsed.

We need to parse the arguments in order to obtain the location of the config file, which we need in order to parse the config file... before we parse the arguments.

Do you see the problem, at least with how this method is currently written?  There's a chicken-and-egg problem (need to parse arguments to determine the location of the config file, but need the config file location before we can parse arguments), *unless* we add another parser that runs first just to parse the config file option.  (I'm not saying we can't do that, it just seems a bit unclean to do it that way.)

As a side note, I'm not terribly invested in this patch being accepted.  It's useful to me; I thought it would be useful to others.  If it's unwanted I can maintain it in my fork for my own use.

Ashex:
Line 38 just includes a set of default locations to look for the config.ini and load it. 

You can specify an argument then where you declare the parser add [args_for_setting_config_file_path](https://github.com/bw2/ConfigArgParse/blob/master/configargparse.py#L331) which tells it which argument to find config.ini in a custom path.

Read the documentation on the library used for parsing arguments and you'll understand why I'm suggesting this approach.

cdhowie:
> You can specify an argument then where you declare the parser add args_for_setting_config_file_path which tells it which argument to find config.ini in a custom path.

Ah, perfect.  I was not aware this option was present.  (*"does support pulling settings from variables"* is a bit vague, so while I didn't understand what you meant at the time, I can see now what you were trying to say.)

invisiblek:
https://github.com/AHAAAAAAA/PokemonGo-Map/pull/3390

PR #3390, opened 2016-08-03
Config parameter

invisiblek:


Ashex:
Maaaaaaan I was just working on this. Can you add support for using variables too?

```python
configargparse.ArgParser(auto_env_var_prefix='POGOM_')
```

invisiblek:
Was an old commit I resurrected from the dead: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2572

auto_env_var_prefix: If set to a string instead of None, all config-
                file-settable options will become also settable via environment
                variables whose names are this prefix followed by the config
                file key, all in upper case. (eg. setting this to "foo_" will
                allow an arg like "--my-arg" to also be set via the FOO_MY_ARG
                environment variable)

FOO_MY_ARG!!!!!!!!!!

sure - i'll add it

invisiblek:
Hmm, auto_env_var_prefix doesn't appear to work for specifying the config file,
This worked however:
`exporting POGOM_DB_TYPE=sqlite`

Ashex:
That sorta makes sense since you can specify all the values by env var. Considering the use case I don't think it's a blocker.

PR #3399, opened 2016-08-03
Add option to only calculate and store scanned locations

invisiblek:
* Useful with a very low --scan-delay to quickly plot a hex

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -287,96 +287,97 @@ class ScannedLocation(BaseModel):
         return scans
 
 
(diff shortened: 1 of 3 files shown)


justsomedudeonthenet:
Will this run without accessing the API at all?  Not even trying to login?

It would be really handy for testing large scan areas without waiting for a whole scan to run.

chuyskywalker:
I kinda feel like we should just create `/beehive` and just plop together some JS for the circle drawing all up in that URL.

